PR #2, opened 2016-11-22
Reorganize steps closer to the goal order, without doing any grouping yet

evykassirer:


aelnaiem:
simplifyBasics isn't a very clear name. What does it mean? Otherwise, this looks good 👍

PR #3, opened 2016-11-22
adding constant fractions, grouping

evykassirer:


lib/ConstantFraction.js
```diff
@@ -176,13 +204,34 @@ ConstantFraction.addConstantAndFraction = function(node) {
     newNode.args[0] = newFractionNode;
     newNode.args[1] = newConstNode;
   }
-  return NodeStatus.nodeChanged(changeType, oldNode, newNode);
+
```
evykassirer: this function might be too long now? not sure if it makes sense to break it up though

lib/NodeStatus.js
```diff
@@ -32,13 +33,18 @@ class NodeStatus {
...
     // clone so that the next step's oldNode doesn't modify this step's newNode
-    this.newNode = clone(this.newNode, false);
-    this.newNode.filter(node => node.changeGroup).forEach(change => {
-      delete change.changeGroup;
-    });
+    this.newNode = NodeStatus.resetChangeGroups(this.newNode);
   }
 }
 
+NodeStatus.resetChangeGroups = function(node) {
```
evykassirer: not sure if this is confusing that two `resetChangeGroups` have the same name (one can be called on a node status object, the other is a static function)

lib/ConstantFraction.js
```diff
@@ -176,13 +212,34 @@ ConstantFraction.addConstantAndFraction = function(node) {
...
+  newNode = NodeStatus.resetChangeGroups(lastStep.newNode);
+
+  return NodeStatus.nodeChanged(changeType, node, newNode, true, subSteps);
 };
 
 // Given a list of nodes `fractionNodes` that all have the same denominator,
 // add them together. e.g. 2/3 + 5/3 -> (2+5)/3
 // Returns the new node.
-function addFractionsWithSameDenominator(fractionNodes) {
+function writeNumeratorsAboveCommonDenominator(fractionNodes) {
```
aelnaiem: why `write` not `combine`?

evykassirer: I think I got the message from some other app

combine makes more sense, especially at this level

lib/NodeStatus.js
aelnaiem: Let's rename one `reset` because it is more generic

aelnaiem:
A few comments, but otherwise 👍

lib/ConstantFraction.js
```diff
@@ -68,12 +68,14 @@ ConstantFraction.simplifyFraction = function(fraction) {
...
-// e.g. 2/3 + 4/6 --> (2*2)/(3*2) + 4/6
-// e.g. 2/5 + 4/5 --> (2+4)/5
-// returns a NodeStatus object.
+// Adds constant fractions -- can start from either step 1 or 2
+// 1A. Find the LCD if denominators are different and multiplies to make
+//     denominators equal, e.g. 2/3 + 4/6 --> (2*2)/(3*2) + 4/6
+// 1B. Multiplies out to make constant fractions again
+//     e.g. (2*2)/(3*2) + 4/6 -> 4/6 + 4/6
+// 2A. Combines numerators, e.g. 4/6 + 4/6 ->  e.g. 2/5 + 4/5 --> (2+4)/5
+// 2B. Adds numerators together, e.g. (2+4)/5 -> 6/5
```
aelnaiem: Should each of these be a small function?

evykassirer: I think because of the changing around of oldNode and newNode and subSteps, calling a function and then updating the variables would be just as much code and a bit less clear

I made this area a bit cleaner by calling functions e.g. `  let newNode = combineNumeratorsAboveCommonDenominator(oldNode.args);` so I'm okay with how it is right now

PR #7, opened 2016-11-23
adding like terms

evykassirer:


lib/PolynomialTermOperations.js
```diff
@@ -107,9 +112,46 @@ function addlikeTermPolynomialNodes(node) {
...
+  nodeStatus = addNegativeOneCoefficient(oldNode);
+  if (nodeStatus.hasChanged()) {
+    subSteps.push(nodeStatus);
+    // clone and reset
+    oldNode = NodeStatus.resetChangeGroups(nodeStatus.newNode);
+  }
+
+  // STEP 3: group the coefficients in a sum
+  let newNode = groupCoefficientsForAdding(clone(oldNode, false));
+  nodeStatus =  NodeStatus.nodeChanged(                                         // TODO - change groups better
```
evykassirer: these change groups should be only on the coefficients and coefficient sum

lib/PolynomialTermOperations.js
```diff
@@ -119,12 +161,80 @@ function addlikeTermPolynomialNodes(node) {
...
+
+function addPositiveOneCoefficient(node) {                                      // TODO - change groups better
+  const oldNode = node;
+  const newNode = clone(node, false);
+  let change = false;
+
+  newNode.args = newNode.args.map(polyNode => {
+    const polyTerm = new PolynomialTermNode(polyNode);
+    if (polyTerm.getCoeffValue() === 1) {
+      change = true;
```
evykassirer: I want to set change groups on the old node and new node with new coefficient

lib/PolynomialTermOperations.js
```diff
@@ -119,12 +161,80 @@ function addlikeTermPolynomialNodes(node) {
...
+  const oldNode = node;
+  const newNode = clone(node, false);
+  // the node is now always a * node with the left child the coefficent sum
+  // e.g. (2 + 4 + 5) and the right node the symbol part e.g. x or y^2
+  const coefficientSum = newNode.args[0];
+  coefficientSum.changeGroup = 1;
+  let newCoefficient = evaluateConstantSum(coefficientSum).newNode;             // TODO steps
+  newCoefficient.changeGroup = 1;
+  newNode.args[0] = newCoefficient;
   return NodeStatus.nodeChanged(
```
evykassirer: this should be a childChanged status

lib/PolynomialTermOperations.js
```diff
@@ -107,9 +112,46 @@ function addlikeTermPolynomialNodes(node) {
...
+  if (nodeStatus.hasChanged()) {
+    subSteps.push(nodeStatus);
+    // clone and reset
+    oldNode = NodeStatus.resetChangeGroups(nodeStatus.newNode);
+  }
+
+  // STEP 3: group the coefficients in a sum
+  let newNode = groupCoefficientsForAdding(clone(oldNode, false));
+  nodeStatus =  NodeStatus.nodeChanged(                                         // TODO - change groups better
+    MathChangeTypes.ADD_POLYNOMIAL_TERMS, oldNode, newNode);
```
evykassirer: make a new change type for grouping coefficients

lib/evaluateArithmetic.js
```diff
@@ -13,6 +13,15 @@ function evaluateArithmeticDFS(node) {
   let innerNodeStatus;
   if (NodeType.isParenthesis(node)) {
     innerNodeStatus = evaluateArithmeticDFS(node.content);
+    if (innerNodeStatus.hasChanged()) {
```
evykassirer: don't have to do this anymore cause it should be handled before this function is called

test/PolynomialTerm.test.js
```diff
@@ -36,13 +36,37 @@ describe('classifies symbol terms correctly', function() {
 function testCombinePolynomialTerms(exprStr, outputStr) {
   it(exprStr + ' -> ' + outputStr, function () {
     const inputNode = flatten(math.parse(exprStr));
-    const combinedNode = PolynomialTermOperations.combinePolynomialTerms(inputNode).newNode;
+    const combinedStatus = PolynomialTermOperations.combinePolynomialTerms(inputNode);
+    combinedStatus.subSteps.forEach(step => {
```
evykassirer: get rid of these lines

lib/PolynomialTermOperations.js
```diff
@@ -119,12 +161,80 @@ function addlikeTermPolynomialNodes(node) {
...
+  if (change) {
+    return NodeStatus.nodeChanged(
+        MathChangeTypes.ADD_COEFFICIENT_OF_ONE, oldNode, newNode);
+  }
+  else {
+    return NodeStatus.noChange(newNode);
+  }
+}
+
+function addNegativeOneCoefficient(node) {                                      // TODO - change groups better
```
evykassirer: todo: docstrings for new functions

evykassirer:
ready to be looked at now, just adding tests

lib/evaluateConstantSum.js
```diff
@@ -0,0 +1,133 @@
...
+// Returns a NodeStatus object.
+function evaluateConstantSum(node) {
+  if (NodeType.isParenthesis(node)) {
+    node = node.content;
+  }
+  if (!NodeType.isOperator(node) || node.op !== '+') {
+    return NodeStatus.noChange(node);
+  }
+
+  // functions needed to evalute the sum
```
evykassirer: **evaluate

lib/PolynomialTermOperations.js
```diff
@@ -107,24 +111,150 @@ function addlikeTermPolynomialNodes(node) {
   if (!canAddLikeTermPolynomialNodes(node)) {
     return NodeStatus.noChange(node);
   }
-  const polynomialTermList = node.args.map(n => new PolynomialTermNode(n));
 
-  // get a list of the coefficients, to be added
+  let oldNode = clone(node, false);
+  const subSteps = [];
+
+  // STEP 1: IF any nodes have no coefficient, make it have coefficient 1
```
aelnaiem: Why is the if all uppercase?

lib/evaluateConstantSum.js
```diff
@@ -0,0 +1,133 @@
...
+  const summingFunctions = [
+    evaluateArithmetic,
+    ConstantFraction.addConstantFractions,
+    ConstantFraction.addConstantAndFraction,
```
aelnaiem: Does addConstantAndFraction add constants to variable fractions or to constant fractions? Or both. This suggests both.

lib/PolynomialTermOperations.js
evykassirer: to try to make it clear that that step doesn't always happen, but I don't think it's very effective :p I'll update it and step 2 as well

aelnaiem:
Couple comments about naming, but otherwise looks good!! 👍

lib/evaluateConstantSum.js
```diff
@@ -0,0 +1,133 @@
...
+        return status;
+      }
+    }
+  }
+
+  let oldNode = clone(node, false);
+  const subSteps = [];
+
+  // STEP 1: group fractions and constants separately
+  let status = groupConstantsAndFractions(oldNode);
```
evykassirer: I use constant in most places in the code to refer to constant nodes, ie a single node with any number value (whole number or decimal number)

should I add this to the readme? should we rename things?

lib/evaluateConstantSum.js
evykassirer: both

lib/evaluateConstantSum.js
aelnaiem: ah I get it, that makes sense

lib/evaluateConstantSum.js
aelnaiem: cool!

PR #8, opened 2016-11-27
adding a simple function to get prime factors for a number

aelnaiem:
[asana task](https://app.asana.com/0/201972293666219/205201279084528)

### DIFF
--- a/lib/Factor.js
+++ b/lib/Factor.js
@@ -2,13 +2,13 @@
 
 // This module deals with getting factors
 
-const Factors = {};
+const Factor = {};
 
-Factors.getPrimeFactors = function(number){
+Factor.getPrimeFactors = function(number){
   let factors = [];
   if (number < 0) {
     factors = [-1];
-    factors = factors.concat(Factors.getPrimeFactors(-1 * number));
+    factors = factors.concat(Factor.getPrimeFactors(-1 * number));
     return factors;
   }
 
@@ -28,10 +28,10 @@ Factors.getPrimeFactors = function(number){
   else {
     // if number isn't prime factor make recursive call
     factors.push(candidate);
-    factors = factors.concat(Factors.getPrimeFactors(number/candidate));
+    factors = factors.concat(Factor.getPrimeFactors(number/candidate));
   }
 
   return factors;
 };
 
-module.exports = Factors;
+module.exports = Factor;
--- a/lib/Factor.js
+++ b/lib/Factor.js
@@ -4,6 +4,7 @@
 
 const Factor = {};
 
+// Given a number, will return all the prime factors of that number as a list
 Factor.getPrimeFactors = function(number){
   let factors = [];
   if (number < 0) {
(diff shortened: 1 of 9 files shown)


test/Negative.test.js
```diff
@@ -15,7 +15,7 @@ function testNegate(exprString, outputStr) {
   });
 }
 
-describe('negatePolynomialTerm', function() {
+describe('negate', function() {
```
evykassirer: thanks!

test/solveEquation.test.js
```diff
@@ -69,15 +69,15 @@ describe('solveEquation for non = comparators', function() {
...
   const leftNode = math.parse(sides[0]);
   const rightNode = math.parse(sides[1]);
 
   const steps = solveEquation(leftNode, rightNode, comparator, debug);
-  const acualExplanation = steps[steps.length -1].explanation;
-  it(equationString + ' -> ' + expectedExplanation, function () {
-    assert.equal(acualExplanation, expectedExplanation);
+  const acualChange = steps[steps.length -1].changeType;
+  it(equationString + ' -> ' + expectedChange, function () {
+    assert.equal(acualChange, expectedChange);
```
evykassirer: *actual

lib/Factor.js
```diff
@@ -0,0 +1,37 @@
...
+  const root = Math.sqrt(number);
+  let candidate = 2;
+  if (number % 2) {
+    candidate = 3; // assign first odd
+    while (number % candidate && candidate <= root) {
+      candidate = candidate + 2;
+    }
+  }
+
+  // if no factor found then number is prime
```
evykassirer: or 1 or 0 ;D

evykassirer: ^ what a nerd

evykassirer:
mostly comments about your comments, but otherwise looks good!

lib/Factor.js
```diff
@@ -0,0 +1,37 @@
...
+  if (candidate > root) {
+    factors.push(number);
+  }
+  else {
+    // if number isn't prime factor make recursive call
```
aelnaiem: No if it's not, then we want to know what the prime factors of it are.

lib/Factor.js
```diff
@@ -34,4 +34,18 @@ Factor.getPrimeFactors = function(number){
   return factors;
 };
 
+Factor.getFactorPairs = function(number){
```
evykassirer: oh yeah! we'll need this for quadratics

docstring here? also should we handle number being negative here or before calling this?

lib/Factor.js
evykassirer: ohhh, got it

can you make the comment "if the factor isn't prime, make a recursive call to add its prime factors" or something like that? I thought number was `number`

aelnaiem:
I also added a function to get factor pairs (which will eventually be used in factoring binomials)

lib/Factor.js
aelnaiem: Number is `number`. Basically, if we find a prime factor, we divide the number by that prime factor and get more prime factors. I'll add that though

evykassirer: oh, yeah that makes sense

PR #9, opened 2016-11-29
collect and combine (now with multiplication)

evykassirer:


aelnaiem:
👍

aelnaiem:
With this stuff, I always think, I've seen this before... hahaha looks good :)

PR #10, opened 2016-11-29
adding root support functionality

aelnaiem:
This code ended up being a mess...

[asana task](https://app.asana.com/0/201972293666219/198715212854040)

### DIFF
--- a/lib/evaluateFunctions.js
+++ b/lib/evaluateFunctions.js
@@ -192,12 +192,18 @@ function nthRoot(node) {
           for (let i = 0; i < groupedChildren.length; i++) {
             const child = groupedChildren[i];
             if (NodeType.isParenthesis(child)) {
-              // TODO(ael): check that it's a multiplication of the same node n times, and
-              // if it is, convert to an exponent...
-              const baseNode = child.content.args[0];
-              const exponentNode = NodeCreator.constant(rootValue);
-              const newChild = NodeCreator.operator('^', [baseNode, exponentNode]);
-              groupedChildren[i] = newChild;
+              const grandChild = child.content;
+              if (NodeType.isOperator(grandChild) && grandChild.op === '*') {
+                const allEqual = grandChild.args.reduce((a, b) => {
+                  return a.equals(b);
+                });
+                if (allEqual) {
+                  const baseNode = child.content.args[0];
+                  const exponentNode = NodeCreator.constant(rootValue);
+                  const newChild = NodeCreator.operator('^', [baseNode, exponentNode]);
+                  groupedChildren[i] = newChild;
+                }
+              }
             }
           }
 
--- a/lib/evaluateFunctions.js
+++ b/lib/evaluateFunctions.js
@@ -166,10 +166,14 @@ function nthRootExponent(node) {
 
 // Returns the nthRoot evaluated for a multiplication node.
 // Expects a multiplication node uder the radicand.
-// If the root is a positive constant, it factors the multiplicands, groups them
-// and converts the multiplications into exponents.
-// Distributes the nthRoot into the children nodes, evaluates those nthRoots
-// and combines them
+// If the root is a positive constant, it:
+//  1A: factors the multiplicands
+//  1B: groups them into groups whose length is the root value
+//  1C: converts the multiplications into exponents.
+// If it's possible to simplify further, it:
+//  2A: Distributes the nthRoot into the children nodes,
+//  2B: evaluates those nthRoots
+//  2C: combines them
 function nthRootMultiplication(node) {
   let newNode = clone(node, false);
   const rootNode = getRootNode(node);
@@ -177,18 +181,21 @@ function nthRootMultiplication(node) {
   let substeps = [];
   let status;
   if (NodeType.isConstant(rootNode) && !Negative.isNegative(rootNode)) {
+    // Step 1A
     status = factorMultiplicands(newNode);
     if (status.hasChanged()) {
       substeps.push(status);
       newNode = NodeStatus.resetChangeGroups(status.newNode);
     }
 
+    // Step 1B
     status = groupTermsByRoot(newNode);
     if (status.hasChanged()) {
       substeps.push(status);
       newNode = NodeStatus.resetChangeGroups(status.newNode);
     }
 
+    // Step 1C
     status = convertMultiplicationToExponent(newNode);
     if (status.hasChanged()) {
       substeps.push(status);
@@ -202,15 +209,18 @@ function nthRootMultiplication(node) {
     }
   }
 
+  // Step 2A
   status = distributeNthRoot(newNode);
   substeps.push(status);
   newNode = NodeStatus.resetChangeGroups(status.newNode);
 
+  // Step 2B
   status = evaluateNthRootForChildren(newNode);
   if (status.hasChanged()) {
     substeps.push(status);
     newNode = NodeStatus.resetChangeGroups(status.newNode);
 
+    // Step 2C
     status = combineRoots(newNode);
     if (status.hasChanged()) {
       substeps.push(status);
@@ -306,14 +316,12 @@ function convertMultiplicationToExponent(node) {
   let newNode = clone(node, false);
 
   const radicandNode = node.args[0];
-  const rootNode = getRootNode(node);
-  const rootValue = parseFloat(rootNode.value);
 
   if (NodeType.isParenthesis(radicandNode)) {
     const child = radicandNode.content;
     if (isMultiplicationOfEqualNodes(child)) {
       const baseNode = child.args[0];
-      const exponentNode = NodeCreator.constant(rootValue);
+      const exponentNode = NodeCreator.constant(child.args.length);
       newNode.args[0] = NodeCreator.operator('^', [baseNode, exponentNode]);
       return NodeStatus.nodeChanged(
         MathChangeTypes.CONVERT_MULTIPLICATION_TO_EXPONENT, node, newNode);
@@ -326,7 +334,7 @@ function convertMultiplicationToExponent(node) {
         const grandChild = child.content;
         if (isMultiplicationOfEqualNodes(grandChild)) {
           const baseNode = grandChild.args[0];
-          const exponentNode = NodeCreator.constant(rootValue);
+          const exponentNode = NodeCreator.constant(grandChild.args.length);
           children.push(NodeCreator.operator('^', [baseNode, exponentNode]));
           return;
         }
@@ -389,26 +397,28 @@ function evaluateNthRootForChildren(node) {
   }
 }
 
-// Given a multiplication nodes, including nthRoots, will combine the nodes
-// with the same radicand as a step
+// Given a multiplication node, with children including nthRoots, will combine
+// the nodes with the same radicand as a step
 // e.g. 2 * nthRoot(2) * nthRoot(x) -> 2 * nthRoot(2 * x)
+// Assumes that all the roots are the same (that this is occuring right
+// after distributeNthRoot and evaluateNthRootForChildren)
 function combineRoots(node) {
   let newNode = clone(node, false);
-  const rootNode = getRootNode(node);
 
+  let rootNode;
   const children = [];
   const radicandArgs = [];
   for (let i = 0; i < newNode.args.length; i++) {
     const child = newNode.args[i];
-    if (NodeType.isFunction(child, 'nthRoot') && getRootNode(child).equals(rootNode)) {
+    if (NodeType.isFunction(child, 'nthRoot')) {
       radicandArgs.push(child.args[0]);
+      rootNode = getRootNode(child);
     }
     else {
       children.push(child);
     }
   }
 
-  // if there are no items moved out of the root, then nothing has changed
   if (children.length > 0) {
     if (radicandArgs.length > 0) {
       const radicandNode = radicandArgs.length === 1 ?
@@ -423,6 +433,7 @@ function combineRoots(node) {
     }
   }
 
+  // if there are no items moved out of the root, then nothing has changed
   return NodeStatus.noChange(node);
 }
 
@@ -484,6 +495,10 @@ function nthRootConstant(node) {
 // The root node is the second child of the nthRoot node, but if one doesn't
 // exist, we assume it's a square root and return 2.
 function getRootNode(node) {
+  if (!NodeType.isFunction(node, 'nthRoot')) {
+    throw Error('Expected nthRoot');
+  }
+
   return node.args.length === 2 ? node.args[1] : NodeCreator.constant(2);
 }
 
--- a/lib/evaluateFunctions.js
+++ b/lib/evaluateFunctions.js
@@ -68,7 +68,7 @@ function evaluateFunctions(node) {
   return NodeStatus.noChange(node);
 }
 
-// Evaluate nthRoot() function
+// Evaluate nthRoot() function.
 // Returns a NodeStatus object.
 function nthRoot(node) {
   if (!NodeType.isFunction(node, 'nthRoot')) {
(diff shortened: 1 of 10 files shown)


lib/evaluateFunctions.js
```diff
@@ -65,6 +68,229 @@ function evaluateFunctions(node) {
...
+      }
+
+      return NodeStatus.nodeChanged(
+        MathChangeTypes.NTH_ROOT_VALUE, oldNode, newNode, true, subSteps);
+    }
+    else if (radicandNode.op === '^') {
+      const baseNode = radicandNode.args[0];
+      const exponentNode = radicandNode.args[1];
+      // TODO(ael): this only works if they actually have a value...
+      // ideally I'd use equals, but the nodes are different for some reason
```
evykassirer: equals() isn't working? or ===

I know that === doesn't work, but I'm surprised about .equals()

lib/evaluateFunctions.js
```diff
@@ -65,6 +68,229 @@ function evaluateFunctions(node) {
...
+  const radicandNode = newNode.args[0];
+  const rootNode = newNode.args.length === 2 ? newNode.args[1] : NodeCreator.constant(2);
+
+  if (NodeType.isOperator(radicandNode)) {
+    // if is operator and operation is addition, we can't do anything
+    if (radicandNode.op === '+' || radicandNode.op === '-') {
+      return NodeStatus.noChange(node);
+    }
+    else if (radicandNode.op === '*' || radicandNode.op === '/') {
+      // distribute the root into the children
```
evykassirer: I'm not sure why this is a step - some tests with substep examples and examples of math in the code comments would both help

evykassirer: because `sqrt(2*18)` is only possible if you have the next step `sqrt(2*2*3*3)`

lib/evaluateFunctions.js
```diff
@@ -65,6 +68,229 @@ function evaluateFunctions(node) {
...
+    else if (radicandNode.op === '^') {
+      const baseNode = radicandNode.args[0];
+      const exponentNode = radicandNode.args[1];
+      // TODO(ael): this only works if they actually have a value...
+      // ideally I'd use equals, but the nodes are different for some reason
+      // even if the value is the same...
+      if (rootNode.value === exponentNode.value) {
+        newNode = baseNode;
+        return NodeStatus.nodeChanged(
+          MathChangeTypes.NTH_ROOT_VALUE, oldNode, newNode);
```
evykassirer: I'd make a special change type for this! it's a fun edge case

lib/evaluateFunctions.js
```diff
@@ -65,6 +68,229 @@ function evaluateFunctions(node) {
...
+      } else {
+        return NodeStatus.noChange(node);
```
evykassirer: there are some other cases we'd want to catch to probably

mainly I'm thinking of exponent % rootnode value === 0

e.g. `sqrt(x^4)`

lib/evaluateFunctions.js
```diff
@@ -65,6 +68,229 @@ function evaluateFunctions(node) {
...
+      newNode = NodeCreator.constant(nthRootValue);
+      return NodeStatus.nodeChanged(
+        MathChangeTypes.NTH_ROOT_VALUE, oldNode, newNode);
+    }
+    else {
+      const subSteps = [];
+
+      // 1. convert the number into the product of its prime factors
+      const factors = Factor.getPrimeFactors(radicandValue);
+      const factorNodes = factors.map(factor => {
```
evykassirer: fun javascript neatness: you can do `factors.map(NodeCreator.constant)`

lib/evaluateFunctions.js
```diff
@@ -65,6 +68,229 @@ function evaluateFunctions(node) {
...
+        return NodeCreator.constant(factor);
+      })
+
+      newNode.args[0] = NodeCreator.operator('*', factorNodes);
+      subSteps.push(NodeStatus.nodeChanged(
+          MathChangeTypes.FACTOR_INTO_PRIMES, oldNode, newNode));
+
+      // TODO: Consider moving steps 2, 3 and 4 into distributing the
+      // radical into children above, and just calling nthNode on the
+      // factored version
```
evykassirer: not sure what you're talking about here

lib/evaluateFunctions.js
```diff
@@ -65,6 +68,229 @@ function evaluateFunctions(node) {
...
+          if (factors[i] !== factors[j]) {
+            break;
+          }
+          j++;
+        }
+        if (j - i >= rootValue) {
+          let groupEnd = i + rootValue;
+          groupedFactors.push(factors.slice(i, groupEnd));
+          hasGroups = true;
+          i = groupEnd;
```
evykassirer: since you reset i to groupEnd, you may as well break out of the while loop once j-i == rootValue (that might be off by one but you get the idea)

lib/evaluateFunctions.js
```diff
@@ -65,6 +68,229 @@ function evaluateFunctions(node) {
...
+        }
+        else {
+          groupedFactors = groupedFactors.concat(factors.slice(i, j));
```
evykassirer: consider having a list of factor group of the size we want, and a separate group of remaining factors we can put at the end?
or would you rather keep the factors all in the same order and just insert the parens

lib/evaluateFunctions.js
```diff
@@ -65,6 +68,229 @@ function evaluateFunctions(node) {
...
+      }
+
+      if (!hasGroups) {
+        return NodeStatus.noChange(node);
+      }
+
+      let children = [];
+      for (let i = 0; i < groupedFactors.length; i++) {
+        const item = groupedFactors[i];
+        if (groupedFactors[i].constructor === Array) {
```
evykassirer: can this happen even if there's more than one but less than the radicandValue?

lib/evaluateFunctions.js
```diff
@@ -65,6 +68,229 @@ function evaluateFunctions(node) {
...
+      // 3. convert grouped factors into exponent nodes
+      oldNode = NodeStatus.resetChangeGroups(newNode);
+      newNode = clone(oldNode, false);
+
+      for (let i = 0; i < children.length; i++) {
+        const child = children[i];
+        if (NodeType.isParenthesis(child)) {
+          const childNodeStatus = convertMultiplicationToExponent(child.content);
+          newNode.args[0].args[i] = childNodeStatus.newNode;
+          subSteps.push(NodeStatus.childChanged(newNode, childNodeStatus, i));
```
evykassirer: the childChanged also updates newNode if you want, but then you need to do a newNode = status.newNode after

lib/evaluateFunctions.js
evykassirer: ohh I see later why you do this
might not be obvious to the student at this stage. pedagogy details though, can always change later

lib/evaluateFunctions.js
```diff
@@ -65,6 +68,229 @@ function evaluateFunctions(node) {
...
+        }
+      }
+
+      // 4. move grouped factors outside of the radical
+      oldNode = NodeStatus.resetChangeGroups(newNode);
+      newNode = clone(oldNode, false);
+
+      const newArgs = [];
+      const newRadicandArgs = [];
```
evykassirer: confused about the difference between these two and why you always push to newRadicandArgs

comment examples would defs help

lib/evaluateFunctions.js
```diff
@@ -96,4 +99,389 @@ function absoluteValue(node) {
...
+    if (status.hasChanged()) {
+      substeps.push(status);
+      newNode = NodeStatus.resetChangeGroups(status.newNode);
+    }
+
+    return NodeStatus.nodeChanged(
+      MathChangeTypes.NTH_ROOT_VALUE, node, newNode, true, substeps);
+  }
+
+  return NodeStatus.noChange(node);
```
evykassirer: ohh okay so you put it back together in combineRoots, but that's not always called after the distribute right?

lib/evaluateFunctions.js
```diff
@@ -96,4 +99,389 @@ function absoluteValue(node) {
...
+// e.g. 2 * nthRoot(2) * nthRoot(x) -> 2 * nthRoot(2 * x)
+function combineRoots(node) {
+  let newNode = clone(node, false);
+  const rootNode = node.args.length === 2 ? node.args[1] : NodeCreator.constant(2);
+
+  const children = [];
+  const radicandArgs = [];
+  for (let i = 0; i < newNode.args.length; i++) {
+    const child = newNode.args[i];
+    if (NodeType.isFunction(child, 'nthRoot') && child.args[1].equals(rootNode)) {
```
evykassirer: what it doesn't have args[1] in the other case of  `node.args.length === 2 ? node.args[1] : NodeCreator.constant(2);`

evykassirer: maybe make a function for getting the rootNode of a function root node

lib/evaluateFunctions.js
```diff
@@ -96,4 +99,389 @@ function absoluteValue(node) {
...
+      MathChangeTypes.EVALUATE_DISTRIBUTED_NTH_ROOT, node, newNode, true, substeps);
+  }
+
+  return NodeStatus.noChange(node);
+}
+
+// Given a multiplication nodes, including nthRoots, will combine the nodes
+// with the same radicand as a step
+// e.g. 2 * nthRoot(2) * nthRoot(x) -> 2 * nthRoot(2 * x)
+function combineRoots(node) {
```
evykassirer: `e.g. 2 * nthRoot(2) * nthRoot(x) ->`

does that mean `node` is a multiplication of root nodes? or a single root node? some of this code seems like it's as if the former is true and some for the latter, though maybe I'm missing something

lib/evaluateFunctions.js
```diff
@@ -96,4 +99,389 @@ function absoluteValue(node) {
...
+// nthRootMultiplication on the new nthRoot
+function nthRootConstant(node) {
+  let newNode = clone(node, false);
+  const radicandNode = node.args[0];
+  const rootNode = node.args.length === 2 ? node.args[1] : NodeCreator.constant(2);
+
+  if (Negative.isNegative(radicandNode)) {
+    return NodeStatus.noChange(node);
+  }
+  else if (!NodeType.isConstant(rootNode) || Negative.isNegative(rootNode)) {
```
evykassirer: sqrt(-36) is still 6 * root(-1), so I think negative should be allowed and the -1 can be part of the prime factoring

also you should probably check integer and not just constant

lib/evaluateFunctions.js
```diff
@@ -96,4 +99,389 @@ function absoluteValue(node) {
   }
 }
 
+// Evaluate nthRoot() function.
+// Returns a NodeStatus object.
+function nthRoot(node) {
```
aelnaiem: Yea, I discuss re-organizing modules in the mathsteps organization doc, let's discuss that more

lib/evaluateFunctions.js
aelnaiem: This actually works fine because there's only steps if we call combine roots

evykassirer: ohhh nice :) can you add some tests for this case?

test/evaluateFunctions.test.js
```diff
@@ -14,6 +14,29 @@ function testEvaluateFunctions(exprStr, outputStr) {
       outputStr);
   });
 }
+describe('nthRoot', function () {
+  const tests = [
+    ['nthRoot(4)', '2'],
```
evykassirer: can you add tests with substeps?

lib/evaluateFunctions.js
```diff
@@ -96,4 +99,389 @@ function absoluteValue(node) {
...
+    }
+    else {
+      children.push(child);
+    }
+  }
+
+  if (children.length !== radicandNode.args.length) {
+    newNode.args[0] = NodeCreator.operator('*', children);
+    return NodeStatus.nodeChanged(
+      MathChangeTypes.FACTOR_INTO_PRIMES, node, newNode);
```
aelnaiem: Let's scrap it everywhere, as long as it doesn't change. In this case, the oldNode is always the original node since I separated each step into a different function. So as long as that's true, we should scrap it.

lib/evaluateFunctions.js
```diff
@@ -96,4 +99,389 @@ function absoluteValue(node) {
...
+// factoring all the arguments under the radicand
+function factorChildren(node) {
+  let newNode = clone(node, false);
+  const radicandNode = node.args[0];
+
+  let children = [];
+  for (let i = 0; i < radicandNode.args.length; i++) {
+    const child = radicandNode.args[i];
+
+    if (NodeType.isConstant(child) && !Negative.isNegative(child)) {
```
aelnaiem: This wouldn't actually happen in the stepper as it would have been simplified, I'm only really testing for negative constants because we don't support complex numbers: sqrt(-1)

lib/evaluateFunctions.js
```diff
@@ -96,4 +99,389 @@ function absoluteValue(node) {
...
+      children = children.concat(radicandNode.args.slice(i, j));
+    }
+    i = j;
+  }
+
+  // if we don't group any factors, then we can't simplify it any more
+  if (hasGroups) {
+    newNode.args[0] = children.length === 1 ?
+        children[0] : NodeCreator.operator('*', children);
+    return NodeStatus.nodeChanged(
```
aelnaiem: I see...

lib/evaluateFunctions.js
aelnaiem: In this case specifically it's a multiplication node, because we've distributed the roots into the children

lib/evaluateFunctions.js
```diff
@@ -96,4 +99,421 @@ function absoluteValue(node) {
...
+  }
+
+  return NodeStatus.noChange(node);
+}
+
+// Returns the nthRoot evaluated for a multiplication node.
+// Expects a multiplication node uder the radicand.
+// If the root is a positive constant, it factors the multiplicands, groups them
+// and converts the multiplications into exponents.
+// Distributes the nthRoot into the children nodes, evaluates those nthRoots
```
evykassirer: I think this could be easier to follow if the steps were numbered - find it a bit hard to follow still

lib/evaluateFunctions.js
```diff
@@ -162,15 +165,19 @@ function nthRootExponent(node) {
...
-// Expects a multiplication node.
+// Expects a multiplication node uder the radicand.
+// If the root is a positive constant, it factors the multiplicands, groups them
+// and converts the multiplications into exponents.
+// Distributes the nthRoot into the children nodes, evaluates those nthRoots
+// and combines them
 function nthRootMultiplication(node) {
   let newNode = clone(node, false);
-  const rootNode = node.args.length === 2 ? node.args[1] : NodeCreator.constant(2);
+  const rootNode = getRootNode(node);
```
evykassirer: love it

lib/evaluateFunctions.js
```diff
@@ -196,10 +203,8 @@ function nthRootMultiplication(node) {
   }
 
   status = distributeNthRoot(newNode);
-  if (status.hasChanged()) {
-    substeps.push(status);
-    newNode = NodeStatus.resetChangeGroups(status.newNode);
-  }
+  substeps.push(status);
```
evykassirer: what do you think about the `STEP X:` comments?

lib/evaluateFunctions.js
evykassirer: cool, noted in the overall restructure list

lib/evaluateFunctions.js
```diff
@@ -96,4 +99,389 @@ function absoluteValue(node) {
...
+    newNode.args[0] = NodeCreator.operator('*', children);
+    return NodeStatus.nodeChanged(
+      MathChangeTypes.FACTOR_INTO_PRIMES, node, newNode);
+  }
+
+  return NodeStatus.noChange(node);
+}
+
+// Given an nthRoot node with a constant positive root, will group the arguments
+// into groups of the root as a step
```
evykassirer: so it could group the arguments into groups of 2.3..?

and in general what happens for nthroot(x, y) where y isn't a whole number?

evykassirer: oo and what if it's 0? is a 0th root a thing?

lib/evaluateFunctions.js
evykassirer: did you go through all your changes and see if they need change groups? or are you leaving that for your other task

lib/evaluateFunctions.js
```diff
@@ -291,53 +289,50 @@ function groupTermsByRoot(node) {
...
       MathChangeTypes.GROUP_TERMS_BY_ROOT, node, newNode);
   }
+  // if we don't group any factors, then we can't simplify it any more
   return NodeStatus.noChange(node);
 }
 
-// Given an nthRoot node with a constant positive root, will group the arguments
+// Given an nthRoot node with a constant positive root,
 // will convert any grouped factors into exponent nodes as a step
 // e.g. nthRoot((2 * 2) * 2, 2) -> nthRoot(2^2 * 2, 2)
```
evykassirer: can you add a disclaimer that groups are expected to be the same length as the root number?

lib/evaluateFunctions.js
```diff
@@ -382,26 +377,30 @@ function evaluateNthRootForChildren(node) {
...
+  }
+  else {
     return NodeStatus.nodeChanged(
       MathChangeTypes.EVALUATE_DISTRIBUTED_NTH_ROOT, node, newNode, true, substeps);
   }
-
-  return NodeStatus.noChange(node);
 }
 
 // Given a multiplication nodes, including nthRoots, will combine the nodes
```
evykassirer: given a root node with multiplication nodes? or just multiplication nodes. still not super clear cause the first thing you do is getRootNode on the node..

lib/evaluateFunctions.js
```diff
@@ -382,26 +377,30 @@ function evaluateNthRootForChildren(node) {
...
 // with the same radicand as a step
 // e.g. 2 * nthRoot(2) * nthRoot(x) -> 2 * nthRoot(2 * x)
 function combineRoots(node) {
   let newNode = clone(node, false);
```
evykassirer: so here it seems that newNode is a root node

lib/evaluateFunctions.js
```diff
@@ -477,11 +478,42 @@ function nthRootConstant(node) {
...
-// Simple helper function which determines if all the given nodes are equal
-function allEqual(nodes) {
-  return nodes.reduce((a, b) => {
+// Helpers
+
+// Given an nthRoot node, will return the root node.
+// The root node is the second child of the nthRoot node, but if one doesn't
+// exist, we assume it's a square root and return 2.
+function getRootNode(node) {
+  return node.args.length === 2 ? node.args[1] : NodeCreator.constant(2);
```
evykassirer: can you make it throw an error if it's not a root function node?

lib/evaluateFunctions.js
aelnaiem: It doesn't, so I'm going to add that as future support

lib/evaluateFunctions.js
aelnaiem: Seems like an edge case check, so I'm going to leave that as a future issue

lib/evaluateFunctions.js
aelnaiem: Yea that's a mistake, it's not a root node. Fixing this now

PR #13, opened 2016-12-03
no plus minus step

evykassirer:
this was more headache than it's worth
from now on, we'll just always print + - as - and not support the + - => - step
some solvers wouldn't even support two operators side by side like that, so I think this is the better choice for the solver


aelnaiem:
👍

PR #14, opened 2016-12-06
fixing oldNode and newNode for substeps in NodeStatus childChanged function

aelnaiem:
This was assigning the substep to a `newNode`, rather than updating the substeps `newNode` and `oldNode`


lib/NodeStatus.js
```diff
@@ -79,28 +79,40 @@ NodeStatus.childChanged = function(node, childStatus, childArgIndex=null) {
     newNode.content = childStatus.newNode;
     oldNode.content = childStatus.oldNode;
     substeps.map(step => {
-      let newStep = clone(node, false);
-      newStep.content = step;
-      return newStep;
+      let newNode = clone(node, false);
```
evykassirer: nit: put old first? I wasn't consistent with it before but it'd be nice to do that

PR #16, opened 2016-12-06
remove unnecessary parens for substeps

evykassirer:


aelnaiem:
Right! Nice and simple 👍

PR #21, opened 2016-12-08
fixing some basic simplifications change groups

aelnaiem:


lib/simplifyExpression.js
```diff
@@ -67,13 +68,6 @@ function stepThrough(node, debug=false) {
     }
   }
 
-  // Update the last step, in case we returned something simplified the last
-  // time that didn't count as a step.
-  if (steps.length > 0) {
```
aelnaiem: This change actually breaks this test `subtraction support -------5 -> -5`, but this is a confusing step. Can you explain to me what's going on and why this is necessary?

evykassirer: the idea is nested unary minuses cancel out - that's what it was testing

not sure why it isn't counted as a step though - maybe it happens in flatten?

lib/PolynomialTermOperations.js
```diff
@@ -195,30 +195,29 @@ PolynomialTermOperations.multiplyLikeTerms = function(node) {
...
-      return newTermWithCeoff;
-    }
-    else {
-      return polyNode;
+
+      newNode.args[i].changeGroup = changeGroup;
+      node.args[i].changeGroup = changeGroup; // note that this is the "oldNode"
+
+      change = true;
+      changeGroup++;
```
evykassirer: i like that they're different colours, cool

evykassirer:
mostly looks good, just confused about how one section works

lib/EquationOperations.js
```diff
@@ -164,8 +164,8 @@ EquationOperations.isolateSymbolOnLeftSide = function(equation, symbolName) {
     else if (leftNode.op === '/') {
       // The non symbol term is always a fraction because it's the
       // coefficient of our symbol term.
```
evykassirer: what if the node is 2x / 3?

evykassirer:
👍

PR #22, opened 2016-12-10
making a separate clone function that is faster

aelnaiem:
[asana task](https://app.asana.com/0/201972293666219/221970758157618)

Just using shallow clone since deep clone is actually broken (adds parens). But it's a pretty simple function and is also faster than clone module.


lib/ConstantFraction.js
```diff
@@ -179,7 +170,7 @@ ConstantFraction.addConstantAndFraction = function(node) {
     return NodeStatus.noChange(node);
   }
 
-  let oldNode = clone(node, false);
+  let oldNode = clone(node);
```
evykassirer: I think you want to change the convention here to call this newNode

lib/ConstantFraction.js
```diff
@@ -245,31 +236,60 @@ ConstantFraction.addConstantAndFraction = function(node) {
     MathChangeTypes.SIMPLIFY_ARITHMETIC, node, newNode, true, substeps);
 };
 
-// Given a list of nodes `fractionNodes` that all have the same denominator,
-// add them together. e.g. 2/3 + 5/3 -> (2+5)/3
+// Given a node with a list of fraction nodes as args that all have the same
```
evykassirer: nit: clarify that it's an operation + node?

lib/ConstantFraction.js
```diff
@@ -245,31 +236,60 @@ ConstantFraction.addConstantAndFraction = function(node) {
...
-  return NodeCreator.operator('/', [newNumerator, commonDenominator]);
+
+  newNode = NodeCreator.operator('/', [newNumerator, commonDenominator]);
+  return NodeStatus.nodeChanged(
+    MathChangeTypes.COMBINE_NUMERATORS, node, newNode);
+}
+
+// Given a node with a numerator that is an addition node, will add
+// all the numerators and return the result
+function addNumeratorsTogether(node) {
```
evykassirer: I really like the way you restructured this module to be more clean :)

lib/clone.js
```diff
@@ -0,0 +1,18 @@
+'use strict';
+
+// Simple clone function, which creates a copy of the given node
+// And recurses on the children (due to the shallow nature of the copy)
```
evykassirer: think you could be more clear, e.g. "creates a deep copy of the given node" and "due to the shallow nature of the mathjs node clone"

lib/LikeTermCollector.js
```diff
@@ -116,7 +115,7 @@ LikeTermCollector.collectLikeTerms = function(node) {
     newOperands = newOperands.concat(terms[OTHER]);
   }
 
-  const newNode = clone(node, false);
+  const newNode = clone(node);
   newNode.args = newOperands;
   const oldNode = node;
```
evykassirer: remove this line and make oldNode just node

lib/PolynomialTermOperations.js
```diff
@@ -22,7 +21,7 @@ PolynomialTermOperations.rearrangeCoefficient = function(node) {
   if (!canRearrangeCoefficient(node)) {
     return NodeStatus.noChange(node);
   }
-  const oldNode = clone(node, false);
+  const oldNode = clone(node);
```
evykassirer: do you want to make the change so this is newNode in this module? or leave that for another time

lib/cancelLikeTerms.js
```diff
@@ -25,7 +25,7 @@ function cancelLikeTerms(node) {
   if (!NodeType.isOperator(node) || node.op !== '/') {
     return NodeStatus.noChange(node);
   }
-  const oldNode = clone(node, false);
+  const oldNode = clone(node);
   let newNode = node;
```
evykassirer: same thing here with newNode = clone(node) and not having an oldNode

evykassirer:
👍 
looks good! I left some comments about if you want to keep oldNode everywhere, since you were starting to change that places (do you want to change it everywhere else at the same time or leave that for later?)

PR #26, opened 2016-12-12
fix tests that rearrange coeff broke

evykassirer:


lib/evaluateFunctions.js
```diff
@@ -241,19 +242,33 @@ function factorMultiplicands(node) {
...
+      if (coeffNode) {
+        const factorNodes = getFactorNodes(coeffNode);
+        if (factorNodes.length > 1) {
+          factored = true;
+        }
+        children = children.concat(factorNodes);
+      }
+      children.push(polyTermNoCoeff);
     }
+
```
aelnaiem: Why this space?

aelnaiem:
One question, otherwise 👍

lib/evaluateFunctions.js
evykassirer: code was feeling crowded - no strong reason
if else should probably be kept together

PR #27, opened 2016-12-12
replacing change type text with more constants

aelnaiem:


evykassirer:
👍 (once we get the other one in)

lib/MathChangeTypes.js
```diff
@@ -1,76 +1,75 @@
 // The text to identify rules for each possible step that can be taken
 
 module.exports = {
-  NO_CHANGE: 'No change',
+  NO_CHANGE: 'NO_CHANGE',
```
evykassirer: can you add the examples you use in Athena here? would make it very helpful for people to know what they do

aelnaiem: Done :)

PR #55, opened 2016-12-14
rename dfs to tree search,and make a template for tree search

evykassirer:
😍

### DIFF
--- a/lib/TreeSearch.js
+++ b/lib/TreeSearch.js
@@ -1,0 +1,73 @@
+'use strict';
+
+const NodeStatus = require('./NodeStatus');
+const NodeType = require('./NodeType');
+
+const TreeSearch = {};
+
+// Returns a function that performs a preorder search on the tree for the given
+// simplifcation function
+TreeSearch.preOrder = function(simplificationFunction) {
+  return function (node) {
+    return search(simplificationFunction, node, true);
+  }
+}
+
+// Returns a function that performs a postorder search on the tree for the given
+// simplifcation function
+TreeSearch.postOrder = function(simplificationFunction) {
+  return function (node) {
+    return search(simplificationFunction, node, false);
+  }
+}
+
+// A helper function for performing a tree search with a function
+function search(simplificationFunction, node, preOrder) {
+  let status;
+
+  if (preOrder) {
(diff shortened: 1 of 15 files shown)


lib/TreeSearch.js
```diff
@@ -0,0 +1,73 @@
...
+// Returns a function that performs a postorder search on the tree for the given
+// simplifcation function
+TreeSearch.postOrder = function(simplificationFunction) {
+  return function (node) {
+    return search(simplificationFunction, node, false);
+  };
+};
+
+// A helper function for performing a tree search with a function
+function search(simplificationFunction, node, preOrder) {
```
aelnaiem: 😍

lib/collectAndCombine.js
```diff
@@ -166,7 +133,7 @@ function multiplyLikeTerms(node) {
     return status;
   }
 
-  status = Fraction.multiplyFractionsDFS(node);
+  status = Fraction.multiplyFractionsTreeSearch(node);
```
aelnaiem: Why not just call it multiplyFractions? All the other simplifications that are tree searches are just exported by the module so they are only referred to by their name e.g. when used collectAndCombineLikeTermsTreeSearch is actually just collectAndCombineLikeTerms

lib/print.js
```diff
@@ -10,25 +10,25 @@ const NodeType = require('./NodeType');
...
-function printDFS(node) {
+function printTreeTraveral(node) {
   if (PolynomialTermNode.isPolynomialTerm(node)) {
     const polyTerm = new PolynomialTermNode(node);
     // This is so we don't print 2/3 x^2 as 2 / 3x^2
     // Still print x/2 as x/2 and not 1/2 x though
     if (polyTerm.hasFractionCoeff() && node.op !== '/') {
       const coeffTerm = polyTerm.getCoeffNode();
-      const coeffStr = printDFS(coeffTerm);
+      const coeffStr = printTreeTraveral(coeffTerm);
```
aelnaiem: Traversal*

lib/collectAndCombine.js
evykassirer: I'll just change this in the next PR

Issue #39, opened 2016-12-13
1 to the power of anything is 1


evykassirer:
Add a new function to simplifyBasics, similar to the others there, to simplify 1^___ => 1

Now, technically 1^infinity is not 1, so we should probably only do this simplification if the exponent has no symbol terms (i.e. it would eventually simplify to a constant, which you can check with `MathResolveChecks.resolvesToConstant`)

evykassirer:
added in #69

PR #69, opened 2017-01-19
Add rule to simplify 1^x to 1

michaelmior:
Rule as suggested in #39.
Note that the test could probably be more robust since it just checks that `1^x` stays as `1^x`.
It probably should check that `1^x` is reduced to `1` if `x` is known to be constant but I didn't see an obvious way to do that in the test suite.


lib/simplifyExpression/basicsSearch/index.js
```diff
@@ -25,6 +26,8 @@ const SIMPLIFICATION_FUNCTIONS = [
   reduceExponentByZero,
   // Check for x^1 which should be reduced to x
   removeExponentByOne,
+  // Check for 1^x which should be reduced to 1
```
evykassirer: maybe add something to the end of this comment ..."if x can be simplified to a constant" to be extra clear about this case

lib/simplifyExpression/basicsSearch/index.js
```diff
@@ -85,6 +88,20 @@ function removeExponentByOne(node) {
...
 }
 
+// If `node` is of the form 1^x, reduces it to a node of the form 1.
+// Returns a Node.Status object.
+function removeExponentBaseOne(node) {
+  if (node.op === '^' &&                         // an exponent with
+      checks.resolvesToConstant(node.args[1]) && // a power not a symbol and
+      Node.Type.isConstant(node.args[0]) &&      // a constant base
+      node.args[0].value === '1') {              // of value 1
+    let newNode = clone(node.args[0]);
```
evykassirer: I need to eventually add this to the linter, and we don't do this right in a bunch of places in the codebase already heh, but this can be `const` instead of `let` cause you don't change it

lib/simplifyExpression/basicsSearch/index.js
```diff
@@ -85,6 +88,20 @@ function removeExponentByOne(node) {
   return Node.Status.noChange(node);
 }
 
+// If `node` is of the form 1^x, reduces it to a node of the form 1.
```
evykassirer: same here :)

test/simplifyExpression/basicsSearch/basicsSearch.test.js
```diff
@@ -28,6 +28,9 @@ describe('simplify basics', function () {
     ['2x*2*-1', '2x * 2 * -1'], // does not remove multiplication by -1
     // removeExponentByOne
     ['x^1', 'x'],
+    // removeExponentBaseOne
+    ['1^3', '1'],
```
evykassirer: can you add another test for if the exponent isn't a constant yet, but will eventually be? that's actually the best case for the use of this rule I think!

e.g. `1^(2 + 3 + 5/4 + 7 - 6/7)` would be simplified to 1 right away without dealing with those messy fractions :)

lib/simplifyExpression/basicsSearch/index.js
michaelmior: Fair :) I just did this because the other tests seem to follow this format.

michaelmior:
Thanks for the feedback! Updated to address your concerns.

evykassirer:
haha fair on the let / const thing - we really should just fix it everywhere so we're not asking for things we don't do ourselves heh

thanks for the quick changes! looks good to me :D

michaelmior:
Cool. Thanks for the merge!

Issue #77, opened 2017-01-20
Installing mathsteps is not straightforward


kovzol:
I tried several ways to install mathsteps on Ubuntu Linux 16.04.1 LTS with just partial success.

First method: "sudo apt install nodejs". It installs node 4.2.6 (which is quite old). Then by running "npm install mathsteps" a folder node_modules is created, it also includes the mathsteps/ folder. Now I tried to create a file with the example .js code with name test.js, but I get the following error when running it by "nodejs test.js":
```
/home/kovzol/workspace/node_modules/mathsteps/lib/simplifyExpression/index.js:6
function simplifyExpressionString(expressionString, debug=false) {
                                                         ^

SyntaxError: Unexpected token =
    at exports.runInThisContext (vm.js:53:16)
    at Module._compile (module.js:374:25)
    at Object.Module._extensions..js (module.js:417:10)
    at Module.load (module.js:344:32)
    at Function.Module._load (module.js:301:12)
    at Module.require (module.js:354:17)
    at require (internal/module.js:12:17)
    at Object.<anonymous> (/home/kovzol/workspace/node_modules/mathsteps/index.js:1:90)
    at Module._compile (module.js:410:26)
    at Object.Module._extensions..js (module.js:417:10)
```

Second method: I built mathsteps as described on the main page. Then I copied the lib folder over the folder where the first method created mathsteps' lib folder. But I got the same error.

Third method: I downloaded newer node versions from its website. The I started "nodejs test.js" by explicitly using the correct path to the different node versions. I got the following for node version 6.9.4:
```
/home/kovzol/workspace/mathsteps-test/test.js:4
  console.log(step.oldNode.toString());    // "2 x + 2 x + x + x"
                          ^

TypeError: Cannot read property 'toString' of null
    at steps.forEach.step (/home/kovzol/workspace/mathsteps-test/test.js:4:27)
    at Array.forEach (native)
    at Object.<anonymous> (/home/kovzol/workspace/mathsteps-test/test.js:3:7)
    at Module._compile (module.js:570:32)
    at Object.Module._extensions..js (module.js:579:10)
    at Module.load (module.js:487:32)
    at tryModuleLoad (module.js:446:12)
    at Function.Module._load (module.js:438:3)
    at Module.runMain (module.js:604:10)
    at run (bootstrap_node.js:394:7)
```
and also the same for node version 7.4.0. It seems that step.oldNode does not exists. If I remove that line, I can make the rest work:
```
[...]

PR #79, opened 2017-01-20
fix a bug where oldNode was null on every step

hmaurer:
Should fix https://github.com/socraticorg/mathsteps/issues/77. As far as I could see there was no point setting `oldNode` to `null` in `Status.reset`. The tests are still passing so I assume I didn't do something completely silly by removing it. Fingers crossed.

### DIFF
--- a/lib/simplifyExpression/stepThrough.js
+++ b/lib/simplifyExpression/stepThrough.js
@@ -2,6 +2,7 @@
 
 const checks = require('../checks');
 const Node = require('../node');
+const Status = require('../node/Status');
 
 const arithmeticSearch = require('./arithmeticSearch');
 const basicsSearch = require('./basicsSearch');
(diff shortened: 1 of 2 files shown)


evykassirer:
yeah, I think this would fix it. Thanks for figuring that out! 

I was trying to remember why the `reset()` function exists at all, to see if this change breaks anything, but it's only used in one place, so let's just scrap `reset()`?

So instead of this change, can you just delete the `reset` function, and then in `simplifyExpression/stepThrough.js`:

- delete the line `nodeStatus.reset();`
-  change `nodeStatus = step(nodeStatus.newNode);` into  

``` 
const nextNode = Status.resetChangeGroups(nodeStatus.newNode);
nodeStatus = step(nextNode);
```

(or something like that)

How's that sound?

hmaurer:
Yeah I think that sounds reasonable.

hmaurer:
Should be good to merge!

lib/simplifyExpression/stepThrough.js
```diff
@@ -44,8 +44,8 @@ function stepThrough(node, debug=false) {
       logSteps(nodeStatus);
     }
     steps.push(removeUnnecessaryParensInStep(nodeStatus));
-    nodeStatus.reset();
-    nodeStatus = step(nodeStatus.newNode);
+    const nextNode = Status.resetChangeGroups(nodeStatus.newNode);
```
evykassirer: merge your change to run tests+linter before committing ;D

you need to import `Status` here

hmaurer: Oops

evykassirer: np :) lemme know when it's ready!

hmaurer: Should be now! I had forgotten to push yesterday...

Issue #80, opened 2017-01-21
Use object literal property value shorthand in es6


cspanda:
Instead of doing something like this inside /index.js:

```javascript

const simplifyExpression = require('./lib/simplifyExpression');
const solveEquation = require('./lib/solveEquation');
const ChangeTypes = require('./lib/ChangeTypes');

module.exports = {
  simplifyExpression: simplifyExpression,
  solveEquation: solveEquation,
  ChangeTypes: ChangeTypes,
};
```
follow airbnb's [style guide](https://github.com/airbnb/javascript/blob/master/README.md#es6-object-concise) and use object literal property value shorthand in es6 so it becomes shorter and more readable. 

```javascript
const simplifyExpression = require('./lib/simplifyExpression');
const solveEquation = require('./lib/solveEquation');
const ChangeTypes = require('./lib/ChangeTypes');

module.exports = {
  simplifyExpression,
  solveEquation,
  ChangeTypes,
};
```

evykassirer:
Woo thanks for making an issue for this ^_^ 

A great addition to the 'small' label (which is good, cause we're running low on them)

evykassirer:
cool, this shorthand is also supported by eslint (which we use): http://eslint.org/docs/rules/object-shorthand

I'll add it to the linter error issue too ( #44 )

evykassirer:
done! thanks @cspanda -- so speeedy 🐰🐇

PR #81, opened 2017-01-21
use property value shorthand

cspanda:
Changed to use property shorthand notation to make shorter and more readable. This should address issue #80


evykassirer:
thanks for fixing!

Issue #44, opened 2016-12-13
Improve our linter


evykassirer:
here's a list of things I'd like to add to `.eslintrc.js`

note: feel free to take on this issue an only add some of the rules and not all of them.

(comment below if you have more suggestions!)

- [x] use `const` in every place it can be used
- [x] never use `==`
- [x] 2 spaces for indentation, no tabs
- [x] no trailing spaces at the end of lines
- [x] http://eslint.org/docs/rules/object-shorthand
- [ ] http://eslint.org/docs/rules/brace-style stroustrup
- [ ] alphabetize `require` statements at top of files
- [ ] make sure we also have the `'use strict'` line
- [x] only use single quotes, no double quotes
- [x] no lines ending in white space
- [ ] space after keywords http://eslint.org/docs/rules/keyword-spacing

biyasbasak:
I am working on this

evykassirer:
woo! thanks @biyasbasak

Raibaz:
Is it still open and assignable? I can add the missing rules if needed.

evykassirer:
yes please! thank you :)

Raibaz:
I opened PR #107 about this, but I have some doubts about the `'use strict'` line: if I enable the [strict rule](http://eslint.org/docs/rules/strict), linting gives errors on almost every file since `use strict' is unnecessary inside of modules`, since ES6 modules are strict by default.

How do you think we should handle this?

evykassirer:
oh! I didn't realize it's strict by default. Let's get rid of `"use strict";` in those files then. Thanks :D

evykassirer:
done!! until we find more rules we want to add haha.

but let's close this for now :)

thanks so much @biyasbasak and @Raibaz for making the mode more beautiful 😍

PR #78, opened 2017-01-20
add pre-commit hook to run tests & linter

hmaurer:
Tiny PR resolving #30. There seemed to be two eslint infractions in the codebase so I fixed those in the same commit. After cloning the repository one has to run npm run setup-hooks to symlink the pre-commit hook from scripts/git-hooks/pre-commit to .git/hooks/pre-commit and it should work from there.

PR #83, opened 2017-01-21
Added Linting rules

biyasbasak:
addresses #44 

1) use const in every place it can be used
2) never use ==
3) no trailing spaces at the end of lines
4) use object shorthand


evykassirer:
thanks so much @biyasbasak 

would you be up for also running the linter on the codebase and fixing all the places we don't follow these rules? (I know that might be a lot of grunt work, so we can get some other people on if you don't want to do it all)

biyasbasak:
@evykassirer actually I did fix some parts of the codebase to follow the rules but I thought it would be better to ask before making the commit, now that you have brought it up, I would fix  code in all the places and make the commit

evykassirer:
sounds good! Asking first works for me - helps us make sure we're on the same page.

Thanks for making the fixes! Looking forward to seeing our code get better 🌟

biyasbasak:
done but it would be great to add pre hooks to git for future linting. 

In the file /mathsteps/lib/solveEquation/stepThrough.js in line 263
` equationStatus.substeps.forEach(substep => logSteps);`
substep variable is never used but I left it the way it is

evykassirer:
See #78 for githooks :D 

The existing linter errors were fixed in that PR too :)

evykassirer:
ahh this is fantastic, two linter PRs merged in one day 😍 

thank you!

PR #88, opened 2017-01-22
Enhancement: Multiply denominators with terms

### LINKED ISSUE #51: Equations: multiply by x on both sides
Blocked on #137 and #139 because without cancelling working, we can't multiply both sides by the denominator and then cancel

---

e.g. `2/x = 1 --> 2/x * x =  1 * x`

Currently there's a function for this in `EquationOperations` called `removeSymbolFromDenominator` but it's not implemented yet.

It should work like this:
- **if** the left node of an equation is a fraction
- **and** the denominator of the fraction has `symbolName` in it (so the fraction could be `1/x` or `(3 + x) / (x^2 + 2)` etc.) 
- **then** multiply both sides by the denominator

arbylee:
Attempt to resolve #51 and implement removeSymbolFromDenominator.
This change also needed to enhance the cancelLikeTerms logic. Cancelling needed to account for denominators matching what was being multiplied. Otherwise if they didn't cancel out, it would continue trying to multiply the denominator across until the equation reaches the length limit. Let me know any feedback


evykassirer:
Looking sweet! Thanks so much for this big change and also all the little changes you're making to make the codebase cleaner 🌟 

I'm feeling a bit sick today, so I'll take a more thorough look early this week. I also hope to chat about this a bit with Ahmed, who wrote the original equations code.

lib/node/PolynomialTerm.js
```diff
@@ -179,4 +112,88 @@ function negativeCoefficient(node) {
   return node;
 }
 
+function parseNode(node, onlyImplicitMultiplication) {
```
evykassirer: can you add a docstring, now that it's its own function?

lib/simplifyExpression/fractionsSearch/cancelLikeTerms.js
```diff
@@ -20,6 +20,52 @@ class CancelOutStatus {
 // e.g. (2x^2 * 5) / 2x^2 => 5 / 1
 // Returns a Node.Status object
 function cancelLikeTerms(node) {
+  // Cancel out terms if the term being multiplied
+  // is the same as the denominator
+  if (Node.Type.isOperator(node) && node.op === '*') {
```
evykassirer: can you merge in master (now with #91) and change this to `isOperator(node, '*')`? thanks!

lib/simplifyExpression/fractionsSearch/cancelLikeTerms.js
```diff
@@ -20,6 +20,52 @@ class CancelOutStatus {
 // e.g. (2x^2 * 5) / 2x^2 => 5 / 1
 // Returns a Node.Status object
 function cancelLikeTerms(node) {
+  // Cancel out terms if the term being multiplied
```
evykassirer: I think this comment isn't super clear about what 'the term being multiplied' is
can you add an example?

lib/simplifyExpression/fractionsSearch/cancelLikeTerms.js
```diff
@@ -20,6 +20,52 @@ class CancelOutStatus {
...
+  // is the same as the denominator
+  if (Node.Type.isOperator(node) && node.op === '*') {
+    let newNode = clone(node);
+    let leftNode = newNode.args[0];
+    let rightNode = newNode.args[1];
```
evykassirer: what if the roles you gave left node and right node are reversed? e.g. the right node is a fraction or sum that cancels out with the left node?

also, how do you know there's only two things being multiplied by each other? there can be any number of terms in a multiplication

lib/simplifyExpression/fractionsSearch/cancelLikeTerms.js
```diff
@@ -20,6 +20,52 @@ class CancelOutStatus {
...
+        return Node.Status.nodeChanged(
+          ChangeTypes.CANCEL_TERMS, node, newNode);
+      }
+    }
+    // leftNode is the sum of multiple terms. Check if all terms
+    // have the same denominator and matches the rightNode
+    // being multiplied
+    if (Node.Type.isOperator(leftNode) && leftNode.op === '+') {
+      let allMatchingDenominators = true;
+      for (let i = 0; i < leftNode.args.length; i++) {
```
evykassirer: you could use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every here, I think

lib/simplifyExpression/fractionsSearch/cancelLikeTerms.js
evykassirer: I'm actually surprised you need this - I think that `multiplyFractions` will turn  `(2+x) * (3+x)/(2+x)` into `((2+x)(3+x))/(2+x)` which _should_ then just cancel out with the existing code... but if you had to add this stuff, there's probably something missing here that needs to be added. Not sure if what you have fills the hole quite right, but let's talk about it more!

arbylee: The code on master doesn't cancel out terms for `(2+x) * (3+x)/(2+x)`, which was what prompted these changes.  Currently it would distribute out the (2+x).  Without canceling, my changes caused an infinite loop by multiplying the denominator, distributing out, multiplying the denominator, etc.

Regarding the left node / right node positioning, I had made an assumption based on what would happen during equations steps, but you're right that it misses out on if an equation starts as `(2+x) * (3+x)/(2+x)`.  I'll work on resolving this in the next commit

From what I can tell that scenario wouldn't occur during equation solving though. Instead of multiplying by an inverted fraction, the steps would multiply by the denominator and then divide by the numerator

arbylee:
I'm actually starting to think that this enhancement is a bit premature.  There's a lot of functionality that I seem to need out of cancelLikeTerms that should probably be done first before working on the equation solving.  For multiplication across, these things need to be handled first
`(x+1)/(1-x) * (1-x)` -> `x + 1`
`(3/(1-x) + 3/(1+x)) * (1-x)` -> `6-3x`
`((x+1)/(1-x) + 3/(1-x)) * (1-x)` -> `x + 4`

I'm also not sure how the current approach would deal with factorization and canceling:
`(x^2 + 4)/(x-2)` -> `x+2`

I think I'll pull out some of the generic changes here out into a separate pull request, but I'm going to pause progress on this PR and maybe look into canceling terms separately when I have some time.  I feel like that's going to be a considerable beast on its own

evykassirer:
I honestly think we can merge a lot of the stuff in this PR :)

if in `simplifyExpression/stepThrough.js` you switch the order of  `multiplyFractionsSearch` and `breakUpNumeratorSearch`, `(x+1)/(1-x) * (1-x)` would become  `[(x+1) * (1-x)] / (1-x)` and then cancelling will work. This will then cover most of the simple cases - which is why I think it's worth making the change even though some other things won't work properly yet.

`(3/(1-x) + 3/(1+x)) * (1-x) -> 6-3x` and `((x+1)/(1-x) + 3/(1-x)) * (1-x) -> x + 4` could be fixed with a change to distribution where if (exactly) one of the blocks being distributed into each other has fractions, we take the other block and add it to the numerator (not sure if I'm describing this well enough, but I think that one case will fix that issue for the most part, as well as `((x+1)/(1-x) + 3/(1-x)) * (1-x)`)

(aside, `(3/(1-x) + 3/(1+x)) * (1-x) -> 6/(x+1)` but we'll have to wait for adding polynomial fractions and common denominators there and stuff to get that covered)

For factoring, we haven't done any factoring stuff yet but it's on the horizon! There's a lot of cases where your code would be helpful without needing factoring, and I think it's quite reasonable to add this functionality now and wait for factoring later.

tl;dr, I think we should:
- switch the order of  `multiplyFractionsSearch` and `breakUpNumeratorSearch` and remove your addition to the cancelling terms code
- merge this change!
- later, add distribution rules and factoring to handle some cancelling cases better
 

what do you think?

kevinbarabash:
`((x+1)/(1-x) + 3/(1-x)) * (1-x) -> x + 4` erases information about which values of `x` are invalid so the expressions are 100% equivalent.  It's probably be worth keeping track of the fact that `1 - x` can't be `0`.

arbylee:
Sure, I'll try flipping those methods tomorrow and see where I get

evykassirer:
@kevinbarabash very true. We don't have anything that stores information like that yet. Do you think adding it to the `NodeStatus` object would be a good way to do it?

kevinbarabash:
@evykassirer a new `NodeStatus` object is created for each step, right?  If that's the cas we'd probably want to copy any solution constraints to all subsequent steps so that it's easy to check.  Any new solution constraints would have to be unioned with previous ones.  The concept of solution constraints could also be applied to things like `nthRoot(x)` and other functions in the future.

evykassirer:
hey @arbylee - did you get a chance to try it? how'd it go? :)

arbylee:
Hey @evykassirer, I did try removing the cancelLikeTerms changes and flipping multiplyFractionsSearch and breakUpNumeratorSearch.  It ended up breaking a few test cases, and I haven't had enough time to look through them (I moved this past weekend).  I've updated the branch if you want to have a look

evykassirer:
thanks! I'll take a look and see what makes sense to fix here and what makes sense to fix in other issues

this change is a nice goldmine of revealing bugs ;D

test/solveEquation/solveEquation.test.js
```diff
@@ -58,6 +58,12 @@ describe('solveEquation for =', function () {
     ['2(x+3)/3 = 2', 'x = 0'],
     ['( u )/( 0.3) = 4u + 6.28', 'u = -9.42'],
     ['- q - 4.36= ( 2.2q )/( 1.8)', 'q = -1.962'],
+    //['2/x = 1', 'x = 2'], TODO: fixme
+    //['2/(4x) = 1', 'x = 1/2'], TODO: bombs out
+    //['2/(8 - 4x) = 1/2', 'x = 1'], TODO: fixme
+    //['2/(1 + 1 + 4x) = 1/3', 'x = 1'], TODO: fixme
+    //['(3 + x) / (x^2 + 3) = 1', '-x^2 + x = 0'], TODO: bombs out
```
evykassirer: this problem is different! it's a bad distribution

`(3 / (x^2 + 3) + x / (x^2 + 3)) * (x^2 + 3)` turns into
`(3 / (x^2 + 3) * x^2 + 9 / (x^2 + 3) + x / (x^2 + 3) * x^2 + 3x / (x^2 + 3))`

I think distribution should check to see if one of the multiplying things is also a list of fractions, and if only one of them is, then multiply the other one into its numerators. I'll make this a separate issue #138

evykassirer: eventually it'd be nice to not split the numerator like that, but I'm not sure when we'd want and when we wouldn't

test/solveEquation/solveEquation.test.js
```diff
@@ -58,6 +58,12 @@ describe('solveEquation for =', function () {
...
+    //['6/x + 8/2x = 10', '(-10x) + 4x^2 = -6'], TODO: fixme
```
evykassirer: this is an interesting case where    `x * 6/x` isn't multiplied to get `(x * 6) / x`, which would cancel out

one solution is just to have better cancelling code that doesn't need it to be a fraction.

another (easier for now) solution would be to add a case for a symbol times a symbol in the denominator (i.e. we don't want `9/2 * x` to multiply to `9x / 2`, but we would want `9/x * x` to multiply to `9`)

... anyways I'll open an issue  #139 for this too

(btw `8/2x` is parsed as `8/2 x` ie `4x`, not as `8 / 2x`)

evykassirer:
also @arbylee 

do you want some mathsteps stickers as a thank you for your contributions? 😸

img_20170218_173323

if you're interested, email mathsteps@socratic.org with your mailing address and I can send them over!

lib/Symbols.js
```diff
@@ -71,7 +101,17 @@ function isSymbolTerm(node, symbolName) {
       return true;
     }
   }
+  else if (hasDenominatorSymbol(node, symbolName)) {
```
arbylee: It's been a while now so I'm trying to remember.  I believe I had changed isSymbolTerm to return true if there was a symbol in the denominator, and there's logic that would then call getLastSymbolTerm.  And then getLastSymbolTerm relies on this isSymbolTerm function

arbylee:
@evykassirer Rebased and pushed with just the comment change for now.  I'll have a look at some of the issues raised to see where I can help.  And yes, I'd love some mathsteps stickers 😄

aelnaiem:
Hi @arbylee, I'd love to get this into master, what can I do to support you?

evykassirer:
there are some things blocking this change from being useful, around cancelling

for this code to work, we first need to fix: #137 #138 and #139

(see conversation above for details on that)

aelnaiem:
Great! So #138 is finished, and we only need #137 and #139 right? I'm thinking about how to denote high-priority issues. Those that block other things basically.

evykassirer:
I think a 'high priority' tag would do the trick!

I'd also use it for things that are blocked but we want to do soon

ldworkin:
So all blocking issues are taken care of now. @evykassirer how should we proceed? I could familiarize myself with this pull and try to fix the conflicts? Not sure if @arbylee is around to help?

evykassirer:
If this is urgent for Socratic you could take it over right away, but if you have stuff to work on and can give @arbylee until next week to see this and finish if they want, I think that'd be ideal. Your call though, I know this was a pretty big thing

ldworkin:
It's not urgent! I mean, needs to be done in the next month or so? Happy to wait for @arbylee!

arbylee:
Sorry I've been MIA.  Just did a quick rebase and uncommented out the previously failing test cases.  I'll carve out some time this weekend to review this again and try to resolve the remaining change requests.  Feel free to take this PR over if I move too slow on it though

evykassirer:
No worries! It only just got ready for ya

This weekend is definitely fine, seems like it only needs to be done within a month

test/solveEquation/solveEquation.test.js
```diff
@@ -58,6 +58,12 @@ describe('solveEquation for =', function () {
     ['2(x+3)/3 = 2', 'x = 0'],
```
arbylee: Updated

lib/Symbols.js
arbylee: Jumping back to this because it should probably be discussed (and have code/comments updated appropriately).  Prior to this change, `isSymbolTerm` specifically was checking only for **polynomial** symbol terms.  So a term like `1/x` would not be considered a symbol term because it's not a polynomial term.  This change makes it so that terms with a symbol in the denominator are considered symbol terms as well.

In turn, this affects `getLastNonSymbolTerm` and `getLastSymbolTerm` which both use `isSymbolTerm`.  I'm not sure if this change is going against the intentions of the Symbols object.  If it is, then I'll need to find some other way to get a symbol term in a denominator

evykassirer:
Wow so cool to see those tests passing! Will review soon 🎉

lib/Symbols.js
ldworkin: This makes sense to me! Agreed re: updating the comments to explain this

lib/Symbols.js
```diff
@@ -69,7 +95,17 @@ function isSymbolTerm(node, symbolName) {
...
+    return true;
+  }
   return false;
 }
 
+function hasDenominatorSymbol(node, symbolName) {
```
ldworkin: Add a quick docstring with examples here?

lib/simplifyExpression/stepThrough.js
```diff
@@ -84,6 +84,8 @@ function step(node) {
     breakUpNumeratorSearch,
     // e.g. 3/x * 2x/5 => (3 * 2x) / (x * 5)
     multiplyFractionsSearch,
+    // e.g. (2 + x) / 4 => 2/4 + x/4
+    breakUpNumeratorSearch,
```
ldworkin: This seems unrelated to the rest of the PR? What's the story here

ldworkin:
Thanks so much for finishing this up, @arbylee ! I just read over it, looks awesome, and I'm comfortable merging if Evy is. There are some merge conflicts now unfortunately, sorry about that -- let me know if you have any trouble resolving them.

ldworkin:
Here's a little snapshot of how it looks in the app!

img_1318

lib/simplifyExpression/stepThrough.js
arbylee: There was a previous suggestion from Evy to swap `breakUpNumberatorSearch` and `multiplyFractionSearch` before we discovered some of the issues blocking this PR.  This just looks like an oversight from merging master in.  I'll remove it

arbylee:
Merging in the latest master wasn't straightforward because some test cases were failing to even start.  I'll have a look further tonight/this week

ldworkin:
Ugh, no worries, I'm taking a look now. I'm not entirely sure if/how I can push to your branch though?

ldworkin:
@arbylee I was able to check out your branch and make some fixes, but I'm in detached-head state and don't know if/how I can push to it? Any ideas? I'm not a git expert

ldworkin:
Well I pushed a copy of the branch with my fix, haha, so you can at least see what I'm doing! 

Here's one fix: https://github.com/socraticorg/mathsteps/commit/6cd48dcb53594a907a06f71ce249daba6a23e3cd

ldworkin:
Ok well in my branch I got everything working except for one test case: `6/x + 8/2x = 10`. It "should" return `(-10x) + 4x^2 = -6` but instead returns no steps. I think this might have to do with @aliang8 's new "find roots" functionality -- because it looks like it actually gets to the step `-10x = -6 - 4x^2` and then `2x * (2x - 5) = -6`, at which point it fails. So the multiplying by both sides part works fine. Instead of returning no steps we should probably return as much simplification as we can? Not sure about this.

arbylee:
I've rebased and I think addressed the remaining change requests on this.  For the test case @ldworkin mentioned, I've only set it to `NO_STEPS` for now. I can't recall the context exactly (I think I just wanted to see how edge cases would behave in existing code), but I originally added that test in `Cleanup from feedback, add comments, and rebase` as `NO_STEPS` anyway.  I think further simplification of that particular case could be handled as a separate issue. Thoughts?

ldworkin:
Awesome, this looks great @arbylee. I'll make a separate note/issue to address that test case. @evykassirer thoughts on merging?!

evykassirer:
kevin mentioned earlier

> `((x+1)/(1-x) + 3/(1-x)) * (1-x) -> x + 4` erases information about which values of `x` are invalid so the expressions are 100% equivalent. It's probably be worth keeping track of the fact that `1 - x` can't be 0.

(see above for more context)

we should open an issue for this thing so we can address it once we make the NodeStatus more complicated and able to support constraints

lib/Symbols.js
```diff
@@ -61,14 +63,53 @@ Symbols.getLastNonSymbolTerm = function(node, symbolName) {
   return null;
 };
 
-// Returns if `node` is a polynomial term with symbol `symbolName`
+// Iterates through a node and returns the denominator if it has a
+// symbolName in its denominator
+// e.g. 1/(2x) with `symbolName=x` would return 2x
+// e.g. 1/(x+2) with `symbolName=x` would return x+2
+// e.g. 1/(x+2) + (x+1)/(2x+3) with `symbolName=x` would return 2x+3
```
evykassirer: I'm curious - does it return 2x+3 or does it wrap it in parens like (2x+3) ?

lib/Symbols.js
```diff
@@ -61,14 +63,53 @@ Symbols.getLastNonSymbolTerm = function(node, symbolName) {
...
+  }
+  return false;
+}
+
+// Return if `node` has a symbol in its denominator
+// e.g. true for 1/(2x)
+// e.g. false for 5x
+function hasDenominatorSymbol(node, symbolName) {
+  if (Node.Type.isOperator(node) && node.op === '/') {
+    const allSymbols = Symbols.getSymbolsInExpression(node);
```
evykassirer: I think this will return true for `x/2` too, so should this line be calling `Symbols.getSymbolsInExpression` only on the denominator?

lib/solveEquation/EquationOperations.js
```diff
@@ -45,12 +45,20 @@ EquationOperations.ensureSymbolInLeftNode = function(equation, symbolName) {
...
 
-// TODO: Ensures that a symbol is not in the denominator by multiplying
-// both sides by the whatever order of the symbol necessary.
-// This is blocked on the simplifying functionality of canceling symbols in
-// fractions (needs factoring for full canceling support)
-EquationOperations.removeSymbolFromDenominator = function(equation) {
-  // pass for now
+// Ensures that a symbol is not in the denominator by multiplying
+// both sides by the denominator if there is a symbol present.
+EquationOperations.removeSymbolFromDenominator = function(equation, symbolName) {
```
evykassirer: 🎉

lib/solveEquation/EquationOperations.js
```diff
@@ -45,12 +45,20 @@ EquationOperations.ensureSymbolInLeftNode = function(equation, symbolName) {
...
+  // Can't multiply a symbol across non-equal comparators
+  // because you don't know if it's negative and need to flip the sign
+  if (equation.comparator !== '=') {
+    return EquationStatus.noChange(equation);
+  }
+  const leftNode = equation.leftNode;
+  const denominator = Symbols.getLastDenominatorWithSymbolTerm(leftNode, symbolName);
+  if (denominator) {
+    return performTermOperationOnEquation(
```
evykassirer: I love how simple this change is thanks to the great work @aliang8 and @ldworkin have done 😍

test/Symbols.test.js
```diff
@@ -0,0 +1,40 @@
...
+describe('getLastSymbolTerm', function() {
+  const tests = [
+    ['1/x', '1 / x'],
+    ['1/(3x)', '1 / (3x)'],
+    ['3x', '3x'],
+    ['x + 3x', '3x'],
+    ['x/(x+3)', 'x / (x + 3)'],
+  ];
+
+  tests.forEach(t => runTest(Symbols.getLastSymbolTerm, t[0], t[1]));
```
evykassirer: can you pass 'x' in too, and then also have some tests where there are multiple symbols and you are looking for one of them? e.g. `'1/x + y', y`

test/Symbols.test.js
```diff
@@ -0,0 +1,40 @@
...
+});
+
+describe('getLastDenominatorWithSymbolTerm', function() {
+  const tests = [
+    ['1/x', 'x'],
+    ['1/(x+2)', '(x + 2)'],
+    ['1/(x+2) + 3x', '(x + 2)'],
+    ['1/(x+2) + 3x/(1+x)', '(1 + x)'],
```
evykassirer: can you add `2 + 2/x + x/2` to test that case I raised above?

test/solveEquation/solveEquation.test.js
```diff
@@ -88,6 +89,12 @@ describe('solveEquation for =', function () {
     ['x^2 (x - 5)^2 = 0', 'x = [0, 0, 5, 5]'],
     ['x^2 = 0', 'x = [0, 0]'],
     ['(x+2)^2 -x^2 = 4(x+1)', '4 = 4'],
+    ['2/x = 1', 'x = 2'],
+    ['2/(4x) = 1', 'x = 1/2'],
+    ['2/(8 - 4x) = 1/2', 'x = 1'],
+    ['2/(1 + 1 + 4x) = 1/3', 'x = 1'],
+    ['(3 + x) / (x^2 + 3) = 1', 'x = [0, 1]'],
+    ['6/x + 8/2x = 10', NO_STEPS],
```
evykassirer: can you add a TODO for this just so it's clear that this isn't what we want?

lib/Symbols.js
```diff
@@ -47,7 +47,9 @@ Symbols.getLastSymbolTerm = function(node, symbolName) {
 // e.g. 4x^2 + 2x + y with `symbolName=x` would return y
 Symbols.getLastNonSymbolTerm = function(node, symbolName) {
   if (isSymbolTerm(node, symbolName)) {
-    return new Node.PolynomialTerm(node).getCoeffNode();
+    if (Node.PolynomialTerm.isPolynomialTerm(node)) {
```
arbylee: That would actually change the behavior of the function, but I've cleaned it up to be more obvious

lib/Symbols.js
arbylee: It is wrapped in parens, updated

lib/Symbols.js
arbylee: You're right.  I've corrected that now

test/Symbols.test.js
arbylee: Added some test cases with other variables

test/Symbols.test.js
arbylee: Done.  There was even an existing issue with how symbols were being parsed that was made obvious with this particular test case.  I've added a commit to recursively parse the nodes for symbols.  It was previously only comparing `2 + 2/x` and `x/2` as the only two nodes, instead of checking `2` and `2/x` individually

test/solveEquation/solveEquation.test.js
arbylee: I realized that the test I wanted from this was actually missing parens in `8/(2x)`.  I've added that and adjusted the test result.

I didn't leave the previous test case without parens because it started to look similar to the test cases in the other TODO related to complex numbers.  I can add it back with a comment if you think it's a separate scenario

ldworkin:
So excited for this, thanks for all your work, @arbylee !!

ldworkin:
Dammit, this change https://github.com/socraticorg/mathsteps/commit/a21240da059d734bec446813d6e45aca64512401 broke something! Investigating, will try to fix.

ldworkin:
Ok, I ended up reverting that change. I need to revisit the logic there. So this should be good -- not sure why it says there are still conflicts/failing checks, I think if we merge the latest master into it, it will be all good.

test/Symbols.test.js
evykassirer: yay! glad we caught that :D

test/solveEquation/solveEquation.test.js
evykassirer: @ldworkin how do you want to track this? You said "I'll make a separate note/issue to address that test case."

is it in your Asana board? or can we open an issue here? should we just put the TODO in the code? I'm fine as long as it's being noted somewhere :)

ldworkin: It's in our Asana

lib/Symbols.js
```diff
@@ -76,10 +83,13 @@ Symbols.getLastDenominatorWithSymbolTerm = function(node, symbolName) {
   // Otherwise, it's a sum of terms. e.g. 1/x + 1(2+x)
   // Look through the operands for a
   // denominator term with `symbolName`
-  else if (Node.Type.isOperator(node) && node.op === '+') {
+  else if (Node.Type.isOperator(node, '+')) {
```
evykassirer: ✨

Issue #64, opened 2016-12-30
figure out what to do about errors from rounding midway through


evykassirer:
so it's good pedagogically to round partway through, but then we can get the wrong answer in the end!

e.g. this gives us 6.399 when it should actually be 6.4 :(

`x - 3.4 = ( x - 2.5)/( 1.3)`

steps (approximately, to give an idea of why this happens):
- `x - 3.4 = x/1.3 - 2.5/1.3`
- `x - 3.4 = x/1.3 - 1.9231` (maybe we wouldn't divide here, but how would we make that call easily?)
- ...
- `(1 - 1/1.3) x = 3.4 - 1.9231`
- ...
- `0.2308x = 1.4769`
- ...
- `x = 6.399`

I'm not really sure what the best pedagogical solution is! having really long numbers in the middle wouldn't really be great. What would we want to show the user?

kevinbarabash:
@evykassirer great project!

In this case you could multiply both sides by 1.3, but that's not a very general solution.  You probably want to use a bignum library, there are a couple written in JS.  I forgot which one I used, but it stored decimals as fractions which should get around this problem.  There's still the question of what to display to students.  Maybe you could use ellipsis when presenting decimals that continue, or, even better, a bar over the repeating part.  In this case 1.9230769230769 the 230769 repeats.

evykassirer:
thanks Kevin! :)

oo the bar or ellipses sounds like a nice solution. 

Although I think the reason we started rounding in the first place was floating point errors - that's what the bignum library is for right? 

Storing decimals as fractions miiiight not work for all cases - e.g. square roots giving irrational numbers (but I think right now it keeps it as a square root and doesn't calculate it)

kevinbarabash:
> Although I think the reason we started rounding in the first place was floating point errors - that's what the bignum library is for right?

Right.

> right now it keeps it as a square root and doesn't calculate it

Excellent!

hmaurer:
Hello! Throwing some thoughts in: ellipsis are an elegant solution if the goal is to obtain an exact result, but sometimes when doing simple arithmetic we only care about an approximate result. As @evykassirer  mentioned in those cases it makes sense to round midway.

Maybe at that point the system could mark all subsequent nodes as "approximate", so that instead of displaying equality signs to the user we may display "≈" signs.

An extra-awesome feature (which would be particularly useful when doing physics-related arithmetic) would be to keep track of the error-margin for each step. Probably outside the scope of this project for now though.

[...]

PR #89, opened 2017-01-22
Use fraction numbers to avoid rounding (wip)

kevinbarabash:
My first attempt to address https://github.com/socraticorg/mathsteps/issues/64.

Output from test.js to solve `x - 3.4 = ( x - 2.5)/( 1.3)`:
```
x - 3.4 = 0.(769230)x - 1.(923076)
(x - 3.4) - 0.(769230)x = (0.(769230)x - 1.(923076)) - 0.(769230)x
0.(230769)x - 3.4 = (0.(769230)x - 1.(923076)) - 0.(769230)x
0.(230769)x - 3.4 = -1.(923076)
(0.(230769)x - 3.4) + 3.4 = -1.(923076) + 3.4
0.(230769)x = -1.(923076) + 3.4
0.(230769)x = 1.4(769230)
(0.(230769)x) / 0.(230769) = 1.4(769230)/0.(230769)
x = 1.4(769230)/0.(230769)
x = 6.4
```

The numbers in extra brackets represent the repeating section of the decimal.  There's still lots to do, but this gives me confidence that it's doable.

### DIFF
--- a/lib/equation/Equation.js
+++ b/lib/equation/Equation.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const math = require('mathjs');
+math.config({number: 'Fraction'});
 
 const clone = require('../util/clone');
 const printNode = require('../util/print');
--- a/lib/simplifyExpression/arithmeticSearch/index.js
+++ b/lib/simplifyExpression/arithmeticSearch/index.js
@@ -22,7 +22,7 @@ function arithmetic(node) {
   // we want to eval each arg so unary minuses around constant nodes become
   // constant nodes with negative values
   node.args.forEach((arg, i) => {
-    node.args[i] = Node.Creator.constant(arg.eval());
+    node.args[i] = Node.Creator.constant(arg.eval().toString(), "number");
   });
 
   // Only resolve division of integers if we get an integer result.
@@ -31,7 +31,7 @@ function arithmetic(node) {
     const numeratorValue = parseInt(node.args[0]);
     const denominatorValue = parseInt(node.args[1]);
     if (numeratorValue % denominatorValue === 0) {
-      const newNode = Node.Creator.constant(numeratorValue/denominatorValue);
+      const newNode = Node.Creator.constant(numeratorValue, denominatorValue);
       return Node.Status.nodeChanged(
         ChangeTypes.SIMPLIFY_ARITHMETIC, node, newNode);
     }
@@ -41,7 +41,7 @@ function arithmetic(node) {
   }
   else {
     const evaluatedValue = evaluateAndRound(node);
-    const newNode = Node.Creator.constant(evaluatedValue);
+    const newNode = Node.Creator.constant(evaluatedValue.toString(), "number");
     return Node.Status.nodeChanged(ChangeTypes.SIMPLIFY_ARITHMETIC, node, newNode);
   }
 }
@@ -49,14 +49,7 @@ function arithmetic(node) {
 // Evaluates a math expression to a constant, e.g. 3+4 -> 7 and rounds if
 // necessary
 function evaluateAndRound(node) {
-  let result = node.eval();
-  if (result < 1) {
-    result  = parseFloat(result.toPrecision(4));
-  }
-  else {
-    result  = parseFloat(result.toFixed(4));
-  }
-  return result;
+  return node.eval();
 }
 
 module.exports = search;
--- a/lib/simplifyExpression/fractionsSearch/divideByGCD.js
+++ b/lib/simplifyExpression/fractionsSearch/divideByGCD.js
@@ -1,6 +1,8 @@
 'use strict';
 
 const math = require('mathjs');
+math.config({number: 'Fraction'});
+
 const ChangeTypes = require('../../ChangeTypes');
 const Node = require('../../node');
 
@@ -22,8 +24,8 @@ function divideByGCD(fraction) {
     return Node.Status.noChange(fraction);
   }
 
-  const numeratorValue = parseInt(fraction.args[0].eval());
-  const denominatorValue = parseInt(fraction.args[1].eval());
+  const numeratorValue = fraction.args[0].eval();
+  const denominatorValue = fraction.args[1].eval();
 
   // The gcd is what we're dividing the numerator and denominator by.
   let gcd = math.gcd(numeratorValue, denominatorValue);
--- a/lib/solveEquation/stepThrough.js
+++ b/lib/solveEquation/stepThrough.js
@@ -58,8 +58,11 @@ function stepThrough(leftNode, rightNode, comparator, debug=false) {
     steps = addSimplificationSteps(steps, equation, debug);
     if (steps.length > 0) {
       const lastStep = steps[steps.length - 1];
-      equation = Equation.createEquationFromString(
-        lastStep.newEquation.print(), equation.comparator);
+      equation.leftNode = lastStep.newEquation.leftNode.clone();
+      equation.rightNode = lastStep.newEquation.rightNode.clone();
+      // console.log(left);
+      // equation = Equation.createEquationFromString(
+      //   lastStep.newEquation.print(), equation.comparator);
     }
 
     equation.leftNode = flattenOperands(equation.leftNode);
--- a/lib/node/Creator.js
+++ b/lib/node/Creator.js
@@ -5,6 +5,8 @@
 */
 
 const math = require('mathjs');
+math.config({number: 'Fraction'});
+
 const NodeType = require('./Type');
 
 const NodeCreator = {
@@ -33,8 +35,8 @@ const NodeCreator = {
       '-', 'unaryMinus', [content]);
   },
 
-  constant (val) {
-    return new math.expression.node.ConstantNode(val);
+  constant (val, valType) {
+    return new math.expression.node.ConstantNode(val, valType);
   },
 
   symbol (name) {
--- a/lib/simplifyExpression/fractionsSearch/addConstantFractions.js
+++ b/lib/simplifyExpression/fractionsSearch/addConstantFractions.js
@@ -1,6 +1,8 @@
 'use strict';
 
 const math = require('mathjs');
+math.config({number: 'Fraction'});
+
 const clone = require('../../util/clone');
 const divideByGCD = require('./divideByGCD');
 const ChangeTypes = require('../../ChangeTypes');
--- a/lib/simplifyExpression/functionsSearch/absoluteValue.js
+++ b/lib/simplifyExpression/functionsSearch/absoluteValue.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const math = require('mathjs');
+math.config({number: 'Fraction'});
 
 const clone = require('../../util/clone');
 const ChangeTypes = require('../../ChangeTypes');
--- a/lib/simplifyExpression/functionsSearch/nthRoot.js
+++ b/lib/simplifyExpression/functionsSearch/nthRoot.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const math = require('mathjs');
+math.config({number: 'Fraction'});
 
 const clone = require('../../util/clone');
 const ConstantFactors = require('../../factor/ConstantFactors');
--- a/lib/simplifyExpression/index.js
+++ b/lib/simplifyExpression/index.js
@@ -1,6 +1,8 @@
 'use strict';
 
 const math = require('mathjs');
+math.config({number: 'Fraction'});
+
 const stepThrough = require('./stepThrough');
 
 function simplifyExpressionString(expressionString, debug=false) {
--- a/lib/simplifyExpression/simplify.js
+++ b/lib/simplifyExpression/simplify.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const math = require('mathjs');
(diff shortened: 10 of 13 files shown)


lib/simplifyExpression/arithmeticSearch/index.js
```diff
@@ -31,7 +31,7 @@ function arithmetic(node) {
     const numeratorValue = parseInt(node.args[0]);
     const denominatorValue = parseInt(node.args[1]);
     if (numeratorValue % denominatorValue === 0) {
-      const newNode = Node.Creator.constant(numeratorValue/denominatorValue);
+      const newNode = Node.Creator.constant(numeratorValue, denominatorValue);
```
kevinbarabash: whoops, I meant to put a `math.divide` call in there.

lib/simplifyExpression/arithmeticSearch/index.js
```diff
@@ -41,22 +41,15 @@ function arithmetic(node) {
...
 function evaluateAndRound(node) {
-  let result = node.eval();
-  if (result < 1) {
-    result  = parseFloat(result.toPrecision(4));
-  }
-  else {
-    result  = parseFloat(result.toFixed(4));
-  }
-  return result;
+  return node.eval();
```
kevinbarabash: This is where the options check should go.

evykassirer:
I'll take a look early this week - today I've been a bit sick and now am finishing up some homework ^_^

thanks for working on this - soo cool!

lib/simplifyExpression/fractionsSearch/divideByGCD.js
```diff
@@ -22,8 +24,8 @@ function divideByGCD(fraction) {
     return Node.Status.noChange(fraction);
   }
 
-  const numeratorValue = parseInt(fraction.args[0].eval());
-  const denominatorValue = parseInt(fraction.args[1].eval());
+  const numeratorValue = fraction.args[0].eval();
+  const denominatorValue = fraction.args[1].eval();
```
kevinbarabash: These both evaluate to a fraction object now.

lib/solveEquation/stepThrough.js
```diff
@@ -58,8 +58,11 @@ function stepThrough(leftNode, rightNode, comparator, debug=false) {
     steps = addSimplificationSteps(steps, equation, debug);
     if (steps.length > 0) {
       const lastStep = steps[steps.length - 1];
-      equation = Equation.createEquationFromString(
-        lastStep.newEquation.print(), equation.comparator);
+      equation.leftNode = lastStep.newEquation.leftNode.clone();
+      equation.rightNode = lastStep.newEquation.rightNode.clone();
+      // console.log(left);
+      // equation = Equation.createEquationFromString(
+      //   lastStep.newEquation.print(), equation.comparator);
```
kevinbarabash: Calling `math.config({number: 'Fraction'})` in every file was not enough to ensure using the same fractions at each step because we were printing out the equation at each step and re-parsing.  Unfortunately, print output repeating fractions as `'0.(3)'`, but mathjs can only parse repeating fractions if we print them out as `'fraction("0.(3)")', see https://github.com/josdejong/mathjs/issues/650.  Cloning the left/right nodes avoids the issue and should be faster b/c we skip the re-parsing.

I should create a new `equation` object as opposed to overwriting the `leftNode` and `rightNode` properties.

kevinbarabash: This is kind of a big change.  I'm not sure how much anyone relies on being able to re-parse the output of `print` beyond what was being done in `stepThrough`.  I could see having some kind of serialization/deserialization method being useful though.

kevinbarabash:
TODO:
- [ ] add Options param
- [ ] make sure test suite is passing
- [ ] fix division bug in lib/simplifyExpression/arithmeticSearch/index.js
- [ ] update `_toTex` method to display repeated part of decimal with `\overline`
- [ ] verify that (de)serialization of math.js data structures still work

I was thinking about repeat bar vs. ellipses and ellipses makes more sense when truncating output of irrational numbers.

lib/equation/Equation.js
```diff
@@ -1,6 +1,7 @@
 'use strict';
 
 const math = require('mathjs');
+math.config({number: 'Fraction'});
```
evykassirer: I wonder if there's a way to just set this globally (cause you're just setting config whenever you `require` `mathjs` right?

lib/simplifyExpression/arithmeticSearch/index.js
```diff
@@ -22,7 +22,7 @@ function arithmetic(node) {
   // we want to eval each arg so unary minuses around constant nodes become
   // constant nodes with negative values
   node.args.forEach((arg, i) => {
-    node.args[i] = Node.Creator.constant(arg.eval());
+    node.args[i] = Node.Creator.constant(arg.eval().toString(), "number");
```
evykassirer: Could you set it up so that the string "number" isn't created here? to avoid typos and stuff

Maybe we could have an enum somewhere... or could create multiple Node.Creator constructors for each case? idk

Also I see that you always set it to "number". Would we ever want it to be something else?

lib/simplifyExpression/arithmeticSearch/index.js
```diff
@@ -41,22 +41,15 @@ function arithmetic(node) {
...
+    const newNode = Node.Creator.constant(evaluatedValue.toString(), "number");
     return Node.Status.nodeChanged(ChangeTypes.SIMPLIFY_ARITHMETIC, node, newNode);
   }
 }
 
 // Evaluates a math expression to a constant, e.g. 3+4 -> 7 and rounds if
 // necessary
 function evaluateAndRound(node) {
-  let result = node.eval();
-  if (result < 1) {
```
evykassirer: would you keep this for the option where we do round as we go?

lib/simplifyExpression/fractionsSearch/divideByGCD.js
evykassirer: the numerator and denominators are both integers that we can call `gcd` on though, yeah?

lib/solveEquation/stepThrough.js
evykassirer: yes I think cloning is best - good call. glad it makes it faster too!

I think cloning resets changeGroups, but if it doesn't we can manually add that too.

I think it's reasonable to have `print()` not be able to be re-parsed, especially if there's an option to turn off the thing that makes it unable to be re-parsed (ie, you can have it print with `...` and repeating fractions iff you're okay with not being able to re-parse)

can't you serialize/deserialize with JSONifying the struct? or is that not what you meant

kevinbarabash:
I tried to make some more progress here, but the back-and-forth conversions to strings by `ContentNode` is really getting me down.  Some fractions result have really long repeaters, 600/791 has a repeater that's over 100 characters long.  It would be much better if we could keep everything a fraction object (or a complex object) all the time and just call methods on those objects.  

Unfortunately, `mathjs` likes to convert everything to a string and then convert it back to object during `eval`.  Before `eval` happens it calls `_compile` on each node which converts it to a different string depending on the number config (BigNumber vs Fraction).

I'd like to make a suggestion: replace `mathjs`.  The main parts of `mathjs` that I see we're using are the parser and the eval code.  Both of these are relatively easy to replace considering most of the heavy lifting for evaluation would be done by `fraction.js` anyways.

Benefits:
- more easily change which numbers we're computing with, e.g. fractions, complex numbers
- have a more regular AST, right now we have EquationNode but we're bound to add more as we go
- avoid re-parsing things over and over
- have a parser that produces a flattened tree by design
- smaller bundle size

evykassirer:
I agree the string thing is pretty annoying, and makes more messy code around the codebase (would some helper functions make it a bit less gross?)

In general I think your suggestion is quite reasonable - having something more custom to our project would be really nice. The one thing is that we've been building a relationship with mathjs, and they mention our project on [their website](http://mathjs.org/download.html). Their community is pretty big and pretty cool, and it'd be nice to keep that relationship with them. Not sure how to balance that with the benefits of creating our own custom parser. 

Thoughts?

evykassirer:
cc @aelnaiem

aelnaiem:
Good point and insight into how we use mathjs. Do you have an alternative parser in mind or would would we need to create one first?

kevinbarabash:
I have a parser that I'd written a while ago for creating a flattened math AST similar to what mathsteps uses.  It was for a hackathon project to allow users to manipulate algebraic expressions/equations so that they could solve them step-by-step instead of just typing in the answer.

I've made some changes recently to bring the AST closer in line with what mathjs produces.  It needs a little work, mainly tests.  It doesn't cover everything that mathjs' parser does, but I think it covers everything we'd need for now except for absolute value.  I'll add that shortly.

https://github.com/kevinbarabash/math-parser

kevinbarabash:
I've added support for absolute value and inequalities to `math-parser`.

aelnaiem:
That's really great Kevin! I think the parsing abstraction is pretty contained if I remember correctly, so it should be pretty straight-forward to swap the parser and see what breaks. We would also want to replace the eval? I particularly appreciate that your parser deals with equations. Very cool 😄

kevinbarabash:
Handling eval is next on my list.  I'd like to avoid building `eval` methods into the nodes though and would rather have a separate function that evaluates an AST.

aelnaiem:
Agreed that makes the most sense. Let me know how I can help

kevinbarabash:
I spent some time last night better specifying the AST that the parser should generate.  It diverges in places from the current AST, but I think it sets us up to be able to cover more math in a more regular way.  I'll add a function that converts the new AST to an AST that aligns better with mathsteps' current AST to make the transition easier.

Please have a look at https://github.com/kevinbarabash/math-parser/issues/12.  I'd love to get your thoughts.  

@aelnaiem I could use some help verifying that the AST (after conversion) matches what mathsteps is currently using as well as actually swapping out the parser and eval in mathsteps when they're ready.  Also, if there's any other work that needs to be on the parser or eval function, please add it to https://github.com/kevinbarabash/math-parser/issues.

aelnaiem:
Agreed about representing more math in a regular way, it's clear that it's covering a lot of the representations that would be useful for mathsteps at least, and for other uses in the future. I think divergence would make sense especially as we would want to support things like sequences and relations within a single structure. 

I also love that the tree would be flattened by default. Curious how you think I can best help. I recently started adding a [goals and priorities doc](https://github.com/socraticorg/mathsteps/wiki/Goals-and-Priorities) which aims to capture the kinds of improvements we want to make. A project around improving the parsing and eval in this way could be something we add there, particularly as it could be a pre-requisite to so many other improvements.

One other improvement that I'm excited about making in terms of architecture is re-visiting NodeStatus and EquationStatus and coming up with something simpler that is consistent to the different types of problems that we'll want to provide step-by-step assistance for. Overall, I'm excited about the opportunities this would present for making the mathsteps code simpler and less bug-prone. 

My major concern is by not using mathjs, we would lose out on the improvements that community will make to their parser as well.

kevinbarabash:
> Curious how you think I can best help.

Swapping out mathjs' expression tree for math-parser's AST.  I have a rough plan for how that might happen:
  - write a function to convert the math-parser AST to match mathjs' expression tree (depends on https://github.com/kevinbarabash/math-parser/issues/8)
  - verify that both match for all expressions that we care about, maybe write a function that traverse either nodes from either tree and produces an s-expression string which can be easily compared
  - update helper functions to handle either type of node
  - replace method calls to .eval(), with an evaluate function that can accept either node
  - swap out the parser

I appreciate any help in this process as I'm very familiar with the mathsteps code base.

> A project around improving the parsing and eval in this way could be something we add there, particularly as it could be a pre-requisite to so many other improvements.

Sounds good to me.

> re-visiting NodeStatus and EquationStatus and coming up with something simpler that is consistent

+1

> My major concern is by not using mathjs, we would lose out on the improvements that community will make to their parser as well.

Even in the current state, there are things mathjs can parse that I think we'd eventually find useful.  There is, however, a mismatch between the use cases.  I think the benefits we get from the new AST outweigh the cost of evolving the parser to handle new syntax.

kevinbarabash:
I made some progress last night converting the math-parser AST to something very similar to the mathjs expression tree.  It doesn't have any of the methods like `eval` on the nodes though.  Here's the output from https://github.com/kevinbarabash/math-parser/blob/f7975c81381da885d7105730410ed7f09c1d1a52/demo.js:
```
original :    --2
printed  :    --2
expr     :    0.5 + atan2(1, 1) + 2 * 3 - 4 / 5
result   :    6.4853981633974485

math-parser AST
{ type: 'Operation',
  op: '-',
  args: [ { type: 'Number', value: '-2' } ] }

mathjs expression tree
{ type: 'OperatorNode',
  op: '-',
  fn: 'unaryMinus',
  args: [ { type: 'ConstantNode', value: '-2', valueType: 'number' } ] }

transformed math-parser AST with unary minuses applied to number nodes
{ type: 'Number', value: 2 }
```
The demo also shows that other transforms are possible on the AST by applying an Operation node with a unary minus to the Number node containing `'-2'`.

Sorry about the messiness of demo.js.  I'll try to get some tests up over the next couple of days.

aelnaiem:
Let's create a new issue for supporting this parser so others can track progress and we can organize what's needed for testing out the requirements.

evykassirer:
done, #128

Issue #31, opened 2016-12-12
Add an optional parameter to isOperator to check the operation type


evykassirer:
there are some places in the codebase where you'll see

`if (NodeType.isOperator(node) && node.op === '+') { `

it would be nice if `isOperator` took an optional argument to check op at the same time, so we could change the above snippet to

`if (NodeType.isOperator(node, '+') {`

and then any `if (NodeType.isOperator(node) && node.op === ... ` can be replaced in the codebase

mcarthurgill:
I've made these changes, but I believe I need you to assign this issue to me. Thanks!

evykassirer:
Thanks! :tada:  

Hmmm either I need to be given higher permissions for this repo (I didn't create it) or I might not be able to assign you at all... looking into it :)

In the mean time consider this conversation to show that you've 'claimed' this issue, feel free to open a PR with your change and I'll take a look

evykassirer:
@mcarthurgill you said you've made the changes - can you open a pull request so they can be merged?

mcarthurgill:
Done! Let me know if you have questions.

evykassirer:
Thanks @mcarthurgill ! 

Feel free to take a look at other open issues, and lemme know if I can help with anything.

PR #91, opened 2017-01-23
Refactor isOperator to accept operator parameter

mcarthurgill:
addresses #31 

Added parameter to accept a String operator into isOperator (defaulted to null). Refactored isOperator usage throughout app where applicable. Adequate tests added as well.

### DIFF
--- a/lib/Symbols.js
+++ b/lib/Symbols.js
@@ -30,7 +30,7 @@ Symbols.getLastSymbolTerm = function(node, symbolName) {
   }
   // Otherwise, it's a sum of terms. Look through the operands for a term
   // with `symbolName`
-  else if (Node.Type.isOperator(node) && node.op === '+') {
+  else if (Node.Type.isOperator(node, '+')) {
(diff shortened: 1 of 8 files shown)


evykassirer:
this is sweet! I've been dreaming of this function for a while, nice to see it existing.

was just looking through for anywhere else it could make sense to use it...

in likeTermCollector `if (!Node.Type.isOperator(node) || !(node.op === '+' || node.op === '*')) ` can just be two calls to isOperator now (I think) which seems cleaner. What do you think?

otherwise looks 👌 👌

mcarthurgill:
Pushed the change. Let me know if that's not what you were thinking.

evykassirer:
yup! that's what I was thinking, much cleaner.

Thanks :)

evykassirer:
hey @mcarthurgill 

I was just wondering if you'd like me to mail you some mathsteps stickers we just got in

img_20170218_173323

if you're interested, email mathsteps@socratic.org with your mailing address and I'll send 'em off to you!

Issue #82, opened 2017-01-21
make it clear that mathsteps requires node 6.0.0 


evykassirer:
`mathsteps` won't work with lower versions of node (e.g. #77), so we should add this requirement to the README

ideally this bullet point would link to a good resource for installing/updating node

emars:
If this is a recurring issue, something like:

http://stackoverflow.com/a/41620850/2816843

could be implemented.

Having a `preinstall` check that verifies that the user is on the proper version of node.

evykassirer:
oh, sweet! 

that looks very helpful and I totally think we should add it, thanks :)

Raibaz:
Nice one @emars :)

evykassirer:
script done! thanks to @Raibaz and to @emars for the suggestion

evykassirer:
there's an issue where if you `npm install mathsteps` there's an error `Error: Cannot find module 'semver'
`

I don't have time to look into it now, but hopefully we can fix it soon. 

@aelnaiem ?

Raibaz:
I think this is because `preinstall` is run before dependencies are installed...perhaps moving `npm run requirements-check` to `install` or `postinstall` would fix it?

evykassirer:
yeah I think that's the issue too.. it's odd because both the stackoverflow answer and medium post user `preinstall` but then dependencies aren't installed yet :p so maybe they're wrong!

I'm also looking around to try to find ways to test installing it again without patching a new version. Let me know if you know of anything

aelnaiem:
It seems that node-semver should be installed with npm according to this:
https://docs.npmjs.com/files/package.json

aelnaiem:
Enforcing the engine [was deprecated](https://github.com/npm/npm/blob/master/CHANGELOG.md#enginestrict) as part of an effort to shift control from module authors to application authors. Curious if this suggests that the best practice is to just notify application authors of the node version requirement in the README, and they can also see the requirement in the package.json

evykassirer:
Sounds OK to me - too bad though :(

evykassirer:
should be fixed now :)

PR #94, opened 2017-01-24
Added a script to check the installed node version

Raibaz:
This should solve #82 with [this](http://stackoverflow.com/questions/29349684/how-can-i-specify-the-required-node-js-version-in-packages-json/41620850#41620850) solution proposed in the issue comments.

A script has been added in the `scripts` folder to check the current node version against the one required in `package.json` and the script execution has been bound to the `preinstall` stage in the package file itself.

Unfortunately this adds a dependency from `semver`, but I guess there is no easier way to handle version comparison than adding an external dependency.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,6 +41,7 @@
   "homepage": "https://github.com/socraticorg/mathsteps#readme",
   "scripts": {
     "requirements-check": "node scripts/check-version.js",
-    "preinstall": "npm run requirements-check"
+    "preinstall": "npm run requirements-check",
+    "test": "mocha --recursive"
   }
 }
--- a/package.json
+++ b/package.json
@@ -4,7 +4,11 @@
   "description": "Step by step math solutions",
   "main": "index.js",
   "dependencies": {
-    "mathjs": "3.8.0"
+    "mathjs": "3.8.0",
+    "semver": "^5.3.0"
+  },
+  "engines": {
+    "node": ">=6.0.0"
   },
   "devDependencies": {
     "eslint": "^3.10.2",
@@ -34,5 +38,9 @@
   "bugs": {
     "url": "https://github.com/socraticorg/mathsteps/issues"
   },
-  "homepage": "https://github.com/socraticorg/mathsteps#readme"
+  "homepage": "https://github.com/socraticorg/mathsteps#readme",
+  "scripts": {
+    "requirements-check": "node scripts/check-version.js",
+    "preinstall": "npm run requirements-check"
+  }
 }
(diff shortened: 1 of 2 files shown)


Raibaz:
Note that i had to add `"test": "mocha --recursive"` to explicitly define the test suite runner script, since having a `scripts` property in `package.json` overrides the defaults.

scripts/check-version.js
```diff
@@ -0,0 +1,10 @@
+const semver = require('semver');
+const engines = require('../package').engines;
+
+const version = engines.node;
+if (!semver.satisfies(process.version, version)) {
+  console.log(`Required node version ${version} not satisfied with current version ${process.version}.`);
+  process.exit(1);
+} else {
```
evykassirer: I added this to #44, but can you put the else on a new line?

````  process.exit(1)
}
else {

File: scripts/check-version.js [hunk 2]
```diff
@@ -0,0 +1,10 @@
+const semver = require('semver');
+const engines = require('../package').engines;
+
+const version = engines.node;
+if (!semver.satisfies(process.version, version)) {
+  console.log(`Required node version ${version} not satisfied with current version ${process.version}.`);
```
evykassirer: 80 char max per line please :)

package.json
```diff
@@ -34,5 +38,10 @@
   "bugs": {
     "url": "https://github.com/socraticorg/mathsteps/issues"
   },
-  "homepage": "https://github.com/socraticorg/mathsteps#readme"
+  "homepage": "https://github.com/socraticorg/mathsteps#readme",
+  "scripts": {
+    "requirements-check": "node scripts/check-version.js",
+    "preinstall": "npm run requirements-check",
+    "test": "mocha --recursive"
```
evykassirer: you don't actually need to add this line :)

just add the ` "requirements-check"` and `"preinstall"` to the scripts starting on line 18

scripts/check-version.js
```diff
@@ -0,0 +1,10 @@
+const semver = require('semver');
+const engines = require('../package').engines;
+
+const version = engines.node;
+if (!semver.satisfies(process.version, version)) {
```
evykassirer: run `npm lint` and you'll see there are some errors

put `/*global process*/` at the top of this file to get rid of one

and then `// eslint-disable-next-line` above the `console.log` lines

scripts/check-version.js
```diff
@@ -0,0 +1,10 @@
+const semver = require('semver');
```
evykassirer: put `'use strict';` at the top of this file

Raibaz:
Whops i didn't notice `package.json` already had a scripts section and I copypasted the stackoverflow solution without linting it...sorry :(

Should be ok now.

evykassirer:
sweet! thanks for the quick fix

Issue #29, opened 2016-12-12
Don't allow dividing by 0


evykassirer:
We don't check for this, which means division by 0 gives weird results.

This functionality would probably go in simplifyBasics, or in a separate DFS before that - since checking for dividing by zero is important before anything else is attempted.

I think (though feel free to approach this a different way) that dividing by zero would look like this:
- the divide by 0 function returns a `NodeStatus.hasChanged` object where `newNode` is actually the same as `oldNode` and the `changeType` is `DIVIDE_BY_ZERO`
- in` simplifyExpression`, in the function `stepThrough`, after each call to `step`, check if the changeType was `DIVIDE_BY_ZERO` -- if so, break out of the while loop

hmaurer:
I am on this; will PR soon. How should we handle the `0/0` case? Return `DIVIDE_BY_ZERO` too?

evykassirer:
yeah, I think I'm comfy having them give the same error. Sound ok to you?

hmaurer:
Yes I think that makes sense! I'll PR later tonight.

evykassirer:
Hey @hmaurer, I see you have a branch open for this and a commit - did you forget to put up a PR or did you want to work on it a bit more first?

evykassirer:
@hmaurer ?

hmaurer:
@evykassirer Sorry for the delay; I have opened a PR

evykassirer:
no problem at all, just wanted to make sure it wasn't forgotten about ^_^

aelnaiem:
This is interesting, I wonder if there should be a NodeStatus for errors like this in general that we check, not just divide by zero.

evykassirer:
@aelnaiem I was thinking about that too!

In the PR, I talked about:
> oh, maybe a nodeStatus should have both a changeType and an errorType (default to null)? idk, still trying to figure out what this would look like long term

what do you think?

aelnaiem:
Yea I think that's a good idea!

evykassirer:
**UPDATE**: now related to #129 -- let's use it as the first error that we implement with the new error structure

I removed the small label, since this is more complicated now

PR #97, opened 2017-01-25
Handle division by 0

hmaurer:
addresses #29 

This PR adds a basic search strategy which generates a `DIVISION_BY_ZERO` step when encountering either `*/0` or `0/0`. In the main loop (`simplifyExpression/stepThrough.js`) we then stop the simplification process if such a step is generated (since division by zero counts as a form of error).

### DIFF
--- a/lib/simplifyExpression/basicsSearch/index.js
+++ b/lib/simplifyExpression/basicsSearch/index.js
@@ -9,6 +9,7 @@ const Node = require('../../node');
 const TreeSearch = require('../../TreeSearch');
 
 const reduceMultiplicationByZero = require('./reduceMultiplicationByZero');
+const handleDivisionByZero = require('./handleDivisionByZero');
 const reduceZeroDividedByAnything = require('./reduceZeroDividedByAnything');
 const reduceExponentByZero = require('./reduceExponentByZero');
 const removeExponentByOne = require('./removeExponentByOne');
@@ -23,6 +24,8 @@ const rearrangeCoefficient = require('./rearrangeCoefficient');
 const SIMPLIFICATION_FUNCTIONS = [
   // multiplication by 0 yields 0
   reduceMultiplicationByZero,
+  // division by zero should fail
+  handleDivisionByZero,
   // division of 0 by something yields 0
   reduceZeroDividedByAnything,
   // ____^0 --> 1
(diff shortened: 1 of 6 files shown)


lib/simplifyExpression/basicsSearch/index.js
```diff
@@ -23,6 +24,8 @@ const rearrangeCoefficient = require('./rearrangeCoefficient');
 const SIMPLIFICATION_FUNCTIONS = [
   // multiplication by 0 yields 0
   reduceMultiplicationByZero,
+  // division by zero should fail
```
evykassirer: I think error checks should actually go in a different tree search. Can you make a folder `errorsSearch` and put `handleDivisionByZero` in that tree search instead? And that search would happen first, before anything else

lib/simplifyExpression/stepThrough.js
```diff
@@ -45,6 +46,10 @@ function stepThrough(node, debug=false) {
       logSteps(nodeStatus);
     }
     steps.push(removeUnnecessaryParensInStep(nodeStatus));
+    // Stop on division by zero
+    if (nodeStatus.changeType === ChangeTypes.DIVISION_BY_ZERO) {
```
evykassirer: this would maybe come later, but eventually this will be a check to see if the changeType is an error

(oh, maybe a nodeStatus should have both a changeType and an errorType (default to null)? idk, still trying to figure out what this would look like long term)

test/simplifyExpression/basicsSearch/handleDivisionByZero.test.js
```diff
@@ -0,0 +1,12 @@
+'use strict';
```
evykassirer: idk how helpful these tests are - in every logic flow (error or not error) we still return the same node unchanged. I think just having the oneStep tests is good. Or you could put those in this file

test/simplifyExpression/oneStep.test.js
```diff
@@ -96,3 +96,15 @@ describe('simplifyDoubleUnaryMinus step actually happens', function () {
     assert.equal(steps[0].changeType, ChangeTypes.RESOLVE_DOUBLE_MINUS);
   });
 });
+
+describe('division by zero', function () {
```
evykassirer: can you name this 'division by zero error' just to be super clear when tests are running?

evykassirer:
hey @hmaurer how's this going?

evykassirer:
@hmaurer wanna close this and start from scratch, or just update this PR?

hmaurer:
Let's close it for now.

Issue #102, opened 2017-01-27
README.md : Add complete code example for solving an equation


karuppiah7890:

karuppiah7890:
Fixed in #103

PR #103, opened 2017-01-27
Add a complete code example for solving an equation in README (#102)

karuppiah7890:


README.md
```diff
@@ -25,7 +25,14 @@ steps.forEach(step => {
 
 To solve an equation:
 ```js
-const steps = mathsteps.solveEquation('2x + 2x + x + x');
+const steps = mathsteps.solveEquation('2x + 3x = 35');
+
+steps.forEach(step => {
+    console.log("before change: " + step.oldEquation.print());  // e.g. before change: 2x + 3x = 35
```
evykassirer: can you put the comment above the line of code so the lines are max 80 characters?

thanks :)

evykassirer:
looks good! just one comment

README.md
karuppiah7890: Oh. Now I see why you put the comments above the lines! Sorry!

karuppiah7890: @evykassirer But the earlier example for `simplifyExpression` also has comments on the same lin as the code and it leads to 92 characters in line 19!

evykassirer: Ooo true.. I should follow my own rules more haha.

What do you think? Is it a lot more readable on the same line (so we should make an exception) or is better to put them all on the line above? I could go either way :)

karuppiah7890: Uh, it's more readable on the same line. But then, later, if we keep doing this, we might end up bringing scroll bars in the code block, and people have to scroll horizontally to see the end of the comment.

evykassirer: Okay, let's keep it the way it is *if* it doesn't scroll horizontally

karuppiah7890: Haha, okay! In that case, we gotta keep some count on how much character the README can take in before it goes for the next line so we don't put that many characters on a single line of code!

karuppiah7890: I think it can take 100 characters with ease without bringing in scroll bars. And a few more characters, from what I can see. Not more. And my code has only one line which has 100 characters and that's the max

Issue #50, opened 2016-12-14
Factoring support for quadratics


evykassirer:
We don't have any factoring polynomial support yet, but that's our next big goal!

We want to make another top level module like `simplifyExpression` and `solveEquation` that factors an expression (note that factoring is sort of the opposite of simplifying, because simplifying distributes and we would want to "un-distribute")

#### We should be able to factor quadratics in the following ways:

- note: we should also explain step by step how to solve it
- note: feel free to only implement one of these in your pull request

sum/product rules
- e.g.` x^2+5x+6`
- Ask: Which two numbers add up to `5` and multiply to `6`? Answer: `2` and `3`
- Final answer: `(x+2)(x+3)`

perfect square
- e.g. `x^2+2x+1`
- rewrite in the form a^2 + 2b + b^2 where a = x and b = 1
- use "square of sum rule" `(a+b)^2 = a^2 + 2ab + b^2` to factor into `(x+1)^2`

quadratic equation
- your favourite: `x = [-b +or- sqrt(b^2 - 4ac) ] / 2a`

sangwinc:
Perhaps you might like to also consider "completing the square" as a method for solving quadratics?

x^2+2*a*x +c = 0,
x*(x+2*a) +c=0, 
(x+a-a)*(x+a+a)+c=0, 
(x+a)^2-a^2+c=0
(x+a)^2=a^2-c
... etc.

I've only just come across your system (and I'm impressed: email to follow).  

Does your system provide more than one solution, or does it always just give "the way" to solve an equation?

evykassirer:
as far as I remember, completing the square tends to be used for graphing, and isn't as useful for factoring (which is what this particular issue is working on, and would be used for `simplifyExpression` and `solveEquation`)

I could totally see another function alongside `solveEquation` that completes the square for quadratics and/or gives steps for graphing a function. Feel free to open an issue for that if you'd like to see it be worked on :)

evykassirer:
unassigned, since the initial pull request went through

we still need to add quadratic equation, and also expose factoring to users of the mathsteps module

aelnaiem:
[...]

PR #104, opened 2017-01-27
adding factoring support

aelnaiem:
Issue: https://github.com/socraticorg/mathsteps/issues/50

Adds support for factoring by:
* difference of squares
* perfect squares
* sum product rule

Quadratic equation to be added in the future. The reason I didn't add it yet is because it requires simplification of only what will be in the brackets (without expansion), and I didn't want to make a decision on how that should happen.


lib/node/Creator.js
```diff
@@ -53,7 +53,7 @@ const NodeCreator = {
     if (exponent) {
       polyTerm = this.operator('^', [polyTerm, exponent]);
     }
-    if (coeff) {
+    if (coeff && coeff.value !== '1') {
```
aelnaiem: This breaks some tests. Considering just doing this check before calling this. What do you think @evykassirer

evykassirer:
I think I'll get to this tomorrow :) this week is a bit busy for me and I think this isn't too urgent, ye?

aelnaiem:
Not urgent

lib/checks/isQuadratic.js
```diff
@@ -0,0 +1,45 @@
+'use strict';
+
+const Node = require('../node');
+
+// TODO(ael)
```
evykassirer: docstring

lib/checks/isQuadratic.js
```diff
@@ -0,0 +1,45 @@
+'use strict';
```
evykassirer: can you merge in the new linter changes? apparently we don't need `'use strict'` in modules because they're strict by default (the linter will catch this now 😄  )

lib/checks/isQuadratic.js
```diff
@@ -0,0 +1,45 @@
...
+
+const Node = require('../node');
+
+// TODO(ael)
+function isQuadratic(node) {
+  if (!Node.Type.isOperator(node) || node.op !== '+') {
```
evykassirer: you can now do `if (!Node.Type.isOperator(node, '+')` thanks to a recent change :)

lib/checks/isQuadratic.js
```diff
@@ -0,0 +1,45 @@
...
+  if (node.args.length > 3) {
+    return false;
+  }
+
+  const secondDegreeTerms = node.args.filter(isSecondDegree);
+  const firstDegreeTerms = node.args.filter(isFirstDegree);
+  const constantTerms = node.args.filter(Node.Type.isConstant);
+
+  if (secondDegreeTerms.length !== 1 || firstDegreeTerms.length > 1 ||
+    constantTerms.length !== 1) {
```
evykassirer: hmm why'd you choose this combination?

I agree that there should be a secondDegreeTerm for sure

but what about `x^2 + x`? that seems like a quadratic, but there's no constant term

lib/checks/isQuadratic.js
```diff
@@ -0,0 +1,45 @@
...
+    return false;
+  }
+
+  if (node.args.length > 3) {
+    return false;
+  }
+
+  const secondDegreeTerms = node.args.filter(isSecondDegree);
+  const firstDegreeTerms = node.args.filter(isFirstDegree);
+  const constantTerms = node.args.filter(Node.Type.isConstant);
```
evykassirer: what if there are other args that don't fall into one of these buckets? we should return false right? so I guess check that the length of each of these add up to the number of args

lib/checks/isQuadratic.js
```diff
@@ -0,0 +1,45 @@
...
+function isFirstDegree(node) {
+  if (Node.PolynomialTerm.isPolynomialTerm(node)) {
+    const polyTerm = new Node.PolynomialTerm(node);
+    const exponent = polyTerm.getExponentNode(true);
+    return exponent && exponent.value === '1';
+  }
+  return false;
+}
+
+function isSecondDegree(node) {
```
evykassirer: could you just have a general function `isPolynomialTermOfDegree(node, degree)` to avoid duplicated code?

test/checks/isQuadratic.test.js
```diff
@@ -0,0 +1,33 @@
+'use strict';
+
+const assert = require('assert');
+const math = require('mathjs');
+
+const flatten = require('../../lib/util/flattenOperands');
+const checks = require('../../lib/checks');
+
+function testIsQuadratic(exprString, result) {
```
evykassirer: I just merged an update to add `testUtil`, so these tests should be of the new format now

test/checks/isQuadratic.test.js
```diff
@@ -0,0 +1,33 @@
...
+    ['x^2 - 4', true],
+    ['x^2 + 2x + 1', true],
+    ['x^2 - 2x + 1', true],
+    ['x^2 + 3x + 2', true],
+    ['x^2 - 3x + 2', true],
+    ['x^2 + x - 2', true],
+    ['x^2 + 4', true],
+    ['x^2 + 4x + 1', true],
+    ['x^2', false],
+    ['x^3 + x^2 + x + 1', false],
```
evykassirer: add tests for `x^2 + x` and `x^2 + 4 + 2^x` and other edge cases I brought up

lib/factor/ConstantFactors.js
```diff
@@ -43,8 +43,11 @@ ConstantFactors.getPrimeFactors = function(number){
 ConstantFactors.getFactorPairs = function(number){
   const factors = [];
 
-  const root = Math.sqrt(number);
-  for (var divisor = 1; divisor <= root; divisor++) {
+  const bound = Math.floor(Math.sqrt(Math.abs(number)));
+  for (var divisor = -bound; divisor <= bound; divisor++) {
```
evykassirer: it'd be nice if we could pass in an argument of if we want negative factors or not - cause there'll be cases where we won't want them right?

lib/factor/factorQuadratic.js
```diff
@@ -0,0 +1,127 @@
...
+const checks = require('../checks');
+const ConstantFactors = require('./ConstantFactors');
+const ChangeTypes = require('../ChangeTypes');
+const flatten = require('../../lib/util/flattenOperands');
+const Negative = require('../Negative');
+const Node = require('../node');
+
+// Given a node, will check if it's in the form of a quadratic equation
+// `ax^2 + bx + c`, and
+// if it is, will factor it using on of the following rules:
```
evykassirer: one of

lib/factor/factorQuadratic.js
```diff
@@ -0,0 +1,127 @@
...
+//    - Difference of squares e.g. x^2 - 4 -> (x+2)(x-2)
+//    - Perfect square e.g. x^2 + 2x + 1 -> (x+1)^2
+//    - Sum/product rule e.g. x^2 + 3x + 2 -> (x+1)(x+2)
```
evykassirer: love these examples 😍

lib/factor/factorQuadratic.js
```diff
@@ -0,0 +1,127 @@
...
+    }
+  }
+
+  // TODO: quadratic formula
+  // a(x - (-b + sqrt(b^2 - 4ac)) / 2a)(x - (-b - sqrt(b^2 - 4ac)) / 2a)
+
+  return Node.Status.noChange(node);
+}
+
+module.exports = factorQuadratic;
```
evykassirer: this functionality won't exposed in mathsteps yet, right?

should we add it to the top level index.js, or wait until we figure out what a more general factor interface will look like?

lib/factor/factorQuadratic.js
```diff
@@ -0,0 +1,127 @@
...
+        firstTermNode = polyTerm;
+      }
+      else if (exponent.value === '1') {
+        secondTermNode = polyTerm;
+      }
+      else {
+        return Node.Status.noChange(node);
+      }
+    }
+    else {
```
evykassirer: when would this catch?

lib/factor/factorQuadratic.js
```diff
@@ -0,0 +1,127 @@
...
+      const polyTerm = new Node.PolynomialTerm(term);
+      const exponent = polyTerm.getExponentNode(true);
+      if (exponent.value === '2') {
+        firstTermNode = polyTerm;
+      }
+      else if (exponent.value === '1') {
+        secondTermNode = polyTerm;
+      }
+      else {
+        return Node.Status.noChange(node);
```
evykassirer: when would this catch?

lib/factor/factorQuadratic.js
```diff
@@ -0,0 +1,127 @@
...
+      }
+    }
+    else {
+      return Node.Status.noChange(node);
+    }
+  }
+
+  if (!firstTermNode || !constantTermValue) {
+    return Node.Status.noChange(node);
```
evykassirer: when would this catch?

lib/factor/factorQuadratic.js
```diff
@@ -0,0 +1,127 @@
...
+  }
+
+  const symbol = firstTermNode.getSymbolNode();
```
evykassirer: ooohhh in the quadratic check can you make sure they have the same symbol? and add a test for that too

lib/factor/factorQuadratic.js
```diff
@@ -0,0 +1,127 @@
...
+  const firstTermCoeffRootValue = Math.sqrt(Math.abs(firstTermCoeffValue));
+  const firstTermCoeffRootNode = Node.Creator.constant(firstTermCoeffRootValue);
+  let constantTermRootValue = Math.sqrt(Math.abs(constantTermValue));
+  let constantTermRootNode = Node.Creator.constant(constantTermRootValue);
+
+  // check if difference of squares: are a and c squares and there is no b
+  if (!secondTermNode) {
+    // must be a difference of squares
+    if (constantTermValue < 0 &&
+        firstTermCoeffRootValue % 1 === 0 &&
```
evykassirer: can you make a helper function to make it clear that you're checking to see if it's an integer?

or you could use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isInteger

lib/factor/factorQuadratic.js
```diff
@@ -0,0 +1,127 @@
...
+  }
+
+  if (!firstTermNode || !constantTermValue) {
+    return Node.Status.noChange(node);
+  }
+
+  const symbol = firstTermNode.getSymbolNode();
+  const firstTermCoeffValue = firstTermNode.getCoeffValue();
+  const firstTermCoeffRootValue = Math.sqrt(Math.abs(firstTermCoeffValue));
+  const firstTermCoeffRootNode = Node.Creator.constant(firstTermCoeffRootValue);
```
evykassirer: this name is probably ok? but might be confusing

at first I thought it was an nthroot fucntion node

lib/factor/factorQuadratic.js
```diff
@@ -0,0 +1,127 @@
...
+  let constantTermRootValue = Math.sqrt(Math.abs(constantTermValue));
+  let constantTermRootNode = Node.Creator.constant(constantTermRootValue);
+
+  // check if difference of squares: are a and c squares and there is no b
```
evykassirer: ... and c is negative

lib/factor/factorQuadratic.js
```diff
@@ -0,0 +1,127 @@
...
+  if (!secondTermNode) {
+    // must be a difference of squares
+    if (constantTermValue < 0 &&
+        firstTermCoeffRootValue % 1 === 0 &&
+        constantTermRootValue % 1 === 0) {
+      const polyTerm = Node.Creator.polynomialTerm(symbol, null, firstTermCoeffRootNode);
```
evykassirer: can you split this line into 2 lines? (and everywhere else you have a line like this)

lib/factor/factorQuadratic.js
```diff
@@ -0,0 +1,127 @@
...
+const Node = require('../node');
+
+// Given a node, will check if it's in the form of a quadratic equation
+// `ax^2 + bx + c`, and
+// if it is, will factor it using on of the following rules:
+//    - Difference of squares e.g. x^2 - 4 -> (x+2)(x-2)
+//    - Perfect square e.g. x^2 + 2x + 1 -> (x+1)^2
+//    - Sum/product rule e.g. x^2 + 3x + 2 -> (x+1)(x+2)
+//    - TODO: quadratic formula
+function factorQuadratic(node) {
```
evykassirer: it'd be nice if this was split into separate function for different cases, but I also see how they all share common variables, so maybe this is best?

lib/factor/factorQuadratic.js
```diff
@@ -0,0 +1,127 @@
...
+  }
+  else {
+    // check if perfect square: are a and c squares and b = 2*sqrt(a)*sqrt(c)
+
+    // if the second term is negative, then the constant in the parens is subtracted
+    // i.e. x^2 -2x + 1 -> (x-1)^2
+    const secondTermCoeffValue = secondTermNode.getCoeffValue();
+    if (secondTermCoeffValue < 0) {
+      constantTermRootValue = constantTermRootValue * -1;
+      constantTermRootNode = Negative.negate(constantTermRootNode);
```
evykassirer: alternatively, you can just make `constantTermRootNode` at the times you need it instead of at the beginning

not sure if that's better or not

I have a slight preference for this because it becomes more obvious why you need it

test/factor/factorQuadratic.test.js
```diff
@@ -0,0 +1,30 @@
...
+    ['x^2 + 2x + 1', '(x + 1)^2'],
+    ['x^2 - 2x + 1', '(x - 1)^2'],
+    ['x^2 + 3x + 2', '(x + 1)(x + 2)'],
+    ['x^2 - 3x + 2', '(x - 1)(x - 2)'],
+    ['x^2 + x - 2', '(x - 1)(x + 2)'],
+    ['x^2 + 4', 'x^2 + 4'],
+    ['x^2 + 4x + 1', 'x^2 + 4x + 1'],
+  ];
+  tests.forEach(t => testFactorQuadratic(t[0], t[1]));
+});
```
evykassirer: can you add some tests for things that shouldn't factor yet? to make sure we're not treating things as difference of square when they're not, etc.

I'm especially cautious about stuff with signs e.g. `-x^2 - 2x - 1`

lib/factor/factorQuadratic.js
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    const perfectProduct = 2 * firstTermCoeffRootValue * constantTermRootValue;
```
evykassirer: some more comments describing how these variables fit into things might be helpful

I understood what was going on, but had to read over the code a couple times to get it

lib/factor/factorQuadratic.js
```diff
@@ -0,0 +1,127 @@
...
+      const newNode = Node.Creator.operator('^', [paren, exponent]);
+      return Node.Status.nodeChanged(
+        ChangeTypes.FACTOR_PERFECT_SQUARE, node, newNode);
+    }
+    else if (firstTermCoeffValue === 1) {
+      // try sum/product rule: find a factor pair of c that adds up to b
+      const factorPairs = ConstantFactors.getFactorPairs(constantTermValue, true);
+      for (const pair of factorPairs) {
+        if (pair[0] + pair[1] === secondTermCoeffValue) {
+          const polyTerm = Node.Creator.polynomialTerm(symbol, null, firstTermCoeffRootNode);
```
evykassirer: we know that the coeff is 1, so the polyTerm is just the symbol right?

lib/checks/isQuadratic.js
aelnaiem: Hmmmmmm. Interesting. You're right :)

lib/factor/ConstantFactors.js
aelnaiem: Yea I agree and I thought about it but for now I think it's better to wait until that case arises. That seems like a better approach in general

lib/factor/factorQuadratic.js
aelnaiem: It shouldn't.

lib/factor/factorQuadratic.js
aelnaiem: It shouldn't.

lib/factor/factorQuadratic.js
aelnaiem: It shouldn't.

lib/factor/factorQuadratic.js
aelnaiem: Nice!

lib/factor/factorQuadratic.js
aelnaiem: Do you have a better name in mind? I agree that the naming is annoying.

lib/factor/ConstantFactors.js
evykassirer: weren't we using this factor functionality for something else? this still works there with this change?

lib/factor/factorQuadratic.js
evykassirer: can we take em out then?

lib/factor/factorQuadratic.js
evykassirer: I think if you wait until you need them to create/name them, it'll be more obvious what they do because the context

lib/factor/factorQuadratic.js
aelnaiem: Did it :) although each function takes a bunch of parameters (a, b and c of the quadratic)

lib/factor/ConstantFactors.js
aelnaiem: I actually pre-emptively added this function when adding other factoring functionality. Which caused me some confusion and which is why I think it's best to wait for the case to arise now hahaha

evykassirer: ahhh okay - good to know XD

lib/simplifyExpression/collectAndCombineSearch/addLikeTerms.js
```diff
@@ -120,6 +121,7 @@ function addNegativeOneCoefficient(node) {
         polyTerm.getSymbolNode(),
         polyTerm.getExponentNode(),
         polyTerm.getCoeffNode(),
+        false,
```
evykassirer: add a comment for what this means maybe

lib/simplifyExpression/collectAndCombineSearch/addLikeTerms.js
```diff
@@ -84,7 +84,8 @@ function addPositiveOneCoefficient(node) {
       newNode.args[i] = Node.Creator.polynomialTerm(
         polyTerm.getSymbolNode(),
         polyTerm.getExponentNode(),
-        Node.Creator.constant(1));
+        Node.Creator.constant(1),
+        true);
```
evykassirer: maybe add comment for what true means

lib/factor/factorQuadratic.js
```diff
@@ -44,84 +44,187 @@ function factorQuadratic(node) {
...
     }
   }
 
-  // TODO: quadratic formula
-  // a(x - (-b + sqrt(b^2 - 4ac)) / 2a)(x - (-b - sqrt(b^2 - 4ac)) / 2a)
-
   return Node.Status.noChange(node);
 }
 
+
```
evykassirer: extra line

lib/node/Creator.js
```diff
@@ -48,12 +48,12 @@ const NodeCreator = {
   // exponent might be null, which means there's no exponent node.
   // similarly, coefficient might be null, which means there's no coefficient
   // the symbol node can never be null.
-  polynomialTerm (symbol, exponent, coeff, coeffIsNegOne=false) {
+  polynomialTerm (symbol, exponent, coeff, coeffIsOne=false, coeffIsNegOne=false) {
```
evykassirer: I wonder if we could combine `coeffIsOne` and `coeffIsNegOne` to a shared variable... something about explicit coefficients

or is that more confusing?

lib/checks/isQuadratic.js
```diff
@@ -12,13 +12,14 @@ function isQuadratic(node) {
     return false;
   }
 
+  // get the number of symbols in the expression and ensure there's only one
```
evykassirer: sorry I'm being really picky now so you don't have to change this but

I think this comment is pretty close to what the code is doing, I meant more like 
"make sure only one symbol appears in the expression, i.e. just x and not both x and y"

the point of confusion for people could be that it's a _set_ (which you show in the var name which is helpful, so it's probably fine) and there can be multiple symbol terms (multiple symbols) as long as they're of the same symbol name

lib/factor/factorQuadratic.js
```diff
@@ -153,61 +144,57 @@ function factorDifferenceOfSquares(node, symbol, aValue, bValue, cValue, negate)
...
     // try sum/product rule: find a factor pair of c that adds up to b
     const factorPairs = ConstantFactors.getFactorPairs(cValue, true);
     for (const pair of factorPairs) {
       if (pair[0] + pair[1] === bValue) {
-        const aRootValue = Math.sqrt(Math.abs(aValue));
-        const aRootNode = Node.Creator.constant(aRootValue);
-
-        const polyTerm = Node.Creator.polynomialTerm(
-          symbol, null, aRootNode);
+        const polyTerm = Node.Creator.polynomialTerm(symbol, null);
```
evykassirer: oh wait you can just use Node.Creator.symbol

evykassirer:
I'm worried that `2x^2 + 2x -> 2x(x + 1)` might still not work

lib/factor/factorQuadratic.js
```diff
@@ -153,61 +144,57 @@ function factorDifferenceOfSquares(node, symbol, aValue, bValue, cValue, negate)
...
-      const newNode = Node.Creator.operator('^', [paren, exponent]);
-      return Node.Status.nodeChanged(
-        ChangeTypes.FACTOR_PERFECT_SQUARE, node, newNode);
+    const polyTerm = Node.Creator.polynomialTerm(symbol, null, aRootNode);
+    let paren = Node.Creator.parenthesis(
+      Node.Creator.operator('+', [polyTerm, cRootNode]));
+    const exponent = Node.Creator.constant(2);
+
+    if (negate) {
+      paren = Negative.negate(paren);
```
evykassirer: if you negate before doing the ^2, won't the negating cancel out?

ie `(-(x+2))^2`

I see your test passes though, which confuses me

evykassirer:
still missing tests for `factorSymbol` with negate - I think there might be bugs there

lib/factor/factorQuadratic.js
```diff
@@ -56,35 +69,13 @@ function factorQuadratic(node) {
     cValue = -cValue;
```
evykassirer: these should default to 0 above, or else negating doesn't really make sense in the null case

evykassirer:
added 6 comments, mostly just stuff you missed from the last code review

lemme know if there's anything I can do to make it easier for you to see and check off my comments to make sure you address them all (I know github doesn't really do a good job of making it easy)

evykassirer:
👍

PR #107, opened 2017-01-30
#44: Improve our linter

Raibaz:
this addresses  #44 

This adds linting rules for:

* Require statements alphabetisation (done with the `sort-requires` plugin)
* Stroustrup brace style enforcing
* Force spaces before and after keywords

The enforcement of the `'use strict'` line is still under discussion.


.eslintrc.js
```diff
@@ -12,7 +15,7 @@ module.exports = {
         "module": true,
         "describe": true,
         "it": true
-    },
+    },
```
evykassirer: remove trailing spaces here please (too bad the linter doesn't lint the linter file haha)

.eslintrc.js
```diff
@@ -47,6 +50,17 @@ module.exports = {
             "always", {
                 "avoidQuotes": true
             }
+        ],
+        "brace-style": [
+            "error",            
+            "stroustrup",
+            {"allowSingleLine": true}
```
evykassirer: can you open this up onto a new line? like the "always" and "avoidQuotes" above?

```
"stroustrup", {
    "allowSingleLine": true}
}
```

lib/node/Status.js
```diff
@@ -1,6 +1,7 @@
 'use strict';
 
 const clone = require('../util/clone');
+
```
evykassirer: don't need new line here I think? or do you want to sort them by directory they're coming from? or by if they start with a capital letter? let me know what you were thinking :)

lib/simplifyExpression/fractionsSearch/addConstantFractions.js
```diff
@@ -1,8 +1,9 @@
 'use strict';
 
-const math = require('mathjs');
 const clone = require('../../util/clone');
 const divideByGCD = require('./divideByGCD');
+const math = require('mathjs');
```
evykassirer: I think this require should go separate at the top, with a new line after it

because it's a require from outside of `mathsteps`

lib/solveEquation/stepThrough.js
```diff
@@ -1,17 +1,15 @@
 'use strict';
 
+const ChangeTypes = require('../ChangeTypes');
+const checks = require('../checks');
 const Equation = require('../equation/Equation');
-const EquationStatus = require('../equation/Status');
 const EquationOperations = require('./EquationOperations');
```
evykassirer: I kinda liked how the equation requires were grouped together, though I'm not that set on it

lib/node/Status.js
Raibaz: I generally grouped the ones that start with a non-capital letter first and the ones starting with a capital letter second, but I don't have a strong opinion on this, what do you think?

lib/simplifyExpression/fractionsSearch/addConstantFractions.js
Raibaz: Yep you're right

Raibaz:
Some of the issues marked as still missing in the issue were actually already there, so of course I didn't add them :)

evykassirer:
oh! haha I didn't realize we already had those rules xD my bad

lib/node/Status.js
evykassirer: seems fine to me :)

evykassirer:
ok!! looks good to me

🎉 🎉 🎉 🎉 🎉 🎉 🎉

PR #108, opened 2017-01-30
Remove semver check

aelnaiem:
https://github.com/socraticorg/mathsteps/issues/82

Removing the semver check for node version, and just adding a requirements section to the README.

We can add a programmatic check once npm provides a cleaner solution. Right now, users can turn on a flag that will warn them based on the engine in the package.json


evykassirer:
Thanks!

Issue #58, opened 2016-12-24
Missing test coverage


evykassirer:
A great first PR is to pick something on this list and write tests for it!

- [ ] `Symbols.js`
- [ ] `solveEquation/EquationOperations.js`
- [ ] in `CombineChecks.test.js` separately test the functions `CombineChecks.canMultiplyLikeTermPolynomialNodes`, `CombineChecks.canAddLikeTermPolynomialNodes`, and `CombineChecks.canRearrangeCoefficient`

feel free to point out more missing tests below!

nkprince007:
Can I be assigned this?

nkprince007:
Ok.

evykassirer:
FYI @nkprince007, @arbylee wrote tests for `Symbols.js` in #88 :)

nkprince007:
Thanks a lot, guys, but I'm a little caught up ATM, definitely like to contribute, soon as I find time.

evykassirer:
Sounds good @nkprince007! If you don't think you can get to this issue for a few weeks and don't have strong feeling towards it, we can un-assign you and when you have more time you take on whatever's free.  If you're exciting about this issue though, we can keep it until you have the time :) let me know!

PR #119, opened 2017-02-05
Missing tests

nitin42:
addresses #58 

Let me know about these ones. yet to write for `EquationOperations.js`


test/addPolynomial.test.js
```diff
@@ -0,0 +1,16 @@
+const canAdd = require('../lib/checks/canAddLikeTermPolynomialNodes');
```
evykassirer: can you name `require`s and filenames after the full function?

e.g. `canAddLikeTermPolynomialNodes.test.js`

and `const canAddLikeTermPolynomialNodes = require('../lib/checks/canAddLikeTermPolynomialNodes');`

test/multiplyPolynomialTest.js
```diff
@@ -0,0 +1,16 @@
+const canMultiply = require('../lib/checks/canMultiplyLikeTermPolynomialNodes');
```
evykassirer: make sure this file name is .test.js (with the dot before `test`)

test/changeType.test.js
```diff
@@ -0,0 +1,18 @@
+const assert = require('assert');
```
evykassirer: I don't think this test file is necessary in my opinion

if it's the dictionary, it exists, right?

test/clone.test.js
```diff
@@ -0,0 +1,18 @@
+const assert = require('assert');
+const math = require('mathjs');
+
+const clone = require('../lib/util/clone');
+
+function testCloner(expNode, new_node) {
+  it ('creates a clone of a node', function () {
+    assert.equal(clone(math.parse(expNode)), new_node);
```
evykassirer: wait, does this pass?

`new_node` would be a string or integer, and `clone(math.parse(expNode))` would be a node

test/clone.test.js
```diff
@@ -0,0 +1,18 @@
...
+  });
+}
+
+describe('Clone', function () {
+  const tests = [
+    [1, 1],
+    ['x', 'x']
```
evykassirer: probably make these more complicated so it nests, also always use strings (`"1"` instead of `1`)

test/equation.test.js
```diff
@@ -0,0 +1,24 @@
+const assert = require('assert');
+
+const Equation = require('../lib/equation/Equation');
+
+function makeEquation(left, right, comp) {
+  return (new Equation(left, right, comp).print());
```
evykassirer: can you split this into two lines:

```
const equation = new Equation(left, right, comp);
return equation.print()
```

test/equation.test.js
```diff
@@ -0,0 +1,24 @@
...
+}
+
+function equationRepresenter(left, right, comp, output) {
+  it ('represents an equation', () => {
```
evykassirer: can you print the `output` here instead?

test/equation.test.js
```diff
@@ -0,0 +1,24 @@
+const assert = require('assert');
+
+const Equation = require('../lib/equation/Equation');
+
+function makeEquation(left, right, comp) {
```
evykassirer: maybe name this `constructAndPrintEquation`

test/multiplyPolynomialTest.js
```diff
@@ -0,0 +1,16 @@
...
+
+const TestUtil = require('./TestUtil');
+
+function canBeMultiplied(expr, multipliable) {
+  TestUtil.testBooleanFunction(canMultiply, expr, multipliable);
+}
+
+describe('Multiply like term polynomial', () => {
```
evykassirer: "can multiply like term polynomials"

test/step.test.js
```diff
@@ -0,0 +1,22 @@
+const assert = require('assert');
+const math = require('mathjs');
+
+const stepThroughExpression = require('../lib/simplifyExpression/stepThrough');
+
+function testStepsInExpression(exprStr, outputStr) {
+  it ('test the steps in simplifying expression', function () {
+    assert.equal((stepThroughExpression(math.parse(exprStr))[0]['changeType']), outputStr);
```
evykassirer: can you make this line shorter than 80chars?

evykassirer: also let's test `newNode`, not just `changeType` :)

test/changeType.test.js
nitin42: OK.

test/clone.test.js
nitin42: It does pass.

test/step.test.js
nitin42: I had the same doubt about testing the changeType.

nitin42: Just a question ! For eg - if the input expression is `2(x + y)`and running `stepThroughExpression` gives the new_node. But what should be tested next!? changeGroup? Value? or the args?

This is the output 
```
Node {
  implicit: true,
  op: '*',
  fn: 'multiply',
  args: 
   [ Node { value: '2', valueType: 'number', changeGroup: undefined },
     Node { name: 'x', changeGroup: undefined } ],
  changeGroup: 1 }
```

test/clone.test.js
nitin42: OK fine. For the simple example we can test the value. But running the `clone()` function on a complex example, we get the child args and function `fn`. What should be actually tested for the more complex example. ?  Same goes for the test mentioned in `step.test.js`.

test/clone.test.js
evykassirer: I'm so confused whyyyy

I'll clone your repo later and play around with it

test/clone.test.js
evykassirer: test that the nodes are equivalent probably, like in the flatten tests

test/step.test.js
evykassirer: test the whole node for equivalence to `math.parse('2(x+y)')` like in flatten, I think

lemme know how that works

nitin42: It works for the examples like `2+2` or `2*3*4`. where we get the changeGroup and changeType and the whole node is tested. But it cannot be used for complex examples.

test/clone.test.js
nitin42: I think their we were simplifying the expression and checked the value of the newNode after simplifying. But in this case we are not simplifying the node, we are just cloning. Let me know what you think !

evykassirer: oh yes, I mean compare node equivalence in the same way but no simplifying this time

test/step.test.js
evykassirer: why doesn't it work for complex examples? wanna chat more about this on gitter? 

here on github is fine too, whatever works for you

test/clone.test.js
nitin42: Ok but one thing .. For eg - 

Node = `math.parse('x')  // Node { name: 'x', comment: '' }`

Node_To_Be_Clone = `clone(math.parse('x') // Node { name: 'x', changeGroup: undefined } `

And the test case that I made is that the value of the node that is being clone should be equal to the original one..  

Sounds good ? or you suggest !

test/step.test.js
nitin42: DM is better..

test/clone.test.js
nitin42: Ok !

nitin42:
I think its `./lib/learn.js` for failed checks !

test/equation.test.js
```diff
@@ -2,14 +2,15 @@ const assert = require('assert');
...
 const Equation = require('../lib/equation/Equation');
 
-function makeEquation(left, right, comp) {
-  return (new Equation(left, right, comp).print());
+function constructAndPrintEquation(left, right, comp) {
+  const equation = new Equation(left, right, comp);
+  return equation.print();
 }
 
 function equationRepresenter(left, right, comp, output) {
```
evykassirer: maybe rename this to `testEquationConstructor`

test/multiplyPolynomialTest.test.js
```diff
@@ -0,0 +1,16 @@
+const canMultiply = require('../lib/checks/canMultiplyLikeTermPolynomialNodes');
```
evykassirer: this file should be `canMultiplyLikeTermPolynomialNodes.test.js

test/equation.test.js
nitin42: Ohk !

test/equation.test.js
```diff
@@ -0,0 +1,25 @@
+const assert = require('assert');
```
evykassirer: capitalize equation in this file name

test/multiplyPolynomialTest.test.js
nitin42: 👍

test/canRearrange.test.js
```diff
@@ -0,0 +1,15 @@
+const canRearrange = require('../lib/checks/canRearrangeCoefficient');
```
evykassirer: `canRearrangeCoefficient` (and same in the file name)

evykassirer: etc for the rest of the files too

nitin42: Fine !

lib/simplifyExpression/index.js
```diff
@@ -16,3 +16,4 @@ function simplifyExpressionString(expressionString, debug=false) {
 }
 
 module.exports = simplifyExpressionString;
+
```
evykassirer: don't need this new line

test/canAddLikeTermPolynomialNodes.test.js
```diff
@@ -0,0 +1,16 @@
+const canAddLikeTermPolynomialNodes = require('../lib/checks/canAddLikeTermPolynomialNodes');
+
+const TestUtil = require('./TestUtil');
+
+function canBeAdded(expr, addable) {
+  TestUtil.testBooleanFunction(canAddLikeTermPolynomialNodes, expr, addable);
+}
+
+describe('Add like term polynomials', () => {
```
evykassirer: 'Can add like term polynomials' (you're not actually adding them, just seeing if you can)

test/canAddLikeTermPolynomialNodes.test.js
```diff
@@ -0,0 +1,16 @@
...
+  const tests = [
+    ['x^2 + x^2', true],
+    ['x + x', true],
+    ['x^3 + x', false],
+  ];
+  tests.forEach(t => canBeAdded(t[0], t[1]));
```
evykassirer: oh sorry I missed these before - they should be called "testCanBeAdded" like in the other test files

and similar for the other test functions

evykassirer:
oops, added a few more comments I forgot last time

thanks your patience - I think we can merge this very soon!

nitin42:
Ya sure. no hurries !!

nitin42:
Sorry... coz previously I added a file by mistake which was for my use only . Deleted in this commit !

lib/simplifyExpression/index.js
```diff
@@ -15,5 +15,4 @@ function simplifyExpressionString(expressionString, debug=false) {
   return [];
 }
 
-module.exports = simplifyExpressionString;
-
+module.exports = simplifyExpressionString;
```
evykassirer: all files should end in exactly one new line

you had two before, which is why I commented

but now wherever you see the 🚫 on github you should add another new line at the end

evykassirer:
just one more small change then I'm happy to merge this :D

evykassirer:
thanks for doing this! :D

nitin42:
hahaha 😆

nitin42:
I need some more clarity on the `collect and combine search` ! R u online on Gitter right now ?

evykassirer:
yup!

PR #121, opened 2017-02-09
Fixes #111 (nthRoot() existence check)

### LINKED ISSUE #111: Node existence check for nthRoot
We should check that a node exists when using nthRoot otherwise `Cannot read property 'type' of undefined` will result from a nonsense query such as `x = nthRoot( )`, rather than identifying that the query doesn't make sense.

This might be worth discussing with mathjs as a root of nothing might not be something would want to parse as part of a valid math tree.

shirleymiao:
Fixes #111 

Adds check that a node exists when using nthRoot, to prevent a `Cannot read property 'type' of undefined` error that results from a nonsense query such as `x = nthRoot( )`.


evykassirer:
<3

PR #123, opened 2017-02-10
Fix #113 (handle exponents on coefficients of polynomial terms)

### LINKED ISSUE #113: handle exponents on coefficients of polynomial terms
Right now there's this bug:

- `(2x)^2` -> `2x^2` (instead of `4x^2`)
- `(-x)^2` -> `-x^2` (instead of `x^2`)

There's nothing that handles exponents on simple polynomial terms like this. We should fix `removeUnnecessaryParens` to not remove parens if there's an exponent, and then implement the polynomial operation to do exponents on polynomial terms. (probably in its own tree search after `divisionSearch`?)

thanks to @kevinbarabash for finding this

shirleymiao:


evykassirer:
:taco:

Issue #115, opened 2017-02-03
don't add parenthesis before subtracting/adding (equations)


evykassirer:
as discussed in #90 (thanks @arve0 !)

> I have this simple equation: `x + 1 = 2`
> 
> Here are the steps from `solveEquation()`:
> 1. SUBTRACT_FROM_BOTH_SIDES
>   x + 1 = 2 => (x + 1) - 1 = 2 - 1
> 2. SIMPLIFY_LEFT_SIDE
>   (x + 1) - 1 = 2 - 1 => x = 2 - 1
>     1. COLLECT_AND_COMBINE_LIKE_TERMS
>       x + 1 - 1 = 2 - 1 => x + 0 = 2 - 1
>         1. COLLECT_LIKE_TERMS
>           x + 1 - 1 = 2 - 1 => x + (1 - 1) = 2 - 1
>         2. SIMPLIFY_ARITHMETIC
>           x + 1 - 1 = 2 - 1 => x + 0 = 2 - 1
>     2. REMOVE_ADDING_ZERO
>       x + 0 = 2 - 1 => x = 2 - 1
> 3. SIMPLIFY_RIGHT_SIDE
>   x = 2 - 1 => x = 1
> 
> I expected `newEquation` of step 1 to be without the parenthesis, like in substep a (2 > i > a). Might be a reason the parenthesis is added? For subtracting several terms, maybe?
> 
> **Edit:** To be a bit more specific, I expected `(x + 1) - 1 = 2 - 1` to be `x + 1 - 1 = 2 - 1`.

evykassirer:
so this happens in `performTermOperationOnExpression` in `EquationOperations.js`

we **should** add parenthesis if the operator is `/` or `*` or `^`
e.g. `2y = 2x + 2 => y = (2x + 2) / 2`

we **should not** though, if the operator is `+` or `-`, because we don't need parens to add/subtract to both sides

this should be a change that's only a few lines :)

rshconn:
I'm working on this!

evykassirer:
RACHEL! 😸  ❤️

excited to have you contributing

kevinbarabash:
@evykassirer are the parens necessary for the `/`, `*`, and `^`?  Wouldn't the following expression tree:
```
{
   type: "OperationNode",
   op: "/",
   args: [
       {
           type: "OperationNode",
           op: "+",
           args: [
               { type: SymbolNode, name: "x" }, 
[...]

PR #125, opened 2017-02-12
parentheses in performTermOperationOnExpression 

rshconn:
Modified performTermOperationOnExpression to add parentheses only for /, *, and ^ operations per #115. Created new test function `testPerformTermOperationExpressionStep` to test the equation at a given step. 

Tests cover +, -, /, and * cases:
-  (From #115)  x + 1 = 2  ->  x + 1 - 1 = 2 - 1 
-  (From #115)  2y = 2x + 2  ->  y = (2x + 2) / 2 
- x - 1 = 2  ->  x - 1 + 1 = 2 + 1
- y/2 = 2x + 2  ->  y = (2x + 2) * 2

### DIFF
--- a/lib/solveEquation/EquationOperations.js
+++ b/lib/solveEquation/EquationOperations.js
@@ -220,12 +220,14 @@ function performTermOperationOnEquation(equation, op, term, changeType) {
 
 // Performs an operation of a term on an entire given expression
 function performTermOperationOnExpression(expression, op, term) {
-  const node = (Node.Type.isOperator(expression) ?
-    Node.Creator.parenthesis(expression) : expression);
-
+  if (Node.Type.isOperator(expression) &&
+    (op === '/' || op === '*' || op === '^')) {
+      var node = Node.Creator.parenthesis(expression);
+  } else {
+    var node = expression;
+  }
   term.changeGroup = 1;
   const newNode = Node.Creator.operator(op, [node, term]);
-
   return newNode;
 }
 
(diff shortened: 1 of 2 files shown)


lib/solveEquation/EquationOperations.js
```diff
@@ -220,12 +220,14 @@ function performTermOperationOnEquation(equation, op, term, changeType) {
 
 // Performs an operation of a term on an entire given expression
 function performTermOperationOnExpression(expression, op, term) {
-  const node = (Node.Type.isOperator(expression) ?
-    Node.Creator.parenthesis(expression) : expression);
-
+  if (Node.Type.isOperator(expression) &&
+    (op === '/' || op === '*' || op === '^')) {
+      var node = Node.Creator.parenthesis(expression);
```
evykassirer: we only use `let` and `const`, so I think you'd need to do `let node;` before the if statement and then assign it inside the if/else (not super elegant, I know)

test/solveEquation/solveEquation.test.js
```diff
@@ -128,3 +128,21 @@ describe('solveEquation errors', function() {
   ];
   tests.forEach(t => testEquationError(t[0], t[1]));
 });
+
+function testPerformTermOperationExpressionStep(equationString, outputString, step, debug=false) {
```
evykassirer: oo I like this idea of testing steps like this - wonder if we should use it elsewhere

it might break if we change how we apply steps, but we can always come back and change the tests if so, I think this is the best way to do it :) nice!

PR #127, opened 2017-02-14
Fix #126 (Add parens in util > print where necessary)

### LINKED ISSUE #120: Incorrect before_node in GROUP_COEFFICIENTS change
@karuppiah7890 noticed that one of the steps prints out the incorrect math statement. The final result is correct, so it's even more curious how this is happening. Most likely due to how we clone the node.

```
before change: 2x + 3x = 35
change: SIMPLIFY_LEFT_SIDE
after change: 5x = 35
# of substeps: 2
Substeps : 
before change: 2x + 3x = 35
change: GROUP_COEFFICIENTS
after change: (2 + 3)x = 35
# of substeps: 0
before change: 2 + 3x = 35 *This is wrong and should say 2x + 3x = 35*
change: SIMPLIFY_ARITHMETIC
after change: 5x = 35
# of substeps: 0

before change: 5x = 35
change: DIVIDE_FROM_BOTH_SIDES
after change: (5x) / 5 = 35/5
# of substeps: 0
before change: 5x / 5 = 35/5
change: SIMPLIFY_LEFT_SIDE
after change: x = 35/5
# of substeps: 0
before change: x = 35/5
change: SIMPLIFY_RIGHT_SIDE
after change: x = 7
# of substeps: 0
```

### LINKED ISSUE #126: Should `print` add in parens automatically where necessary?
Having to add explicit `ParenthesesNode`s makes the expression tree harder to manipulate and may result in output which doesn't look like what the expression tree represents.

Right now `print({ "/" [ { "+" [ "x", "2" ] } "2" }) => "x + 2 / 2"`.  Re-parsing `"x + 2 / 2"` results in `{ "+" [ "x", { "/" [ "2", "2" ] } ] }` which is not the same expression.

I'd like to propose the following behavior change for `print`:
```
print({ "/" [ { "+" [ "x", "2" ] } "2" }) => "(x + 2) / 2"
print({ "+" [ { "+" [ "x", "x" ] } { "+" [ "2", "2" ] } ] }) => "(x + x) + (2 + 2)"
```

The change would update `print` to add parens around any `+` operation automatically if it isn't the root node.

Flyr1Q:
As @kevinbarabash proposed in #126, add parenthesis around any + (as well as -) operation if it isn't the root node. It also fixes #120 issue as well.
 
```
print({ "/" [ { "+" [ "x", "2" ] } "2" }) => "(x + 2) / 2"
print({ "+" [ { "+" [ "x", "x" ] } { "+" [ "2", "2" ] } ] }) => "(x + x) + (2 + 2)"
```


test/util/removeUnnecessaryParens.test.js
```diff
@@ -22,7 +22,7 @@ describe('removeUnnecessaryParens', function () {
     ['(((-5)))', '-5'],
     ['((4+5)) + ((2^3))', '(4 + 5) + 2^3'],
     ['(2x^6 + -50 x^2) - (x^4)', '2x^6 - 50x^2 - x^4'],
-    ['(x+4) - (12 + x)', 'x + 4 - (12 + x)'],
+    ['(x+4) - (12 + x)', '(x + 4) - (12 + x)'],
```
evykassirer: I would expect `x + 4 - (12 + x)` actually, you don't need parenthesis around x and 4

and when flattened you should get three children of +: `x`, `4`, and `-(12 + x)`

lib/util/print.js
```diff
@@ -91,4 +96,25 @@ function printTreeTraversal(node) {
...
+// Need to add parens around any [+, -] operation
+// next to [/, *, ^] operation or parenthesis node
+// Examples:
+// { "/" [{ "+" ["x", "2"] }, "2"] } -> (x + 2) / 2.
+// { "+" [{ "+" ["x", "x"] }, { "+" ["2", "2"] }] } -> (x + x) + (2 + 2)
+// Check #120, #126 issues for more details.
+function needToAddParens(node, parentNode) {
+  if (!parentNode ||
+      !Node.Type.isOperator(parentNode) ||
+      '*/^'.indexOf(node.op) !== -1) {
```
evykassirer: I find this a bit confusing, since we're wanting the parent to have + or -

I'd prefer `'+-'.indexOf(node.op) === -1`

ooo or alternatively, you could just do

```
if (!parentNode ||
    !Node.Type.isOperator(parentNode, '+') ||
    !Node.Type.isOperator(parentNode, '-') || {
```

not sure which one is cleaner - what do you like better?

lib/util/print.js
```diff
@@ -91,4 +96,25 @@ function printTreeTraversal(node) {
...
+    return false;
+  }
+
+  if ('+-'.indexOf(parentNode.op) === -1 ||
```
evykassirer: same here, `'*/^'.indexOf(parentNode.op) >= 0` shows what we're looking for instead of what we're not looking for, which I find easier to follow

lib/util/print.js
```diff
@@ -91,4 +96,25 @@ function printTreeTraversal(node) {
   }
 }
 
+// Need to add parens around any [+, -] operation
+// next to [/, *, ^] operation or parenthesis node
```
evykassirer: I'd replace 'next to' with 'nested in' or something showing that it's above it in the tree, not beside it

lib/util/print.js
```diff
@@ -91,4 +96,25 @@ function printTreeTraversal(node) {
...
+// Check #120, #126 issues for more details.
+function needToAddParens(node, parentNode) {
+  if (!parentNode ||
+      !Node.Type.isOperator(parentNode) ||
+      '*/^'.indexOf(node.op) !== -1) {
+    return false;
+  }
+
+  if ('+-'.indexOf(parentNode.op) === -1 ||
+      parentNode.args.some(Node.Type.isParenthesis)) {
```
evykassirer: why is this condition needed? 

would i mean that `2  + (4 - 5)` be printed as  `(2)  + (4 - 5)` ?

test/util/removeUnnecessaryParens.test.js
Flyr1Q: Oh, that what was proposed by @kevinbarabash in #126. I thought you agreed on that. That is actually even better, I will change ASAP

test/util/removeUnnecessaryParens.test.js
```diff
@@ -22,7 +22,7 @@ describe('removeUnnecessaryParens', function () {
...
     ['(2x)^2', '(2x)^2'],
     ['((4+x)-5)^(2)', '(4 + x - 5)^2'],
```
Flyr1Q: yeah, totally

lib/util/print.js
Flyr1Q: No, current node has to be Operation type, so 2 alone cannot be wrapped. But it does not matter, if 3 + 4 + (4 - 5) is preferable over (3 + 4) + (4 - 5). I will remove it nonetheless.

test/util/removeUnnecessaryParens.test.js
evykassirer: I think if you flatten it, it would take out the parens regardless :)

lib/util/print.js
evykassirer: sounds good!

test/util/removeUnnecessaryParens.test.js
Flyr1Q: Actually, if I flattenOperands(node) in print, it heavily breaks tests (about 20 tests). I suppose simplify relied on print in some way. That is why I did't do it in first place.

lib/util/print.js
Flyr1Q: I just wanted to be careful, since `'*/^'.indexOf('') => 0`.

test/util/removeUnnecessaryParens.test.js
evykassirer: Simplify does use print, but I'm confused why this would..

OH maybe try cloning the node first and then printing it? 

If that doesn't work, lemme know which tests fail and I can help you look into it

lib/util/print.js
evykassirer: Ooo good call. What about `parentNode.op &&`?

test/util/removeUnnecessaryParens.test.js
Flyr1Q: If I flattenOperands cloned node, 22 tests fail.

```
function print(node, showPlusMinus=false) {
  // let string = printTreeTraversal(node);

  const inputNode = flattenOperands(clone(node));
  let string = printTreeTraversal(inputNode);

  if (!showPlusMinus) {
    string = string.replace(/\s*?\+\s*?\-\s*?/g, ' - ');
  }
```

Makes this tests fail. I'll remove similar, for brevity.

```
  1) Equation constructor 2*x^2 + x = 4:
     TypeError: node.clone is not a function

  4) factorQuadratic x^2 + 2x -> x(x + 2):
      AssertionError: 'x * (x + 2)' deepEqual 'x(x + 2)'
      -x * (x + 2)
      +x(x + 2)

  14) combinePolynomialTerms multiplication 2x * x^2 * 5x -> 10x^4:
      AssertionError: '10x^4' deepEqual '10 * x^4'
      -10x^4
      +10 * x^4

  17) distribute (x+2+y)*x*7 -> (x^2 + 2 * x + y * x) * 7:
      AssertionError: '(x * x + 2x + y * x) * 7' deepEqual '(x * x + 2 * x + y * x) * 7'
      -(x * x + 2x + y * x) * 7
      +(x * x + 2 * x + y * x) * 7

  19) collects and combines like terms 2x^2 * y * x * y^3 -> 2 * x^3 * y^4:
      AssertionError: '2x^3 * y^4' deepEqual '2 * x^3 * y^4'
      -2x^3 * y^4
      +2 * x^3 * y^4

  22) distribution - x*y^4 (6x * y^2 + 5x*y - 3x) -> -6 * x^2 * y^6 - 5 * x^2 * y^5 + 3 * x^2 * y^4:
      AssertionError: '-6x^2 * y^6 - 5x^2 * y^5 + 3x^2 * y^4' deepEqual '-6 * x^2 * y^6 - 5 * x^2 * y^5 + 3 * x^2 * y^4'
      --6x^2 * y^6 - 5x^2 * y^5 + 3x^2 * y^4
      +-6 * x^2 * y^6 - 5 * x^2 * y^5 + 3 * x^2 * y^4

Comment (evykassirer) [inline test/util/removeUnnecessaryParens.test.js]: iiinteresting.... I'll try to poke around on that today

**Comment by evykassirer** [comment]
update: I'm heading home for reading week tomorrow and am not totally sure how much time I have, but I really wanna check this out so we can finish up this change :) I'll hopefully work on it in the car but I'll keep you updated

**Comment by Flyr1Q** [comment]
@evykassirer Ok. Btw it is actually hard to test this case, since nor removeUnnecessaryParens, nor flattenOperands does not remove parenthesis it this case - `(2 + 3)x`. It is removed between substeps in collectAndCombineSearch. Will work on this

**Comment by evykassirer** [comment]
alright I have investigated the broken tests and am here to report how I fixed it 🎉 

step 1:

Cloning does help, because the flatten function changes the structure of the node, and sometimes it needs to stay the way it was for the next step to work. So some of the tests were fixed with this:

screen shot 2017-02-18 at 8 56 12 pm

step 2:

The equation tests were broken for a totally unrelated reason - it was never turning the node into a string. It hadn't been caught yet, until I added that clone in, and then it was like "hey you can't clone a string" so I fixed that too

screen shot 2017-02-18 at 8 56 29 pm

step 3: update the remaining tests that are failing. 

one of them was related to flattening and getting rid of parens:

screen shot 2017-02-18 at 8 56 47 pm

the rest are all because flattening makes some multiplication implicit and others not implicit and so it changes things around... so just update the tests like this to be passing again:

screen shot 2017-02-18 at 8 57 27 pm

you good to add these changes into your diff? let me know if you have any questions :)

**Comment by Flyr1Q** [comment]
Ok, I've updated print.js according to comments & fixed the tests. Everything is great.

The last thing is to add new tests for this fix. 

As I wrote before, the parenthesis were removed only between substeps in collectAndCombineSearch. Precisely, newNode after GROUP_COEFFICIENTS & oldNode before SIMPLIFY_ARITHMETIC, due to evaluateConstantSum.

If you pass node with parenthesis to evaluateConstantSum, it will return status with oldNode w/o parenthesis. 

TestUtil.testSubsteps only checking newNodes. Should I change TestUtil, so that it will check oldNodes as well?

**Comment by evykassirer** [comment]
ah yes, this is from the other bug that was found right?

I think testing old nodes would be great, yeah :) if it's straightforward to add, that'd help a lot. right now we're not testing them at all 😕

**Comment by Flyr1Q** [comment]
I added tests for old nodes, as usual some tests are failing.

```
before change: 2x + 4x + y
change: COLLECT_LIKE_TERMS
after change: (2x + 4x) + y
before change: 2x + 4x + y
change: ADD_POLYNOMIAL_TERMS
after change: 6x + y
```

The parenthesis are removed in evaluateConstantSum & combineLikeTerms, because of the expression like `node = node.content;`. Subsequent statuses are based on this mutated node, so status.oldNode is always w/o parenthesis. In order to fix this I need to make several changes like that:

```
  // STEP 2A: evaluate arithmetic IF there's > 1 constant
  // (which is the case if it's a list surrounded by parenthesis)
  if (Node.Type.isParenthesis(constants)) {
    const constantList = constants.content;
    const evaluateStatus = arithmeticSearch(constantList);
    evaluateStatus.oldNode = constants; // <- assign oldNode manually
    status = Node.Status.childChanged(newNode, evaluateStatus, 0);
    substeps.push(status);
    newNode = Node.Status.resetChangeGroups(status.newNode);
  }
```

It seems bad & dirty, is there a cleaner way to do this?

Also, if this is beyond the scope of this PR, I can just provide tests for print, but then I have to manually remove parenthesis before printing, like that.

```
function testPrint(exprStr, outputStr, fn) {
  let input = math.parse(exprStr);
  if (fn) {
    fn(input);
  }
  TestUtil.testFunctionOutput(print, input, outputStr);
}

describe('print asciimath', function () {
// ...
});

describe('print with parenthesis', function () {
  const tests = [
    ['(2 + 3)x', '(2 + 3) * x'],
    ['(7 - 4)^x', '(7 - 4)^x'],
    ['(9 + 2)/x', '(9 + 2) / x'],
  ];
  tests.forEach(t => {
    testPrint(
      t[0], t[1],
      (input) => { input.args[0] = input.args[0].content });
  });
})
```

**Comment by evykassirer** [comment]
If I'm understanding what's going on correctly (correct me if not) I think you can fix with `clone` (and that's actually why we started cloning in the first place! but we probably missed a few places)

"Subsequent statuses are based on this mutated node" - it should clone before mutating, so the original one stays the same

I'm a bit tired today so I don't have the brainpower to figure out where the clone would happen, but take a crack and it and I can take a look later today or tomorrow if you need another pair of eyes on it :)

**Comment by evykassirer** [comment]
I think it also is reasonable to just put this in another PR, especially if that's easier (either is fine for me)

why do you have to remove the parens before printing? how will the test fail otherwise? (this seems unrelated to the other issue, right?)

**Comment by evykassirer** [comment]
hey btw do you want some mathsteps stickers as a thank you for your contributions? 😸 

img_20170218_173323

if you're interested, email mathsteps@socratic.org with your mailing address and I can send them over!

**Comment by Flyr1Q** [comment]
> why do you have to remove the parens before printing? how will the test fail otherwise? (this seems unrelated to the other issue, right?)

They will not fail. The point of this PR is to print ({ "/" [{ "+" ["x", "2"] }, "2"] }) => "(x + 2) / 2". It is just hard to reproduce this node, since nor removeUnnecessaryParens, nor flattenOperands does not remove parens, they are removed between substeps in evaluateConstantSum.

**Comment by evykassirer** [comment]
Ohhh okay I'd say just create nodes with the mathjs constructors then, like in the flatten tests. Do you think that would work?

**Comment by Flyr1Q** [comment]
Ohh, how did I miss that? Yeah, totally, will add ASAP

File: test/TestUtil.js [hunk 7]
```diff
@@ -33,10 +33,10 @@ TestUtil.testSimplification = function (simplifyingFunction, exprString,
 };
 
 // Test the substeps in the expression
-TestUtil.testSubsteps = function (collectAndCombineSearch, exprString, outputList,
+TestUtil.testSubsteps = function (fn, exprString, outputList,
```
evykassirer: ah, thanks for fixing this ^__^

test/util/print.test.js
```diff
@@ -1,20 +1,48 @@
...
     ['2+3+4', '2 + 3 + 4'],
     ['2 + (4 - x) + - 4', '2 + (4 - x) - 4'],
     ['2/3 x^2', '2/3 x^2'],
     ['-2/3', '-2/3'],
   ];
-  tests.forEach(t => testPrint(t[0], t[1]));
+  tests.forEach(t => testPrintStr(t[0], t[1]));
 });
+
+describe('print with parenthesis', function () {
```
evykassirer: perfect! exactly what I was thinking

this rarely shows up in the tests so it's fair that you didn't think of it

(fun fact: all the tests looked like this for a bit, until I realized it'd be much easier to parse strings :p)

evykassirer:
everything looks good! just fix the linter issues

https://travis-ci.org/socraticorg/mathsteps/builds/203692251

 (you can run `npm run lint` to see them locally)

test/util/print.test.js
Flyr1Q: Ahah, indeed, it is a lot easier :D

Flyr1Q:
DONE [x]
Thank you a lot for helping me with this PR.

evykassirer:
perfect! thanks so much for taking this on!

are you up for taking on the other change we were discussing in another PR?

also don't forget to email us if you want stickers ^__^ 

🎉 🎉 🎉

Flyr1Q:
Yeah, I planned to check it carefully before creating an issue.

> childChanged is used in the case where the content of a parenthesis node is changed, so it should add the parens back around both the old node and new node... hm

I've sent an email stickers ^^

evykassirer:
hi @Flyr1Q are you still interested in working on the followup PR to this?

Flyr1Q:
sorry @evykassirer, don't think I'll be able to work on this in the near future :(

evykassirer:
no problem :) thanks for replying so quickly

PR #143, opened 2017-02-25
readme changed

nitin42:
A little more organised.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,7 @@ npm install mathsteps --save
 The main objective of the mathsteps is to continue to evolve and provide more operations and features. Read below to learn how can you contribute to the project.
 
 **Contributing guide**
+
 Read our [contributing guide](CONTRIBUTING.md) to suggest more features, report bugs or contribute code.
 
 ## Test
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,74 +1,62 @@
...
-[![Join the chat at https://gitter.im/mathsteps-chat/Lobby](https://badges.gitter.im/mathsteps-chat/Lobby.svg)](https://gitter.im/mathsteps-chat/Lobby?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
-[![Build Status](https://travis-ci.org/socraticorg/mathsteps.svg?branch=master)](https://travis-ci.org/socraticorg/mathsteps)
+Mathsteps is a step-by-step math solver. It helps in solving equations, simplifying an expression and factoring a polynomial. (more to come 😄)
 
-https://www.youtube.com/watch?v=iCrargw1rrM
+**Solve an equation**
 
-## Requirements
-
-Mathsteps requires Node version > 6.0.0
```
kevinbarabash: is this no longer the case?

evykassirer:
One thing you did that I like was linking to the wiki - we should definitely add that to the README.

But in general, I still prefer the phrasing and content in the original README (the youtube video, the message about contributing, the node dependencies, the build instructions, etc)

If you'd like to argue for this change being merged, could you explain a bit about what specific things you wanted to change here and why?

nitin42:
nothing specific as such !

evykassirer:
If there's nothing in particular you were looking to change, then how about we close this PR and if you want you can open another one that just links to the wiki?

If there's anything else here you feel strongly about changing, let me know what and why and we can reopen this PR :)

README.md
nitin42: totally missed!

Issue #142, opened 2017-02-24
changeGroups vs tree positions


evykassirer:
we use `changeGroups` right now to show what part of a tree has changed

@tkosan: 

> tree positions are an effective and elegant notation for identifying the part of a tree where an error occurred. A short explanation of how tree positions work is present in the following document starting at the first use of the word "position":
> 
> [The language of expression trees](http://patternmatics.org/temp_1/mathfuture/lesson4/)
> 
> I learned about tree positions when I first started studying Dr. Alan Bundy's research, and now I use them for all kinds of things in MathPiper.

tree positions have been brought up a few times, so let's discuss them more here - what are their pros/cons vs changeGroups?

evykassirer:
@hmaurer : 

> one benefit of tree positions would be that you don't actually modify the AST of an expression, you "point" at a node in the structure.

hmaurer:
@evykassirer Also, tree positions are more general. Change groups are a subset of the cases where tree positions could be applied.

evykassirer:
true! errors aren't really "changes" anyways

are there other cases you can think of in mathsteps where it'd be handy to use tree positions?

I guess a nodestatus would have a treePosition list of changed nodes then, yea?

hmaurer:
@evykassirer yah that sounds right. Errors too, to point at the node which caused the error. I am not sure about other uses in mathsteps but on one of my pet projects, a first-order logic derivation engine, I used tree positions to "point" at a variable in an expression to get scoping information. For example, given `forall x Px -> Q`, if the user hovers "x" it will ask the engine where "x" is bound based on a tree position. e.g. `getVariableInfo(variablePosition, expr)`.

tkosan:
Tree positions are also used in MathPiper to "point" to subexpressions in an expression. Some use cases include identifying a subexpression to highlight, locating the subexpression that was clicked on in a view of an expression, and identifying a subexpression to apply a rewrite rule too. Here are some additional examples that use tree positions:

[Expression tree](http://patternmatics.org/mathsteps/tmp/tree1.png)

Assign the above expression to the variable "tree".
```
In> tree := '(c-b+a^d+d/d+a/c-b*a*d);
Result: ((((c - b) + a^d) + d/d) + a/c) - ((b*a)*d)
```

Obtain the subexpression that is at position 1,2.
```
In> tree[1][2]
Result: a/c
```

Obtain the subexpression that is at position 1,1,1,2.
```
In> tree[1][1][1][2]
[...]

PR #144, opened 2017-02-25
migrate to math-parser for parsing [WIP]

kevinbarabash:
Initial attempt at integrating [math-parser](https://github.com/kevinbarabash/math-parser) into mathsteps.  Still lots to do, but I got the example from the README working.

TODO:
- [x] remove uses of mathjs parser
- [ ] make all the tests pass
  - [x] rewrite flattenOperands
  - [x] figure out why substeps are different now
  - [ ] add hadParens prop to AST and use that to render parens even when they're not necessary
  - [ ] figure out why it isn't simplifying `-x * 6xy^6`
- [ ] cleanup code

Proof:
```
const mathsteps = require('./index');

const steps = mathsteps.simplifyExpression('2x + 2x + x + x');

steps.forEach(step => {
    console.log("before change: " + step.oldNode);         // before change: 2 x + 2 x + x + x
    console.log("change: " + step.changeType);             // change: ADD_POLYNOMIAL_TERMS
    console.log("after change: " + step.newNode);          // after change: 6 x
    console.log("# of substeps: " + step.substeps.length); // # of substeps: 3
});
```
produces the following output:
```
before change: [object Object]
change: ADD_POLYNOMIAL_TERMS
after change: [object Object]
# of substeps: 3
```


lib/checks/canAddLikeTermPolynomialNodes.js
```diff
@@ -1,5 +1,34 @@
 const Node = require('../node');
 
+const equal = (left, right) => {
```
kevinbarabash: extract into utils

lib/node/Creator.js
```diff
@@ -11,16 +11,41 @@ const NodeCreator = {
   operator (op, args, implicit=false) {
     switch (op) {
     case '+':
-      return new math.expression.node.OperatorNode('+', 'add', args);
+      return {
+        type: 'OperatorNode',
+        op: '+',
+        fn: 'add',
```
kevinbarabash: simplify with a dictionary lookup

kevinbarabash:
Making progress, some equations are actually being solved now such as `x + 3 = 4`.  Need to figure out why `evaluateConstantSum` is causing an infinite loop.

lib/checks/canAddLikeTermPolynomialNodes.js
```diff
@@ -24,7 +25,7 @@ function canAddLikeTermPolynomialNodes(node) {
   return restTerms.every(term => {
     const haveSameSymbol = sharedSymbol === term.getSymbolName();
     const exponentNode = term.getExponentNode(true);
-    const haveSameExponent = exponentNode.equals(sharedExponentNode);
+    const haveSameExponent = equals(exponentNode, sharedExponentNode);
```
evykassirer: do you prefer this over having `equals` be a method on nodes?

lib/node/Creator.js
```diff
@@ -4,45 +4,64 @@
...
 
+const opFns = {
+  '+': 'add',
+  '-': 'subtract',
+  '*': 'multiply',
+  '/': 'divide',
+  '^': 'pow',
+};
+
+const operatorNode = (op, fn, args) => {
```
evykassirer: I was assuming this would live in `math-parser` instead of `mathsteps` (because `mathjs` did both as well)

what are your thoughts on where it should go and why?

lib/util/equals.js
```diff
@@ -0,0 +1,31 @@
+function equals(left, right) {
```
evykassirer: also curious about this living here vs in `math-parser`

evykassirer:
took a skim and added a few comments, but I can see you're still working on it so I'll let you keep going and clean stuffs up ^_^ 

lemme know if you want more input on things as you're working!

lib/node/Creator.js
kevinbarabash: Eventually I'd want to move node creation functions into a separate package for math-ast tools.  The benefit of having tools such as node creation, traversal, transform, etc. methods separate from the parser is that people may want to build different parsers but using the same AST and related tools.

Before that though, I'd like to transition mathsteps to using the math-ast nodes instead of converting those nodes to mathjs-like nodes as we're doing now.

lib/util/equals.js
kevinbarabash: This should eventually go in a math-ast tools package.  This function though could be replaced with any function that does deep comparison of two objects.

lib/checks/canAddLikeTermPolynomialNodes.js
kevinbarabash: Having methods on nodes makes serialization more difficult.  The methods either need to be stripped before calling `JSON.stringify` or they need to the prototype which means that when they're deserialized the method needs to be added to the deserialized version.  Without them cloning becomes trivial as `const clone = JSON.parse(JSON.stringify(ast))`.

kevinbarabash:
@evykassirer thanks for the initial review.  I'll put in TODOs for things that will be address in future diffs like moving some of these helper functions into a separate package and migrating from mathjs-style nodes to math-ast nodes.

kevinbarabash:
I ran into a big of a snag with `simplifyPolynomialFraction`.  Either the `flattenOperands` or `mathjs.parse` itself changes the precedence of `*` and `/` depending on whether or not the operands involved are variables or not.  This will require changes to math-parser which I'm tracking in https://github.com/kevinbarabash/math-parser/issues/25.

kevinbarabash:
Continuing to make progress, down to 44 test failures.

kevinbarabash:
@evykassirer I'm confused about `canCollectLikeTerms`.  It returns false for `x * x + x * 2 + x * y`, but can't `x * x` be collected into `x ^ 2`?

kevinbarabash:
There are some situations where math-parser will need to emit `Parentheses` nodes where it isn't doing so right now so that tests like:
```
keeping parens in important places, on printing 5 + (3*6) + 2 / (x / y) -> 5 + (3 * 6) + 2 * y / x
```
will pass.  This work is being tracked by https://github.com/kevinbarabash/math-parser/issues/26.

evykassirer:
`canCollectLikeTerms` collects the node for `x * x` but not the higher level node `x * x + x * 2 + x * y` - it only returns true for nodes that are only a sum or product of the terms that can be collected

thanks - if this isn't clear, I should add some documentation somewhere!

kevinbarabash:
@evykassirer I should've been more specific.  As part of this test: https://github.com/socraticorg/mathsteps/blob/master/test/simplifyExpression/distributeSearch/distributeSearch.test.js#L33-L37, `canCollectLikeTerms` terms gets called on `x * x`, but it fails b/c `termTypes.length > 1` is not true (there's only one term type, `'x'`) on https://github.com/socraticorg/mathsteps/blob/4e81626f3d295f963652d21c0e17c96600b18c9e/lib/simplifyExpression/collectAndCombineSearch/LikeTermCollector.js#L43-L44.  I don't understand why `termTypes.length > 1` needs to be true to combine these terms in this case.  I tried getting rid of the check, but ran into some other issue.  I need to step through things with the current parser to see what's supposed to happen.

evykassirer:
oooo okay so I think I remember how this works 😅 

"collect" would be: `x*y*x -> (x*x)*y` or something like that, where it collects it to be in the parens

just `x*x` is already ready to be combined, doesn't need to be "collected" - if that makes sense

kevinbarabash:
@evykassirer I'll pick this up again on the weekend.

hmaurer:
@kevinbarabash Hi! Can I help on this? It would be nice to get it out of the way asap since multiple other issues sort of depend on it.

kevinbarabash:
@hmaurer I'll make a list of issues that I could use help with tomorrow morning.

kevinbarabash:
@hmaurer sorry for the late followup.  Looking at current failing tests the main issues seem to be:
- output differs slightly, but the result seems okay, e.g. 
  ```
  18) simplify nthRoot nthRoot(x * x * (2 + 3), 2) -> x * nthRoot((2 + 3), 2):
      AssertionError: 'x * nthRoot(2 + 3, 2)' deepEqual 'x * nthRoot((2 + 3), 2)'
      + expected - actual
      -x * nthRoot(2 + 3, 2)
      +x * nthRoot((2 + 3), 2)
      
      at Context.it (test/TestUtil.js:30:12)
  ```
- ordering of terms in different, but the terms are correct, e.g.
  ```
  20) collects and combines like terms 10 + (y^2 + y^2) -> 10 + 2y^2:
      AssertionError: '2y^2 + 10' deepEqual '10 + 2y^2'
      + expected - actual
      -2y^2 + 10
      +10 + 2y^2
      
      at Context.<anonymous> (test/simplifyExpression/oneStep.test.js:24:12)
  ```
- implicitness of multiplication is being copied, e.g.
  ```
  24) collects and combines like terms y * 2x * 10 -> 20x * y:
      AssertionError: '20 * x * y' deepEqual '20x * y'
      + expected - actual
      -20 * x * y
      +20x * y
      
      at Context.<anonymous> (test/simplifyExpression/simplify.test.js:10:12)
  ```
- some searches are failing for unknown reasons, e.g.
  ```
  28) cancelling out (x^3*y)/x^2 + 5 -> x * y + 5:
      AssertionError: 'x^3 * y / (x^2) + 5' deepEqual 'x * y + 5'
      + expected - actual
      -x^3 * y / (x^2) + 5
      +x * y + 5
      
      at Context.<anonymous> (test/simplifyExpression/simplify.test.js:10:12)
  ```

I believe that some searches are failing because `flattenOperands` doesn't know how to deal with trees that math-parser produces.  `flattenOperands` does a few things:
- convert from binary trees to n-ary
- convert subtraction to adding the additive inverse
- remove parenthesis that were in the input

I believe that we still need it to remove parenthesis, e.g. `(+ (+ 1 2) 3) => (+ 1 2 3)` (and the same for multiplication.  I'm going to try rewriting `flattenOperands` and see how many of these issues that takes care of.  If you have time to investigate the formatting differences that would be greatly appreciated.

hmaurer:
@kevinbarabash Thank you! I am travelling tomorrow but I will investigate this on Tuesday.

lib/solveEquation/EquationOperations.js
```diff
@@ -224,7 +224,9 @@ function performTermOperationOnExpression(expression, op, term) {
     Node.Creator.parenthesis(expression) : expression);
 
   term.changeGroup = 1;
-  const newNode = Node.Creator.operator(op, [node, term]);
+  const newNode = op === '-'
+    ? Node.Creator.operator('+', [node, Node.Creator.unaryMinus(term)])
+    : Node.Creator.operator(op, [node, term]);
```
kevinbarabash: Since math-parser doesn't output subtract nodes and the new flattenOperands doesn't handle them I thought we may as well handle the operation the way it would be parse.

lib/util/flattenOperands.js
```diff
@@ -323,4 +323,93 @@ function hasMultiplicationBesideDivision(node) {
...
+          }
+        }
+        return Node.Creator.operator('+', args, node.implicit);
+      } else if (node.type === 'OperatorNode' && node.op === '*') {
+        const args = [];
+
+        for (const arg of node.args) {
+          if (arg.type === 'OperatorNode' && arg.op === '*' && !arg.implicit) {
+            args.push(...arg.args);
+          } else if (arg.type === 'OperatorNode' && arg.op === '/' && arg.args[0].type === 'SymbolNode' && args.length > 0 && args[args.length - 1].type === 'ConstantNode') {
```
kevinbarabash: This could be cleaned up to use `isPolynomialTerm`

kevinbarabash:
Updating `flattenOperands` and modifying `performTermOperationOnExpression` to add a negative instead of doing subtraction has fixed a number of tests.  Many of the still failing tests appear to be failing b/c steps that should be substeps are being added a top-level steps which is causing many of the test comparisons to fail.

kevinbarabash:
Down to 14 test failures.

lib/node/Creator.js
```diff
@@ -14,22 +14,20 @@ const opFns = {
...
     args: args,
+    implicit: implicit,
   };
 };
 
 const NodeCreator = {
   operator (op, args, implicit=false) {
     if (op in opFns) {
-      if (op === '-' && args.length === 1) {
-        debugger;
```
evykassirer: I have no idea why we chose to put `debugger` here, but it's true that if it's actually a unary minus, we shouldn't be using this constructor and should use the `unaryMinus` constructor instead. It'd be worth throwing an error here

lib/solveEquation/EquationOperations.js
evykassirer: that makes sense. maybe add a comment though explaining that choice :)

evykassirer: (some people seeing this might not know about the subtract node thing)

evykassirer:
I'm only barely skimming things, not really reading anything - assuming it's best if you just get all your stuff together and then I take a look. But let me know if you'd like my input on anything as you go :)

Thanks so much for making this big change! I'm extra excited to not only have this new parser, but also that you'll know the codebase fairly well now and hopefully that'll make it easier for you to work on it more in the future or make suggestions for things I can change hehe ^_^

lib/node/Creator.js
kevinbarabash: I'm pretty sure I put the `debugger` there in a previous diff.  Sorry about the individual diffs being so messy.  I'll probably end up squashing everything in the end.

kevinbarabash:
@evykassirer thanks for being patient with these changes.  Hopefully I address most/all of the remaining issues this weekend.

lib/node/Creator.js
evykassirer: ooo of course, my bad. I'll just not look at diffs haha

evykassirer:
thank *you* for patiently going through to make everything work!! :)

hmaurer:
Hi @kevinbarabash, sorry for the late reply... Is your latest work pushed to your fork? Your last commit is from 11 days ago so I assume it isn't, especially since files like `lib/node/Type.js` haven't been touched.

kevinbarabash:
@hmaurer I didn't get to do anything last weekend :( so what's on github is current.

hmaurer:
@kevinbarabash Ok! I am doing some work on this today. How should we handle the fact that the new parser doesn't have parenthesis nodes? It makes perfect sense but some of the code in mathsteps relies on it :/

kevinbarabash:
There are a number of test failures that are proving really tricky to fix.  Fixing one test breaks others that I fixed previously.  :(  This is a combination of me not having a comprehensive understanding of all of the different search and transform routines and their order and trying to work at too low a level of abstraction.

I'm going to put this PR on hold while I look at creating matcher functions as described in https://github.com/socraticorg/mathsteps/issues/142 for math-ast.  I think having these functions will result in code that's easier to reason about and easier to maintain.

hmaurer:
@kevinbarabash Yeah, I tried to fix some of the tests yesterday but could not wrap my head around some of the issues. The new parser creates a slightly differently structured tree and this causes rewrite rules to break in quite unpredictable ways. A higher-level handling of rewrite rules (e.g. with matcher functions) will help.

evykassirer:
if I can help by describing stuff that's going on in rewrite rules, I'd be happy to

matcher functions sounds pretty awesome and I agree it'd be easier to read/reason about! let me know if I can help at all

do you think the matcher function code would make more sense in math-steps or in math-ast?

kevinbarabash:
I'll probably start working on the matcher inside of `math-parser` and then split it out into its own repo afterwards.  The `math-parser` itself will need to be able to recognize placeholder nodes (with qualifiers).  I think it makes sense to include a description of what those nodes look like in the `math-ast` spec.

tkosan:
@kevinbarabash you might find the [pattern matcher that is in Mathics](https://github.com/mathics/Mathics/blob/master/mathics/core/pattern.py) (which uses the Wolfram Language's syntax) to be useful as a reference.

Issue #138, opened 2017-02-22
handle distribution with fractions more smartly


evykassirer:
right now `(3 / x^2 + x / (x^2 + 3)) * (x^2 + 3)` turns into
`(3 / x^2 * x^2 + 9 / x^2 + x / (x^2 + 3) * x^2 + 3x / (x^2 + 3)` which isn't super helpful

I think distribution should check to see if one of the groups we're multiplying contains fractions. If one (not both, not none) is, then multiply the non-fraction term into the numerators of the other term. 

so here we'd have

 `(3 / x^2 + x / (x^2 + 3)) * (x^2 + 3)` -> `(3 *(x^2 + 3)) / x^2 + (x *(x^2 + 3)) / (x^2 + 3)`

(this also helps fix an issue in #88)

lexiross:
Hey @evykassirer - this one looks cool. Doing some digging now (and refreshing my algebra skills, lol)...

evykassirer:
really hope what I wrote actually covers the cases well haha XD

and if it does, I think this change will be really great

thanks for taking it on :D

lexiross:
Okay, I'm gonna try to rephrase to make sure I'm understanding right:

Given `(a + b) * (c + d)` where `a` or `b` or both are expressed as fractions, but `c` and `d` are _not_, instead of distributing them using the FOIL method (`a*c + a*d + b*c + b*d`), treat `(c + d)` as one term, so it would look like `a*(c + d) + b*(c + d)`.

is that right? Also, if only `a` or only `b` is a fraction, should this rule apply?

evykassirer:
almost! (at least, in what I was thinking)

If only `a` or only `b` yeah I think it should still apply.

I would say if `a` or `b` is a fraction, and both `c` and `d` are not fractions, we'd not do FOIL (as you said)
and instead  treat `(c+d)` as one term (like you said), but move it into the numerator of any fractions

e.g.

`(a/x + b/y) * (c + d) => (a*(c+d))/x + (b*(c+d))/y`

`(a + b/y) * (c + d) => a*(c+d) + (b*(c+d))/y`

`(a + b) * (c + d) => a*c + a*d + b*c + b*d`

how's that look?

lexiross:
Oh yeah, that makes sense! :)

evykassirer:
yay - thanks for helping us clarify!

lexiross:
[...]

PR #147, opened 2017-02-28
Better handling of distribution with fractions

lexiross:
Addresses #138 

This is just a first stab - happy to change around anything that should be different!


lib/simplifyExpression/distributeSearch/index.js
```diff
@@ -197,6 +216,14 @@ function distributeTwoNodes(firstNode, secondNode) {
   return Node.Creator.parenthesis(Node.Creator.operator('+', newArgs));
 }
 
+function hasFraction(args) {
+  return args.filter(isFraction).length > 0;
+}
+
+function isFraction(node) {
```
lexiross: I thought this might be a util function somewhere but I couldn't find it. Do you think maybe we should export functions like this in some kind of util so they can be used around the codebase?

test/simplifyExpression/distributeSearch/distributeSearch.test.js
```diff
@@ -2,7 +2,7 @@ const distributeSearch = require('../../../lib/simplifyExpression/distributeSear
 
 const TestUtil = require('../../TestUtil');
 
-function testDistributeMinus(exprStr, outputStr) {
+function testDistributeResult(exprStr, outputStr) {
```
lexiross: I renamed this function because it seemed more general than just testing the `minus` operation - does that seem okay?

test/simplifyExpression/distributeSearch/distributeSearch.test.js
```diff
@@ -50,3 +50,24 @@ describe('distribute', function () {
...
+
+describe('distribute with fractions', function () {
+  const tests = [
+    // distribute the non-fraction term into the numerator(s)
+    ['(1 / x +  x^2) * (x^3 + 1)',
+      '(1 * (x^3 + 1) / x + (x^5 + x^2))',
+    ],
+
+    // same
+    ['(3 / x^2 + x / (x^2 + 3)) * (x^2 + 3)',
```
lexiross: example from #138

evykassirer:
woohoo! thanks

I'm studying for a test today, but hope to review this before end of day tomorrow :)

lib/simplifyExpression/distributeSearch/index.js
evykassirer: there are some similar functions in `Node.Type`, so you could add it there if you want

lib/simplifyExpression/distributeSearch/index.js
```diff
@@ -197,6 +216,14 @@ function distributeTwoNodes(firstNode, secondNode) {
...
+  return Node.Type.isOperator(node) && node.op === '/';
```
evykassirer: you can just use the function `Node.Type.isOperator(node, '/')` - I find that a little cleaner :)

lib/simplifyExpression/distributeSearch/index.js
```diff
@@ -175,9 +175,28 @@ function distributeTwoNodes(firstNode, secondNode) {
   // the new operands under addition, now products of terms
   const newArgs = [];
 
+  // if exactly one group contains at least one fraction, multiply the
```
evykassirer: wow I love how clean this change is - just a single block of code 😍

test/simplifyExpression/distributeSearch/distributeSearch.test.js
evykassirer: yes! 

We had a big change go through recently that changed a lot of the tests and not all of them were named well- I thought I caught and fixed most of them but I should probably do a pass. 

thanks :)

evykassirer: actually could you just call it `testDistribute`

test/simplifyExpression/distributeSearch/distributeSearch.test.js
```diff
@@ -50,3 +50,24 @@ describe('distribute', function () {
   ];
   tests.forEach(t => testDistributeSteps(t[0], t[1]));
 });
+
+describe('distribute with fractions', function () {
```
evykassirer: could you also add a section for testing with substeps? because the code you changed does a specific step that's only part of the overall distribution (right? cause for example in your first test it then goes on to multiply x^2 by x^3 and stuff)

lib/simplifyExpression/distributeSearch/index.js
```diff
@@ -175,9 +175,28 @@ function distributeTwoNodes(firstNode, secondNode) {
...
+  // non-fraction group into the numerators of the fraction group
+  if ([firstArgs, secondArgs].filter(hasFraction).length === 1) {
+    const firstArgsFraction = hasFraction(firstArgs);
```
evykassirer: slight preference for naming this `firstArgsHasFraction` (to make it clearer it's a boolean) but I can see that it's kinda long, so it's cool if you'd prefer not

test/simplifyExpression/distributeSearch/distributeSearch.test.js
lexiross: cool, will do

lib/simplifyExpression/distributeSearch/index.js
lexiross: no I agree, I'll change it!

test/simplifyExpression/distributeSearch/distributeSearch.test.js
lexiross: sure, will do

lexiross:
thanks for the review @evykassirer! and thanks lol - the fat cat thing was quite fun and I had no idea so many people on the internet would be into it 😂 

just updated - lmk if there's anything else I can do

lib/simplifyExpression/distributeSearch/index.js
```diff
@@ -221,7 +221,7 @@ function hasFraction(args) {
 }
 
 function isFraction(node) {
```
evykassirer: do you want to move this to `Node.Type`? I could go either way

test/simplifyExpression/distributeSearch/distributeSearch.test.js
```diff
@@ -69,5 +64,21 @@ describe('distribute with fractions', function () {
     ],
   ];
 
-  tests.forEach(t => testDistributeResult(t[0], t[1]));
+  const multiStepTests = [
+
+    ['(2 / x +  3x^2) * (x^3 + 1)',
+      ['(2 * (x^3 + 1) / x + 3x^2 * (x^3 + 1))',
```
evykassirer: I'm kinda surprised that there's not parenthesis around `2 * (x^3 + 1)` - what do you think?

test/simplifyExpression/distributeSearch/distributeSearch.test.js
```diff
@@ -69,5 +64,21 @@ describe('distribute with fractions', function () {
...
+        '(2 * (x^3 + 1) / x + (3x^5 + 3x^2))']
+    ],
+
+    ['(2x + x^2) * (1 / (x^2 -4) + 4x^2)',
```
evykassirer: could you make one of them have a fancy numerator that need to get multiplied?

e.g. `(2x + x^2) * (3x^2 / (x^2 -4) + 4x^2)`

I'm not sure if this would end up getting multiplied actually... you can see how `(1 * (2x + x^2)` isn't resolved. Do you think it would be pedagogically useful to multiply it out as a substep? (vs how much extra code would it be :p )

lib/simplifyExpression/distributeSearch/index.js
lexiross: Oh yeah - I was thinking no since the functions in that class check for more specific types of fractions. I think it would make sense to move this one once it gets defined more than once :)

test/simplifyExpression/distributeSearch/distributeSearch.test.js
lexiross: Oh, good point - I had thought that given `a * b / c`, the `a * b` is considered the numerator. Is that not the case? If not, how would I fix that? Is there a way to explicitly parenthesize an expression? I think on lines 19 and 20 the order of ops is already correct...?

test/simplifyExpression/distributeSearch/distributeSearch.test.js
lexiross: Okay, I added that as a test (which does end up getting multiplied out). I think the existing test not getting fully distributed is due to the `1` being a constant rather than say `2x`? If we wanted a quick fix for that test, we could just add a rule that `1 * (any expr) = (any expr)`. What do you think?

test/simplifyExpression/distributeSearch/distributeSearch.test.js
evykassirer: yeah, that's what I was thinking too. You can explicitly parenthesize it by wrapping it in a `Node.Creator.Parenthesis` node, I hope that fixes this

test/simplifyExpression/distributeSearch/distributeSearch.test.js
evykassirer: ah okay, cool

there is a rule for that, but it's not used for distribution simplifications. maybe I'll just open another issue to add that in, since it's not really related to this PR - sound good?

lexiross: sounds good!

lexiross:
@evykassirer added the parenthesis step - but then ended up with an edge case for `2(x+3)/3`, which should have solved to `2/3 x + 2` but was instead `2x / 3 + 2`. I decided to only do the parentheses check if the denominator of the fraction we're working with is NOT a constant. Does that sound right?

lib/simplifyExpression/distributeSearch/index.js
```diff
@@ -178,13 +178,19 @@ function distributeTwoNodes(firstNode, secondNode) {
   // if exactly one group contains at least one fraction, multiply the
   // non-fraction group into the numerators of the fraction group
   if ([firstArgs, secondArgs].filter(hasFraction).length === 1) {
+    // console.log("FIRST ARGS", firstArgs);
```
evykassirer: (remove these before we merge)

evykassirer:
Hmmm yeah we should maybe standardize if we do `2/3 x` or `2x / 3` but I believe both can show up in the solver and I don't really mind if it ends up one way or the other. I'll open an issue to discuss that!

So I think it's alright either way you do it (though I have slight preference for always adding the parens)

Thanks for your patience this week btw :) It's been really heavy workload for me so I haven't been able to be on GitHub as much, but you're bringing up lots of good points and also this diff is great! :D

lib/simplifyExpression/distributeSearch/index.js
lexiross: whoops, my bad 🙈

lexiross:
Okay! So if we do the parentheses all the time, it's okay to change the existing test?

And no worries at all!! I'm in no rush :)

evykassirer:
Yeah, I don't think it changes the correctness or pedagogy, so go for it :)

lexiross:
Okay, updated :)

PR #148, opened 2017-03-27
Exposing the factoring module and adding more coverage

aelnaiem:
This covers two things. Increasing coverage by handling factoring cases when a != 1, and exposing the ability to get steps for factoring through `mathsteps.factor()`


lib/factor/factorQuadratic.js
```diff
@@ -192,22 +197,62 @@ function factorPerfectSquare(node, symbol, aValue, bValue, cValue, negate) {
...
         const secondParen = Node.Creator.parenthesis(
-          Node.Creator.operator('+', [symbol, Node.Creator.constant(pair[1])]));
+          Node.Creator.operator('+', [firstCoeff, secondCoeff]));
+
 
         // create a node in the general factored form for expression
         let newNode = Node.Creator.operator('*', [firstParen, secondParen], true);
+        if (gcd !== 1) {
+          const gcdNode = Node.Creator.constant(gcd);
+          newNode = Node.Creator.operator('*', [gcdNode, newNode], true);
```
evykassirer: I think it should be `'*', [gcdNode, firstParen, secondParen]`

lib/factor/factorQuadratic.js
```diff
@@ -192,22 +197,62 @@ function factorPerfectSquare(node, symbol, aValue, bValue, cValue, negate) {
...
 }
 
-// Will factor the node if it's in the form of x^2 + bx + c (i.e. a is 1), by
-// applying the sum product rule: finding factors of c that add up to b.
-// e.g. x^2 + 3x + 2 -> (x + 1)(x + 2)
+// Will factor the node if it's in the form of x^2 + bx + c, by
+// applying the sum product rule: finding factors of a*c that add up to b.
+// e.g. x^2 + 3x + 2 -> (x + 1)(x + 2) or
+// or   2x^2 + 5x + 3 -> (2x - 1)(x + 3)
 function factorSumProductRule(node, symbol, aValue, bValue, cValue, negate) {
```
evykassirer: I think this function has a few different steps (you can tell in how you've written out the code) and should return substeps for the factoring process

lib/simplifyExpression/stepThrough.js
```diff
@@ -43,8 +43,10 @@ function stepThrough(node, debug=false) {
       logSteps(nodeStatus);
     }
     steps.push(removeUnnecessaryParensInStep(nodeStatus));
-    const nextNode = Status.resetChangeGroups(nodeStatus.newNode);
-    nodeStatus = step(nextNode);
+
+    node = Status.resetChangeGroups(nodeStatus.newNode);
+    nodeStatus = step(node);
```
evykassirer: nice

test/factor/factor.test.js
```diff
@@ -0,0 +1,33 @@
+const assert = require('assert');
+const factor = require('../../lib/factor');
+const print = require('../../lib/util/print');
+
+const NO_STEPS = 'no-steps';
+
+function testFactor(expressionString, outputStr, debug=false) {
```
evykassirer: this seems really similar to other tests right? can we use `TestUtil`?

test/factor/factor.test.js
```diff
@@ -0,0 +1,33 @@
...
+  }
+  it(expressionString + ' -> ' + outputStr, (done) => {
+    assert.equal(lastStep, outputStr);
+    done();
+  });
+}
+
+describe('factor expressions', function () {
+  const tests = [
+    ['x^2', NO_STEPS],
```
evykassirer: can you add some that aren't quadratics at all, too?

test/factor/factorQuadratic.test.js
```diff
@@ -7,28 +7,42 @@ function testFactorQuadratic(input, output) {
...
+    // factor symbol
     ['x^2 + 2x', 'x * (x + 2)'],
     ['-x^2 - 2x', '-x * (x + 2)'],
     ['x^2 - 3x', 'x * (x - 3)'],
-    ['2x^2 + 2x', '2x * (x + 1)'],
+    ['2x^2 + 4x', '2x * (x + 2)'],
+    // difference of squares
     ['x^2 - 4', '(x + 2) * (x - 2)'],
-    ['x^2 + 4', 'x^2 + 4'],
+    ['-x^2 + 1', '-(x + 1) * (x - 1)'],
```
evykassirer: so many more tests yayy

test/factor/factorQuadratic.test.js
```diff
@@ -7,28 +7,42 @@ function testFactorQuadratic(input, output) {
...
+    ['4x^2 - 9', '(2x + 3) * (2x - 3)'],
+    ['4x^2 - 16', '4 * (x + 2) * (x - 2)'],
+    ['-4x^2 + 16', '-4 * (x + 2) * (x - 2)'],
```
evykassirer: this is a great case

index.js
```diff
@@ -1,8 +1,10 @@
 const ChangeTypes = require('./lib/ChangeTypes');
+const factor = require('./lib/factor');
 const simplifyExpression = require('./lib/simplifyExpression');
 const solveEquation = require('./lib/solveEquation');
 
 module.exports = {
+  factor,
```
aelnaiem: So the reasoning I had was that you can't factor an equation, you can solve for x. So I would expect us to factor within solveEquation when solving an equation that can only be solved by factoring (a quadratic equation)

lib/factor/factorQuadratic.js
```diff
@@ -192,22 +197,62 @@ function factorPerfectSquare(node, symbol, aValue, bValue, cValue, negate) {
...
+        const secondGroup = [cValue, pair[1]];
+        const firstGroupGcd = math.gcd(...firstGroup);
+        const secondGroupGcd = math.gcd(...secondGroup);
+
+        // factor the first group: e.g. 2x^2 + 4x -> 2x * (x + 1)
+        const firstCoeffValue = Node.Creator.constant(firstGroupGcd);
+        const firstCoeff = Node.Creator.polynomialTerm(
+          symbol, null, firstCoeffValue);
+
+        const firstTermInGroupCoeff = Node.Creator.constant(aValue/firstGroupGcd);
```
aelnaiem: Yea, I played around with something like that too, but I agree it feels just as confusing. A comment at the top that explains it all, could be the trick though. Let me try that.

test/factor/factor.test.js
aelnaiem: No it's actually more like the `solveEquation.test.js` and `oneStep.test.js`

lib/factor/factorQuadratic.js
aelnaiem: I agree, but will leave a comment for this to be added at a later time (with an appropriate issue). I looked into doing this initially and it wasn't simple. Similarly above, I think that factoring out the gcd should be a substep for the other functions.

lib/factor/factorQuadratic.js
aelnaiem: Also, why u, r and t specifically?

lib/factor/factorQuadratic.js
```diff
@@ -112,9 +112,13 @@ function factorDifferenceOfSquares(node, symbol, aValue, bValue, cValue, negate)
     return Node.Status.noChange(node);
   }
 
+  // we factor out the gcd first, providing us with a modified expression to
+  // factor with new a and c values
```
evykassirer: yes! this helps a lot :)

lib/factor/factorQuadratic.js
```diff
@@ -246,12 +254,11 @@ function factorSumProductRule(node, symbol, aValue, bValue, cValue, negate) {
         const secondParen = Node.Creator.parenthesis(
           Node.Creator.operator('+', [firstCoeff, secondCoeff]));
 
-
         // create a node in the general factored form for expression
         let newNode = Node.Creator.operator('*', [firstParen, secondParen], true);
         if (gcd !== 1) {
           const gcdNode = Node.Creator.constant(gcd);
-          newNode = Node.Creator.operator('*', [gcdNode, newNode], true);
+          newNode = Node.Creator.operator('*', [gcdNode, firstParen, secondParen], true);
```
evykassirer: since it's this instead, I think it should be an if / else for if gcd is 1 or not

also I think this line is too long now

lib/factor/factorQuadratic.js
evykassirer: no particular reason, they just weren't being used yet

you haven't added this comment yet,right?

test/factor/factor.test.js
evykassirer: is it close enough that we can DRY with TestUtil?

lib/factor/factorQuadratic.js
```diff
@@ -222,44 +224,55 @@ function factorSumProductRule(node, symbol, aValue, bValue, cValue, negate) {
...
-          symbol, null, firstTermInGroupCoeff);
-        const secondTermInGroupValue = pair[0]/firstGroupGcd;
-        const secondTermInGroup = Node.Creator.constant(secondTermInGroupValue);
-
-        // factor the second group: e.g. 3x + 3 -> 3(x + 1)
-        // Since (x + 1) is the same after factoring both groups, we only find it once
-        let secondCoeffValue = secondGroupGcd;
-        if (pair[1] < 0) {
-          secondCoeffValue = secondCoeffValue * -1;
+        // To factor, we go through some transformations (these should be actual
```
evykassirer: can you explicitly add a TODO for this?

lib/factor/factorQuadratic.js
```diff
@@ -222,44 +224,55 @@ function factorSumProductRule(node, symbol, aValue, bValue, cValue, negate) {
...
+        // substeps in the future)
+        // 1. Break apart the middle term into two terms using our factor pair (p and q):
+        //    e.g. ax^2 + bx + c -> ax^2 + px + qx + c
+        // 2. Consider the first two terms together and the second two terms
+        //    together (this doesn't require any actual change to the expression)
+        //    e.g. first group: [ax^2 + px] and second group: [qx + c]
+        // 3. Factor both groups separately
+        //    e.g first group: [ux(rx + s)] and second group [v(rx + x)]
```
evykassirer: second group [v(rx + **s**)]

lib/factor/factorQuadratic.js
```diff
@@ -222,44 +224,55 @@ function factorSumProductRule(node, symbol, aValue, bValue, cValue, negate) {
...
+        const p = pair[0];
+        const q = pair[1];
+
+        // factor the first group to get u, r and s:
+        // u = gcd(a, p), r = a/u and s = p/u
+        const u = Node.Creator.constant(math.gcd(aValue, p));
+        const r = Node.Creator.constant(aValue/u);
+        const s = Node.Creator.constant(p/u);
+
+        // create the first parenthesis
```
evykassirer: > create the first group (rx + s)

^ better comment?

lib/factor/factorQuadratic.js
```diff
@@ -222,44 +224,55 @@ function factorSumProductRule(node, symbol, aValue, bValue, cValue, negate) {
...
+        // 1. Break apart the middle term into two terms using our factor pair (p and q):
+        //    e.g. ax^2 + bx + c -> ax^2 + px + qx + c
+        // 2. Consider the first two terms together and the second two terms
+        //    together (this doesn't require any actual change to the expression)
+        //    e.g. first group: [ax^2 + px] and second group: [qx + c]
+        // 3. Factor both groups separately
+        //    e.g first group: [ux(rx + s)] and second group [v(rx + x)]
+        // 4. Finish factoring by combining the factored terms through grouping:
+        //    e.g. (ux + v)(rx + s)
+        const p = pair[0];
```
evykassirer: consider calling these pvalue and qValue just cause they're numbers and not nodes

evykassirer:
ahhh this is so much nicer ^_^ good idea, me

Looks fine but maybe change some of the comments/naming. Go ahead and merge!

PR #151, opened 2017-04-04
dont change the change type for simplifying the left and right side for one step

aelnaiem:
If when simplifying the left or right side, there is only one step, we should just use the change_type of that step. Currently we change the change type to"Simplify left side" or "Simplify right side".

This is simpler and more useful


evykassirer:
I might have done this for some reason - but it would have been related to Athena (Socratic's private repo that uses this) so I'm okay merging this

thanks!

PR #153, opened 2017-04-13
Multiply powers integers

sebastianalbert:
Better pedagogical approach for multiplication of powers.


sebastianalbert:
Pretty much it takes calculations such as `10^2 * 10^3` and returns a Node with the same substeps as it does when it calculates `x^2 * x^3`

evykassirer:
these pull requests are super exciting! I'm in the middle of exam studying time, so I'm a bit slower on looking at this than I want to be, but I should be able to do a pass in the next day or two :D 

(it's near the top of my list of random stuff to do)

sebastianalbert:
No hurry! Good luck with your exams :D

.idea/codeStyleSettings.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
evykassirer: I think these files in `.idea` were your own playing around and not meant to be part of this PR - if you meant to include them can you explain what/why?

.npmignore
```diff
@@ -0,0 +1,4 @@
+*swp
```
evykassirer: oh - thanks for thinking of this!

I just did a search around and turns out you don't need this cause 

> If there's no .npmignore file, but there is a .gitignore file, then npm will ignore the stuff matched by the .gitignore file.

(from https://docs.npmjs.com/misc/developers)

lib/TreeSearch.js
```diff
@@ -21,14 +21,12 @@ TreeSearch.postOrder = function(simplificationFunction) {
 // A helper function for performing a tree search with a function
 function search(simplificationFunction, node, preOrder) {
   let status;
-
```
evykassirer: can you add these spaces back in? I like them for code readability, and they're also changing files you didn't change otherwise which is confusing to include in this PR for future investigators :)

(same for the other deleted and added lines - you can see where they are below in the diff)

lib/checks/canMultiplyLikeTermConstantNodes.js
```diff
@@ -0,0 +1,30 @@
+const Node = require('../node');
+
+
```
evykassirer: just to match what other files are doing right now, can you do only one new line between the require and the comment, and no new lines at all between the comment and function? ^_^

lib/checks/canMultiplyLikeTermConstantNodes.js
```diff
@@ -0,0 +1,30 @@
...
+// Returns true if the nodes are constant terms with the same constant and has an exponent each
```
evykassirer: two comments on your comment (hehe)

1. make sure lines are max 80 char (there are some exceptions, but I like to try to keep it that way)
2. it's often super hard to talk about math nodes clearly (I often write things that are confusing and other ppl have to point it out) but what do you think about "Returns true if `node` is a multiplication of constant nodes where can combine their exponents, e.g. 10^2 * 10^4 * 10 can become 10^7"

wow that was hard for me to rewrite too >.> I think describing it in accurate detail is too long to be worth it, but the example will help and be enough

what do you think?

lib/checks/canMultiplyLikeTermConstantNodes.js
```diff
@@ -0,0 +1,30 @@
...
+
+function canMultiplyLikeTermConstantNodes(node) {
+  if (!Node.Type.isOperator(node) || node.op !== '*' || node.args[1].op === '/') {
```
evykassirer: why do you have `node.args[1].op === '/'`?

lib/checks/canMultiplyLikeTermConstantNodes.js
```diff
@@ -0,0 +1,30 @@
...
+    return false;
+  }
+  const args = node.args;
+  if (args.some(n=> Node.PolynomialTerm.isPolynomialTerm(n))) {
```
evykassirer: so I'm not sure if this is even worth asking cause it's so small, but since you're updating this file anyways can you add a space before `=>` to match the rest of em?

lib/simplifyExpression/stepThrough.js
```diff
@@ -72,11 +72,12 @@ function step(node) {
     divisionSearch,
     // Adding fractions, cancelling out things in fractions
     fractionsSearch,
-    // e.g. 2 + 2 => 4
-    arithmeticSearch,
     // e.g. addition: 2x + 4x^2 + x => 4x^2 + 3x
     // e.g. multiplication: 2x * x * x^2 => 2x^3
+    // e.g multiplication: 10^3 * 10^2 => 10^5
```
evykassirer: probably should make this more obvious what the difference is now

```
// e.g. addition of polynomial terms: 2x + 4x^2 + x => 4x^2 + 3x
// e.g. multiplication of polynomial terms: 2x * x * x^2 => 2x^3		      
// e.g. multiplication of constants: 10^3 * 10^2 => 10^5
```

lib/checks/canMultiplyLikeTermConstantNodes.js
```diff
@@ -0,0 +1,30 @@
...
+    return false;
+  }
+  if (!args.every(n => Node.ConstantTerms.isConstantTerm(n))) {
```
evykassirer: hmmmmmm I think we make this a lot simpler!

I see you're basing this around `PolynomialTerm` which is definitely an intuitive way to think about this

`PolynomialTerm` is used in a lot of places and has coefficients, whereas I'm not sure if `ConstantTerm` is as clear of a concept and I don't think it needs to have a coefficient (`2*5^2` is just a multiplication of two things, whereas `2x^2` is more of a single unit, if that makes sense)

my recommendation is to have a helper function that returns true if the node is a constant with the right value or an exponent node where the base is a constant with the right value - that's all we need right?

let me know what ya think - it'd be a bit of a redesign from your current solution

lib/node/PolynomialTerm.js
```diff
@@ -36,6 +36,7 @@ class PolynomialTerm {
         if (node.args.length !== 2) {
           throw Error('Expected two arguments to *');
         }
+
```
evykassirer: (same with removing these spaces)

lib/simplifyExpression/collectAndCombineSearch/multiplyLikeTerms.js
```diff
@@ -5,23 +5,20 @@ const multiplyFractionsSearch = require('../multiplyFractionsSearch');
 
 const ChangeTypes = require('../../ChangeTypes');
 const Node = require('../../node');
-
 // Multiplies a list of nodes that are polynomial like terms. Returns a node.
 // The polynomial nodes should *not* have coefficients. (multiplying
 // coefficients is handled in collecting like terms for multiplication)
-function multiplyLikeTerms(node, polynomialOnly=false) {
+function multiplyLikeTerms(node, polynomialOnly = false) {
```
evykassirer: From what I remember, the lack of spaces is a special syntax thing to show that it's a default value for `polynomialOnly` instead of the usual use of `=`

lib/simplifyExpression/collectAndCombineSearch/multiplyLikeTerms.js
```diff
@@ -93,51 +87,82 @@ function multiplyPolynomialTerms(node) {
...
     return Node.Status.nodeChanged(
-        ChangeTypes.ADD_EXPONENT_OF_ONE, node, newNode, false);
+      ChangeTypes.ADD_EXPONENT_OF_ONE, node, newNode, false);
   }
   else {
     return Node.Status.noChange(node);
   }
 }
-
 // Given a product of polynomial terms, groups the exponents into a sum
```
evykassirer: if you keep it this way, you should update this comment

though, related - do you think it'd be clearer to have separate `collectPolynomialExponents` and `collectConstantExponents` functions, since they have fully separate code?

lib/simplifyExpression/collectAndCombineSearch/multiplyLikeTerms.js
```diff
@@ -93,51 +87,82 @@ function multiplyPolynomialTerms(node) {
...
+    // symbol. Get that from the first node.
+    const symbolNode = polynomialTermList[0].getSymbolNode();
+    // The new exponent will be a sum of exponents (an operation, wrapped in
+    // parens) e.g. x^(3+4+5)
+    const exponentNodeList = polynomialTermList.map(p => p.getExponentNode(true));
+    const newExponent = Node.Creator.parenthesis(
+      Node.Creator.operator('+', exponentNodeList));
+    const newNode = Node.Creator.polynomialTerm(symbolNode, newExponent, null);
+    return Node.Status.nodeChanged(
+      ChangeTypes.COLLECT_EXPONENTS, node, newNode);
```
evykassirer: I think it'd be nice to have a new ChangeType for this, just to clarify that it's two different cases (on polynomial terms and on constants)

lib/simplifyExpression/collectAndCombineSearch/multiplyLikeTerms.js
```diff
@@ -93,51 +87,82 @@ function multiplyPolynomialTerms(node) {
...
+        change = true;
+        changeGroup++;
+      }
+    });
+  }
 
   if (change) {
     return Node.Status.nodeChanged(
-        ChangeTypes.ADD_EXPONENT_OF_ONE, node, newNode, false);
+      ChangeTypes.ADD_EXPONENT_OF_ONE, node, newNode, false);
```
evykassirer: Oh nice! Thanks for fixing that indentation. I'm not sure why the linter didn't notice it before

sebastianalbert:
For the .idea and weird spaces and stuff, thats probably just my WebStorm acting weird. I will fix that!

lib/simplifyExpression/collectAndCombineSearch/multiplyLikeTerms.js
sebastianalbert: That's true, i wanted to make one but was afraid of making something else not work haha. But I guess its an easy fix

evykassirer: no worries - if things don't work and it gets annoying I'm happy to help troubleshoot

.npmignore
sebastianalbert: Oh! The more you know :D

evykassirer: we both learned something today, haha

lib/checks/canMultiplyLikeTermConstantNodes.js
sebastianalbert: That should be clear enough! Great!

lib/checks/canMultiplyLikeTermConstantNodes.js
sebastianalbert: Because it failed One test and couldnt think of any other way of fixing it. Although i changed some stuff after so i will give a try without it!

lib/checks/canMultiplyLikeTermConstantNodes.js
sebastianalbert: Hmm maybe! But what is a large number written as 3*10^6 and have to multiply this with another large number written the same way. Then we might need to  look at the "coefficent" :) but a helper function could be enough

lib/simplifyExpression/collectAndCombineSearch/multiplyLikeTerms.js
sebastianalbert: That's true, that would be a lot clearer. Easy fix, nice comment!

lib/checks/canMultiplyLikeTermConstantNodes.js
evykassirer: I think in this case we'd want to do a "collect like terms" where for example:

3 * 10^6 * 5 * 10^2 => 3 * 5 * (10^6 * 10^2) and then we combine exponents

this already happens with symbols in cases like

`y * x^6 * z * x^2 => y * z * (x^6 * x^2)`

so I'd suggest following that kind of pattern :)

lib/simplifyExpression/collectAndCombineSearch/multiplyLikeTerms.js
evykassirer: yay!

lib/checks/canMultiplyLikeTermConstantNodes.js
sebastianalbert: Oh, of course. I didnt think of that. That should be possible. I will take a look into that!

sebastianalbert:
I will take a look at it in more detail in the next few days! Now i have to sleep haha, its 02.30 here.. But great comments! So easy to miss things with so much code.

evykassirer:
Good night :)

Glad they were helpful!

lib/simplifyExpression/collectAndCombineSearch/multiplyLikeTerms.js
sebastianalbert: Hmm, been thinking about this some more. If we add those two functions then the call for them in the code above will be more unclear, so it still might be better to keep as it is. That is to have one function with two cases.

lib/checks/canMultiplyLikeTermConstantNodes.js
sebastianalbert: Okay, I've been looking into this too and I'm not quite sure how to rewrite it. I'm not sure than an helper function will be enough, because I have to contruct the power-node somehow and this is currently done in `ConstantTerm` since it cant be done in `PolynomialTerm`. But i agree about coefficient thing, that's not needed.

lib/simplifyExpression/collectAndCombineSearch/multiplyLikeTerms.js
evykassirer: why not do the check before calling the function instead of inside it?

lib/checks/canMultiplyLikeTermConstantNodes.js
evykassirer: can you explain what you mean about constructing the power node? I think because it's a lot simpler structure than Polynomial Terms you could just build it yourself. But I know that stuff is confusing, so I can help you figure out that code and stuff if you want!

lib/ChangeTypes.js
```diff
@@ -42,6 +42,12 @@ module.exports = {
   // e.g. x + 2 + x^2 + x + 4 -> x^2 + (x + x) + (4 + 2)
   COLLECT_LIKE_TERMS: 'COLLECT_LIKE_TERMS',
 
+  // MULTIPLYING/DIVIDING CONSTANT POWERS
+
+  // e.g. 10^2 * 10^3 -> 10^(2+3)
+  // e.g. 10^4 / 10^2 -> 10^(4-2)
+  COLLECT_CONSTANT_EXPONENTS: 'COLLECT_CONSTANT_EXPONENTS',
```
evykassirer: just have the multiply one here because division is in the other PR - you can add the division one in the other PR

sebastianalbert: Of course, i was too fast when copying that change from the other branch.

lib/simplifyExpression/collectAndCombineSearch/multiplyLikeTerms.js
sebastianalbert: Smart! I thought it looked a little messy but maybe its clearer, I'll try.

lib/checks/canMultiplyLikeTermConstantNodes.js
sebastianalbert: I need something which can return the base and the exponent of the power, which is now done in `ConstantTerms`. This is needed to collect the exponents and getting the value of the base. I'm not quite sure how to do this in another way or where to place the code for that if not in `ConstantTerms`. I rewrote it a little bit and deleted lots of unnecessary code in it. So now it creates a power node with base and exponent if all terms are constant and the operation is multiplication.

sebastianalbert: It could probably be called PowerTerms instead to be a little more clear.

evykassirer: I think a helper function that takes something of the form c^n or c can return the base and exponent

e.g. 10^3 => returns 10 and 3
e.g. 10 => returns 10 and 1

with polynomial terms there's also a coefficient which makes everything much more messy which is why there's a whole class for it

sebastianalbert: Oh! So just a helper function in `canMultiplyLikeTermsConstantNode` that does this?

evykassirer: yeah I think something like that would work

.idea/codeStyleSettings.xml
evykassirer: can you remove these files?

.npmignore
evykassirer: and this one!

lib/ChangeTypes.js
```diff
@@ -42,6 +42,11 @@ module.exports = {
   // e.g. x + 2 + x^2 + x + 4 -> x^2 + (x + x) + (4 + 2)
   COLLECT_LIKE_TERMS: 'COLLECT_LIKE_TERMS',
 
+  // MULTIPLYING CONSTANT POWERS
+
```
evykassirer: don't need this extra ine

lib/TreeSearch.js
```diff
@@ -28,7 +28,6 @@ function search(simplificationFunction, node, preOrder) {
       return status;
     }
   }
-
```
evykassirer: don't need to delete this line

lib/checks/canAddLikeTermPolynomialNodes.js
```diff
@@ -19,7 +19,6 @@ function canAddLikeTermPolynomialNodes(node) {
   const firstTerm = polynomialTermList[0];
   const sharedSymbol = firstTerm.getSymbolName();
   const sharedExponentNode = firstTerm.getExponentNode(true);
-
```
evykassirer: don't need to delete this line

lib/checks/canMultiplyLikeTermConstantNodes.js
```diff
@@ -0,0 +1,63 @@
+const Node = require('../node');
+const NodeType = require('../node/Type');
+
+// Returns true if node is a multiplication of constant power nodes
```
evykassirer: hmm maybe clarify they don't all have to have exponents?

lib/checks/canMultiplyLikeTermConstantNodes.js
```diff
@@ -0,0 +1,63 @@
...
+// where you can combine their exponents, e.g. 10^2 * 10^4 * 10 can become 10^7
+
```
evykassirer: no extra space here :)

lib/checks/canMultiplyLikeTermConstantNodes.js
```diff
@@ -0,0 +1,63 @@
...
+function canMultiplyLikeTermConstantNodes(node) {
+  if (!Node.Type.isOperator(node) || node.op !== '*') {
+    return false;
+  }
+  const args = node.args;
+  if (!args.every(n => isPowerTerm(n))) {
```
evykassirer: `isConstantOrConstantPower`? hmm that's kinda long, but I think adding more detail to the name would be better. maybe you can think of something with more detail but not that long (if you don't like my name)

lib/checks/canMultiplyLikeTermConstantNodes.js
```diff
@@ -0,0 +1,63 @@
...
+    return false;
+  }
+  const constantTermList = node.args.map(n => createPowerNode(n));
+  const firstTerm = constantTermList[0];
+  const restTerms = constantTermList.slice(1);
+  // they're considered like terms if they have the same base value
+  return restTerms.every(term => getBaseValue(firstTerm) === getBaseValue(term));
+}
+function createPowerNode(node) {
```
evykassirer: new line before this function declaration

lib/checks/canMultiplyLikeTermConstantNodes.js
```diff
@@ -0,0 +1,63 @@
...
+  }
+  catch (err) {
+    return false;
+  }
+}
+module.exports = {
+  canMultiplyLikeTermConstantNodes,
+  createPowerNode
+};
+
```
evykassirer: don't need this extra line

lib/checks/canMultiplyLikeTermConstantNodes.js
```diff
@@ -0,0 +1,63 @@
...
+  }
+}
+
+function getBaseValue(node) {
+  return node.base.value;
+}
+
+function isPowerTerm(node) {
+  try {
+    // will throw error if node isn't constant power term
```
evykassirer: I don't think you need to go the error throwing route here - that was a not super great solution to some complicated issue with polynomial terms

the check for if something is a polynomial term is just (I think)

```
if ((NodeType.isOperator(node, '^') && NodeType.isConstant(node.args[0])) ||
    NodeType.isConstant(node))
```

lib/checks/canMultiplyLikeTermConstantNodes.js
```diff
@@ -0,0 +1,63 @@
...
+  return restTerms.every(term => getBaseValue(firstTerm) === getBaseValue(term));
+}
+function createPowerNode(node) {
+  if (NodeType.isOperator(node)) {
+    if (node.op === '^') {
+      const constantNode = node.args[0];
+      if (!NodeType.isConstant(constantNode)) {
+        throw Error('Expected constant term, got ' + constantNode);
+      }
+      node.base = constantNode;
```
evykassirer: I don't think it's a great idea to add extra fields to the mathjs node like that

How about this: make two functions that are given a node which is a constant or a constant with an exponent

one returns the base (or the node itself) and the other returns the exponent (or the constant 1 if there's no exponent)

evykassirer: let me know if you think that's better, or if it's more confusing :)

lib/checks/canMultiplyLikeTermPolynomialNodes.js
```diff
@@ -13,7 +13,6 @@ function canMultiplyLikeTermPolynomialNodes(node) {
   if (args.length === 1) {
     return false;
   }
-
```
evykassirer: careful with these extra lines added/removed - if you scan through the diff you should be able to see them all without me having to comment

lib/node/Creator.js
```diff
@@ -71,6 +71,13 @@ const NodeCreator = {
   nthRoot (radicandNode, rootNode) {
     const symbol = NodeCreator.symbol('nthRoot');
     return new math.expression.node.FunctionNode(symbol, [radicandNode, rootNode]);
+  },
+  powerTerm (base, exponent) {
```
evykassirer: this is better than the constantTerm, but I still think we can get away with not having be a node type at all. maybe that'll get too complicated though. feel free to message me on gitter and we can chat back and forth more if you want!

https://gitter.im/mathsteps-chat/Lobby

test/simplifyExpression/collectAndCombineSearch/collectAndCombineSearch.test.js
```diff
@@ -52,6 +52,21 @@ describe('combinePolynomialTerms addition', function() {
   ];
   tests.forEach(t => testCollectAndCombineSubsteps(t[0], t[1]));
 });
+describe('combineConstantPowerTerms multiplication', function() {
+  const tests = [
+    ['10^2 * 10',
```
evykassirer: yay these tests look awesome thanks!

lib/simplifyExpression/collectAndCombineSearch/multiplyLikeTerms.js
```diff
@@ -141,13 +146,12 @@ function addOneExponent(node) {
 // e.g. 10^2 * 10^3 -> 10^(2+3)
 // Returns a Node.Status object.
 function collectConstantExponents(node) {
-  const constantTermList = node.args.map(n => new checks.canMultiplyLikeTermConstantNodes.createPowerNode(n));
   // If we're multiplying constant nodes together, they all share the same
   // base. Get that from the first node.
-  const baseNode = constantTermList[0].base;
+  const baseNode = checks.canMultiplyLikeTermConstantNodes.getBaseNode(node.args[0]);
```
sebastianalbert: Yeah, I also think that making a new module is the way. Because it will also be used for when dividing power terms, so a new module will be neater than having the same code in two different modules.

lib/simplifyExpression/collectAndCombineSearch/multiplyLikeTerms.js
```diff
@@ -39,7 +41,7 @@ function multiplyLikeTerms(node, polynomialOnly=false) {
 }
 
 function multiplyPolynomialTerms(node) {
-  if (!checks.canMultiplyLikeTermPolynomialNodes(node) && !checks.canMultiplyLikeTermConstantNodes.canMultiplyLikeTermConstantNodes(node)) {
+  if (!checks.canMultiplyLikeTermPolynomialNodes(node) && !checks.canMultiplyLikeTermConstantNodes(node)) {
```
evykassirer: this line is >80 chars

lib/simplifyExpression/collectAndCombineSearch/ConstantOrPowerTerm.js
```diff
@@ -0,0 +1,45 @@
+const NodeType = require('../../node/Type');
```
evykassirer: consider calling this file `ConstantOrConstantPower` even though it's longer? especially since your checker function is `isConstantOrConstantPower`

just because now it seems like it's either a constant term, or a "power" term, where power is kinda ambiguous

lib/checks/canMultiplyLikeTermConstantNodes.js
```diff
@@ -9,52 +9,14 @@ function canMultiplyLikeTermConstantNodes(node) {
...
   }
   const args = node.args;
-  if (!args.every(n => isConstantOrConstantPower(n))) {
+  if (!args.every(n => ConstantOrPowerTerm.isConstantOrConstantPower(n))) {
     return false;
   }
-  const constantTermList = node.map(n => getBaseNode(n));
-  const firstTerm = constantTermList.args[0];
-  const restTerms = constantTermList.args.slice(1);
+  const constantTermList = node.args.map(n => ConstantOrPowerTerm.getBaseNode(n));
```
evykassirer: curious - what kind of tests would have failed before you fixed this? or does this code do the same thing?

if this would have broken something before, see if you can add at test that only passes with the fixed code

lib/simplifyExpression/collectAndCombineSearch/ConstantOrPowerTerm.js
```diff
@@ -0,0 +1,45 @@
...
+  return node.value;
+}
+
+// Returns the node that is an exponent to a constant, or a constant node with
+// value 1 if there's no exponent.
+// e.g. on the node representing 2^3, returns a constant node with value 3
+// e.g. on the node representing 2, returns a constant node with value 1 (since 2 = 2^1)
+function getExponentNode(node) {
+  if (node.args === undefined || !node.args[1]) {
+    return false;
```
sebastianalbert: If it doesnt return false, then it for some reason will skip the part with adding the constant 1 as exponent in multiplyLikeTerms which means that it will skip substeps, not good.

lib/checks/canMultiplyLikeTermConstantNodes.js
sebastianalbert: No it does the same thing, dont know why I wrote it the first way haha. Instead of looking at the args for the map I just checked them after, so it does the same thing now but much nicer.

lib/simplifyExpression/collectAndCombineSearch/ConstantOrPowerTerm.js
evykassirer: ohh okay instead of returning false, return the constant node 1 (like in your comment above)

also the comment is longer than 80 characters :p can you install something in your text editor so you can see when lines go over?

lib/simplifyExpression/collectAndCombineSearch/ConstantOrConstantPower.js
```diff
@@ -1,8 +1,17 @@
...
+// and the exponent of the power, it can also check if the node is constant or
+// constant power.
+// e.g 2^10 is an constant power, while x^10 is not
+// e.g 2 is an constant, while x is not
+
+// Returns the base if the node is on power form
+// else returns the node as it is constant.
 // e.g 2^4 returns 2
-// e.g 3 returns 3, as that is the node
+// e.g 3 returns 3, since 3 is equal to 3^1 which has a base of 3
```
evykassirer: very clear comment, nice

.idea/codeStyleSettings.xml
evykassirer: these are still here - please remove them!

lib/TreeSearch.js
```diff
@@ -3,15 +3,15 @@ const Node = require('./node');
 const TreeSearch = {};
 
 // Returns a function that performs a preorder search on the tree for the given
-// simplifcation function
+// simplification function
```
evykassirer: thanksss ^_^ i bet there are so many typos in this codebase

lib/checks/canMultiplyLikeTermConstantNodes.js
```diff
@@ -0,0 +1,22 @@
...
+// The node can either be on form c^n or c, as long as c is the same for all.
+function canMultiplyLikeTermConstantNodes(node) {
+  if (!Node.Type.isOperator(node) || node.op !== '*') {
+    return false;
+  }
+  const args = node.args;
+  if (!args.every(n => ConstantOrPowerTerm.isConstantOrConstantPower(n))) {
+    return false;
+  }
+  const constantTermList = node.args.map(n => ConstantOrPowerTerm.getBaseNode(n));
```
evykassirer: actually you can just use `args` here since you've already defined it

evykassirer: can you rename `constantTermList` to `constantTermBaseList` to be more accurately descriptive?

lib/checks/canMultiplyLikeTermPolynomialNodes.js
evykassirer: go to https://github.com/socraticorg/mathsteps/pull/153/files and look for all the remaining lines that need to be added back in

lib/checks/index.js
```diff
@@ -1,13 +1,16 @@
 const canAddLikeTermPolynomialNodes = require('./canAddLikeTermPolynomialNodes');
+const canMultiplyLikeTermConstantNodes = require('./canMultiplyLikeTermConstantNodes');
 const canMultiplyLikeTermPolynomialNodes = require('./canMultiplyLikeTermPolynomialNodes');
 const canRearrangeCoefficient = require('./canRearrangeCoefficient');
 const canSimplifyPolynomialTerms = require('./canSimplifyPolynomialTerms');
 const hasUnsupportedNodes = require('./hasUnsupportedNodes');
 const isQuadratic = require('./isQuadratic');
 const resolvesToConstant = require('./resolvesToConstant');
 
+
```
evykassirer: or removed

lib/node/Creator.js
evykassirer: can you get rid of this now? and just use the getter functions

sebastianalbert: Sadly, no. It is still needed in multiplyLikeTerms... hard to create the args for the node else.

evykassirer: are you sure you can't just use the getter functions instead?

evykassirer: i only see you using it in two places, where you could just be creating a ^ node instead

am I missing something?

lib/simplifyExpression/collectAndCombineSearch/ConstantOrPowerTerm.js
sebastianalbert: but if I return constant node 1, it will skip the addOneExponent because of (!ConstantOrPowerTerm.getExponentNode(child)). So I dont really have to return a constant node 1 here, because it will be fixed in addOneExponent.

evykassirer: ohhh so there's a step for adding the exponent of 1 in the substeps? I must have missed that - that sounds like the right thing to do

in that case, you should just returns null here (not False) and update the comment

lib/simplifyExpression/collectAndCombineSearch/ConstantOrConstantPower.js
```diff
@@ -24,10 +25,10 @@ function getBaseNode(node) {
 // Returns the node that is an exponent to a constant, or a constant node with
 // value 1 if there's no exponent.
 // e.g. on the node representing 2^3, returns a constant node with value 3
-// e.g. on the node representing 2, returns a constant node with value 1 (2 = 2^1)
+// e.g 3 returns 3, since 3 is equal to 3^1 which has a base of 3
```
evykassirer: should be "3 returns 1"

lib/simplifyExpression/collectAndCombineSearch/multiplyLikeTerms.js
```diff
@@ -106,10 +108,10 @@ function addOneExponent(node) {
   let changeGroup = 1;
   if (checks.canMultiplyLikeTermConstantNodes(node)) {
     newNode.args.forEach((child, i) => {
-      if (!ConstantOrPowerTerm.getExponentNode(child)) {
-        newNode.args[i] = Node.Creator.powerTerm(
-          ConstantOrPowerTerm.getBaseNode(child),
-          Node.Creator.constant(1));
+      if (child.args === undefined) {
```
evykassirer: can you add a comment to explain when this is the case?

lib/simplifyExpression/collectAndCombineSearch/multiplyLikeTerms.js
```diff
@@ -145,22 +147,24 @@ function addOneExponent(node) {
...
 // Returns a Node.Status object.
 function collectConstantExponents(node) {
   // If we're multiplying constant nodes together, they all share the same
   // base. Get that from the first node.
-  const baseNode = ConstantOrPowerTerm.getBaseNode(node.args[0]);
+  const baseNode = ConstantOrConstantPower.getBaseNode(node.args[0]);
   // The new exponent will be a sum of exponents (an operation, wrapped in
   // parens) e.g. 10^(3+4+5)
-  const exponentNodeList = node.args.map(p => ConstantOrPowerTerm.getExponentNode(p));
+  const exponentNodeList = node.args.map(p => ConstantOrConstantPower.getExponentNode(p));
```
evykassirer: > 80 chars

.npmignore
evykassirer: still here

lib/checks/canMultiplyLikeTermPolynomialNodes.js
evykassirer: still

lib/node/Creator.js
```diff
@@ -71,7 +71,7 @@ const NodeCreator = {
   nthRoot (radicandNode, rootNode) {
     const symbol = NodeCreator.symbol('nthRoot');
     return new math.expression.node.FunctionNode(symbol, [radicandNode, rootNode]);
-  }
+  },
```
evykassirer: don't need this

lib/simplifyExpression/collectAndCombineSearch/ConstantOrConstantPower.js
```diff
@@ -25,7 +25,7 @@ function getBaseNode(node) {
 // Returns the node that is an exponent to a constant, or a constant node with
 // value 1 if there's no exponent.
 // e.g. on the node representing 2^3, returns a constant node with value 3
-// e.g 3 returns 3, since 3 is equal to 3^1 which has a base of 3
+// e.g 3 returns 1, since 3 is equal to 3^1 which has a base of 3
```
evykassirer: oh, "which has an exponent of 1"

lib/simplifyExpression/collectAndCombineSearch/ConstantOrConstantPower.js
```diff
@@ -38,7 +38,8 @@ function getExponentNode(node) {
 // Checks if the node is an constant or a power with constant base.
 // e.g. 2^3 is a constant power node, 5 is a constant node, x and x^2 are not
 function isConstantOrConstantPower(node) {
-  return ((NodeType.isOperator(node, '^') && NodeType.isConstant(node.args[0])) ||
+  return ((NodeType.isOperator(node, '^') &&
+          NodeType.isConstant(node.args[0])) ||
```
evykassirer: wait why'd you change it?

I think 
```
(a and b) OR
c
```
is more clear than
```
(a and
b) OR
c
```

was it over 80 characters? it doesn't seeeem like it

lib/simplifyExpression/collectAndCombineSearch/multiplyLikeTerms.js
```diff
@@ -108,7 +108,7 @@ function addOneExponent(node) {
   let changeGroup = 1;
   if (checks.canMultiplyLikeTermConstantNodes(node)) {
     newNode.args.forEach((child, i) => {
-      if (child.args === undefined) {
+      if (child.args === undefined) { // true if child is a constant node, e.g 3
```
evykassirer: you could do `Node.Type.Constant(child)` then

lib/simplifyExpression/collectAndCombineSearch/multiplyLikeTerms.js
```diff
@@ -157,7 +157,8 @@ function collectConstantExponents(node) {
   const baseNode = ConstantOrConstantPower.getBaseNode(node.args[0]);
   // The new exponent will be a sum of exponents (an operation, wrapped in
   // parens) e.g. 10^(3+4+5)
-  const exponentNodeList = node.args.map(p => ConstantOrConstantPower.getExponentNode(p));
+  const exponentNodeList = node.args.map(
```
evykassirer: OH i just remembered you can do 

`node.args.map(ConstantOrConstantPower.getExponentNode)`

it's a special case when you just do `p => f(p)`, then you can just pass the function you're using `f` instead

.npmignore
evykassirer: please delete this file @sebastianalbert

lib/simplifyExpression/collectAndCombineSearch/ConstantOrConstantPower.js
sebastianalbert: It showed as 81 or 82 in webstorm haha. Can change it back

evykassirer: ohhh yeah nvm you're right my bad

can you do this then?

```
(a &&
 b) ||
c
```

note the extra space to line a and b up together - idk it's still kinda hard to read but I guess it's ok :p

lib/checks/canMultiplyLikeTermPolynomialNodes.js
evykassirer: @sebastianalbert

test/simplifyExpression/simplify.test.js
```diff
@@ -163,6 +163,14 @@ describe('handles unnecessary parens at root level', function() {
   tests.forEach(t => testSimplify(t[0], t[1], t[2]));
 });
 
+describe('multiplication of powers support', function(){
```
evykassirer: Actually I don't think you need these tests, since this would be the answer it'd get to anyways even if it didn't have multiplication of powers support - it'd just get to the answer a different way. And having the tests here might be confusing then

evykassirer:
I think it's pretty much ready to go!! Because I've looked at it for so long and it's a big change, @aliang8 do you think you could take a look before we merge?

test/simplifyExpression/simplify.test.js
sebastianalbert: True, you have a point!

evykassirer: do you think it makes sense to remove them then?

sebastianalbert: Yeah, it makes sense!

evykassirer:
🎉 thanks for your patience getting this in tip top shape! looking forward to getting the division one in too - probably will be a lot less work haha

sebastianalbert:
No problem! Lots of the things are the same with division so it should be quick! I Will make some changes this weekend and push it 😄

PR #155, opened 2017-04-13
Flow code

nitin42:
Added Flow code to the `./lib/node/Creator.js`


nitin42:
Still need to add type checking for the return type of the function. But this serve as a basic example of how we could use Flow.

PR #156, opened 2017-04-14
TypeScript Rewrite

elu00:
Hi all,
Here's my proof of concept for a TypeScript port. For @nitin42 's proposed port to Flow, see #155     
Here's what I've done so far:
- Updated all module imports/exports to TypeScript syntax.
- Fixed almost all TypeScript compiler errors directly from the port
- Refactored the `Node` class to `MathNode`
- All temporary objects ported to classes with static methods.
- Added static typings for `mathjs` functions from `DefinitelyTyped`.
- Added static typing for most `node` paramters
### To Do:
- Add tsconfig.json for automated compilation/syntax changing.
- Create library declaration file
- Add type annotations to the rest of the variables and parameters.
- Fix `Creator.ts`   
I'd welcome any feedback or comments!    
I'll also try to finish up the to-do list and the rest of the port within the next several weeks.
EDIT: Actually, given what I've learned so far going through the codebase and the all the other feature's I've learned about, I'll probably actually be completely rebasing and re-doing this port if we decide to go the way of TS instead of flow


kungfooman:
What a messy PR, just add JSDoc with a `jsconfig.json` for full Intellisense + typing:

```js
{
  "compilerOptions": {
    "checkJs": true,
    "strict": true
  }
}
```

Issue #76, opened 2017-01-20
Better break-up fraction


hmaurer:
I have no idea how to go about this in a general way, but currently break-up fraction isn't very smart. 

For example, `(2x+3)/(2x+2)` simplifies to `2 x / (2 x + 2) + 3 / (2 x + 2)`. Instead, it might be better if it did: `(2x+3)/(2x+2)` -> `(2x+2 + 1)/(2x+2)` -> `(2x+2)/(2x+2) + 1/(2x+2)` -> `1 + 1/(2x+2)`.

Thoughts?

evykassirer:
agreed! that'd be sweet.

we could start with a rule that handles only this case, where the denominator can be found in the numerator

so here you'd check if an x term is in the numerator and if a constant term is in the numerator, and then split it up into 2x + 2 and the rest of the numerator. 

I think we should keep the current break-up fraction functionality, but add this as a more specific check before we do the general one, so we do the better one instead if it's possible.

How's that sound?

eltonlaw:
Hi, if there's nobody working on this currently, could I take it?

evykassirer:
Yes! I'd love to get this stuff into mathsteps. 

It's definitely a more open ended problem, so starting with what I described sounds good, and I can take a look and then we can discuss from there? Happy to help if you have any questions about the code or how the new steps will work

And I'm definitely down to brainstorm further things than the case above if you want someone to bounce ideas around with

PS I noticed you're a UW student (heh) and if you're on campus and want a mathsteps sticker for getting involved, lemme know - I'm around until the 24th ^^

eltonlaw:
Hi, awesome, yes please can we brainstorm a bit?

I'm a bit stuck on the rule formulation though, how general would you like this rule to be? Like the main idea is by adding a new term we should be able to partially factor the equation into an integer right? It makes sense for linear variables, but would we want to extend that to n >=2 ? Like would this be the proper course of action: (2x^3+ 2)/(x^3+4x^2+2x+5) => 2 - (8x^2+4x+8)/(x^3+4x^2+2x+5). Would this be overcomplicating things?

Either way, I think I would go about first checking if the highest power variable in the numerator and denominator were == 1 (or n), then doing numerator - denominator to find out how many copies of the denominator can fit into the numerator. From there, we would get the integers(# of full fits) and then whatever remainder we had earlier divided by the denominator. 

[...]

PR #158, opened 2017-04-21
#76 - Better break-up fraction. 

eltonlaw:
#76 This is only solves the initial case posed of (2x+3)/(2x+2). The output is now
(2x+2)/(2x+2) + (1)/(2x+2).

### DIFF
--- a/lib/simplifyExpression/breakUpNumeratorSearch/index.js
+++ b/lib/simplifyExpression/breakUpNumeratorSearch/index.js
@@ -1,3 +1,4 @@
+const canFindDenominatorInNumerator = require('../../checks/canFindDenominatorInNumerator');
 const ChangeTypes = require('../../ChangeTypes');
 const Node = require('../../node');
 const TreeSearch = require('../../TreeSearch');
@@ -27,8 +28,33 @@ function breakUpNumerator(node) {
   // At this point, we know that node is a fraction and its numerator is a sum
   // of terms that can't be collected or combined, so we should break it up.
   const fractionList = [];
-  const denominator = node.args[1];
-  numerator.args.forEach(arg => {
+  let denominator = node.args[1];
+
+  // Check if we can add a constant to make the fraction nicer
+  // fraction e.g. (2+x)/(5+x) -> (5+x)/(5+x) + 3/(5+x)
+  if (canFindDenominatorInNumerator(node)) {
+    let denominatorParenRemoved = false;
+    if (Node.Type.isParenthesis(denominator)) {
+      denominatorParenRemoved = true;
+      denominator = denominator.content;
+    }
+    const newNumerator = [];
+
+    // The constant value difference between the numerator and the denominator
+    const numeratorConstant = parseInt(numerator.args[1].value);
+    const denominatorConstant = parseInt(denominator.args[1].value);
+    const addedConstant = numeratorConstant - denominatorConstant;
+
+    newNumerator.push(Node.Creator.constant(addedConstant));
+    newNumerator.push(denominator);
+
+    numerator = newNumerator;
+
+    if (denominatorParenRemoved) {
+      denominator = Node.Creator.parenthesis(denominator);
+    }
+  }
+  numerator.forEach(arg => {
     const newFraction = Node.Creator.operator('/', [arg, denominator]);
     newFraction.changeGroup = 1;
     fractionList.push(newFraction);
@@ -41,5 +67,4 @@ function breakUpNumerator(node) {
   return Node.Status.nodeChanged(
     ChangeTypes.BREAK_UP_FRACTION, node, newNode, false);
 }
-
 module.exports = search;
--- a/test/checks/checks.test.js
+++ b/test/checks/checks.test.js
@@ -29,3 +29,10 @@ describe('canSimplifyPolynomialTerms addition', function() {
   ];
   tests.forEach(t => testCanCombine(t[0], t[1]));
 });
+
+describe('canSimplifyPolynomialTerms denominator in numerator', function() {
+  const tests = [
+    ['(2x + 3)/(2x + 2)', true],
+  ];
+  tests.forEach(t => testCanCombine(t[0], t[1]));
+});
--- a/test/simplifyExpression/breakUpNumeratorSearch/breakUpNumeratorSearch.test.js
+++ b/test/simplifyExpression/breakUpNumeratorSearch/breakUpNumeratorSearch.test.js
@@ -11,6 +11,7 @@ describe('breakUpNumerator', function() {
     ['(x+3+y)/3', '(x / 3 + 3/3 + y / 3)'],
     ['(2+x)/4', '(2/4 + x / 4)'],
     ['2(x+3)/3', '2 * (x / 3 + 3/3)'],
+    ['(2x + 3)/(2x + 2)', '(1 / (2x + 2) + (2x + 2) / (2x + 2))'],
   ];
   tests.forEach(t => testBreakUpNumeratorSearch(t[0], t[1]));
 });
--- a/lib/checks/canSimplifyPolynomialTerms.js
+++ b/lib/checks/canSimplifyPolynomialTerms.js
@@ -1,4 +1,5 @@
 const canAddLikeTermPolynomialNodes = require('./canAddLikeTermPolynomialNodes');
+const canFindDenominatorInNumerator = require('./canFindDenominatorInNumerator');
 const canMultiplyLikeTermPolynomialNodes = require('./canMultiplyLikeTermPolynomialNodes');
 const canRearrangeCoefficient = require('./canRearrangeCoefficient');
 
@@ -6,6 +7,7 @@ const canRearrangeCoefficient = require('./canRearrangeCoefficient');
 // polynomial terms that can be combined in some way.
 function canSimplifyPolynomialTerms(node) {
   return (canAddLikeTermPolynomialNodes(node) ||
+          canFindDenominatorInNumerator(node) ||
           canMultiplyLikeTermPolynomialNodes(node) ||
           canRearrangeCoefficient(node));
 }
--- a/lib/checks/index.js
+++ b/lib/checks/index.js
@@ -1,4 +1,5 @@
 const canAddLikeTermPolynomialNodes = require('./canAddLikeTermPolynomialNodes');
+const canFindDenominatorInNumerator = require('./canFindDenominatorInNumerator');
 const canMultiplyLikeTermPolynomialNodes = require('./canMultiplyLikeTermPolynomialNodes');
 const canRearrangeCoefficient = require('./canRearrangeCoefficient');
 const canSimplifyPolynomialTerms = require('./canSimplifyPolynomialTerms');
@@ -8,6 +9,7 @@ const resolvesToConstant = require('./resolvesToConstant');
 
 module.exports = {
   canAddLikeTermPolynomialNodes,
+  canFindDenominatorInNumerator,
   canMultiplyLikeTermPolynomialNodes,
   canRearrangeCoefficient,
   canSimplifyPolynomialTerms,


lib/simplifyExpression/breakUpNumeratorSearch/index.js
```diff
@@ -27,8 +28,33 @@ function breakUpNumerator(node) {
   // At this point, we know that node is a fraction and its numerator is a sum
   // of terms that can't be collected or combined, so we should break it up.
   const fractionList = [];
-  const denominator = node.args[1];
-  numerator.args.forEach(arg => {
+  let denominator = node.args[1];
+
+  // Check if we can add a constant to make the fraction nicer
+  // fraction e.g. (2+x)/(5+x) -> (5+x)/(5+x) + 3/(5+x)
```
aliang8: I think you mean - 3/ (5 + x) *.

lib/checks/canFindDenominatorInNumerator.js
```diff
@@ -0,0 +1,35 @@
...
+    numerator = numerator.content;
+  }
+  if (Node.Type.isParenthesis(denominator)) {
+    denominator = denominator.content;
+  }
+
+  if (numerator.op !== '+') {
+    return false;
+  }
+  if (!('args' in denominator)) {
```
aliang8: I'm not understanding what this part does. Can you explain?

lib/checks/canFindDenominatorInNumerator.js
```diff
@@ -0,0 +1,35 @@
...
+  let numerator = node.args[0];
+  let denominator = node.args[1];
+  if (Node.Type.isParenthesis(numerator)) {
+    numerator = numerator.content;
+  }
+  if (Node.Type.isParenthesis(denominator)) {
+    denominator = denominator.content;
+  }
+
+  if (numerator.op !== '+') {
```
aliang8: Try adding the case where the numerator has minus operation too.

lib/checks/canFindDenominatorInNumerator.js
eltonlaw: line 23 is to check whether the numerator is the '+' op between nodes, so that's to check that there's an 'x + y' kind of thing going on, but now that you mention it, it's not actually required. cause we should be able to go from x/(x+1) -> (x+1)/(x+1) -1)/(x+1). I'll take this out.

line 26 is in there because tests were failing by having a constant in the bottom, i think a constant in the bottom is resolved best the original way.

lib/simplifyExpression/breakUpNumeratorSearch/index.js
eltonlaw: Whoops, yeah haha

lib/checks/canFindDenominatorInNumerator.js
```diff
@@ -0,0 +1,35 @@
+const Node = require('../node');
+
+// Returns true if by adding a term you can simplify part of the function into an integer
```
evykassirer: I know it's almost 80 characters here so it feels annoying to put it in two lines, but... 80 character max please ^_^

lib/checks/canFindDenominatorInNumerator.js
```diff
@@ -0,0 +1,35 @@
...
+// e.g. (2x+1)/(2x+3) -> True
```
evykassirer: it would be easiest to understand this function (from an outsider perspective) if you explain why it's true as well

```
// e.g. (2x+1)/(2x+3) -> True 
// because of the simplification  (2x+1)/(2x+3) -> (2x + 3)/(2x + 3) - 2/(2x + 3) -> 1 - 2/(2x + 3) 
```

lib/checks/canFindDenominatorInNumerator.js
evykassirer: > actually I think with this we probably need to check that there's some sort of subtraction or addition somewhere in the fraction, so maybe set it to try to catch any of those scenarios. What do you think?

I think the best way to check this is to make sure that the numerator and denominator aren't equal. if they're both `x` then they're equal and there's nothing to do. but also if they're both `x+1` then this step doesn't apply, we can just cancel (that step's code is elsewhere). How's that sound?

lib/checks/canFindDenominatorInNumerator.js
```diff
@@ -0,0 +1,35 @@
...
+    denominator = denominator.content;
+  }
+
+  if (numerator.op !== '+') {
+    return false;
+  }
+  if (!('args' in denominator)) {
+    return false;
+  }
+  const numeratorFirstTerm = new Node.PolynomialTerm(numerator.args[0]);
```
evykassirer: most of the time I think it'll be sorted so that the first term is the poly term, but I don't think this is guaranteed. you could have `(1+x) / (x+2)`. So we might need to have a step somewhere that sorts, or otherwise iterate through the arguments to find this poly term

Also, where do you make sure they have the same exponent? (i.e. for now that it's always 1)

lib/simplifyExpression/breakUpNumeratorSearch/index.js
evykassirer: so then you'd also want to say add/subtract ^_^

lib/simplifyExpression/breakUpNumeratorSearch/index.js
```diff
@@ -27,8 +28,33 @@ function breakUpNumerator(node) {
...
+  if (canFindDenominatorInNumerator(node)) {
+    let denominatorParenRemoved = false;
+    if (Node.Type.isParenthesis(denominator)) {
+      denominatorParenRemoved = true;
+      denominator = denominator.content;
+    }
+    const newNumerator = [];
+
+    // The constant value difference between the numerator and the denominator
+    const numeratorConstant = parseInt(numerator.args[1].value);
```
evykassirer: I don't think you ever make sure that this is actually a constant - do a check either here or in `canFindDenominatorInNumerator` or else you'll get an error here

lib/simplifyExpression/breakUpNumeratorSearch/index.js
```diff
@@ -27,8 +28,33 @@ function breakUpNumerator(node) {
...
+
+    newNumerator.push(Node.Creator.constant(addedConstant));
+    newNumerator.push(denominator);
+
+    numerator = newNumerator;
+
+    if (denominatorParenRemoved) {
+      denominator = Node.Creator.parenthesis(denominator);
+    }
+  }
```
evykassirer: I think it'd be best if we can break this up into two steps. the step you added, and the break up fraction that comes after (and then maybe we can group in the cancelling that comes after). and then we'd have substeps for breaking up the fraction.

what do you think? (let me know if you want me to explain more what this means)

test/checks/checks.test.js
```diff
@@ -29,3 +29,10 @@ describe('canSimplifyPolynomialTerms addition', function() {
   ];
   tests.forEach(t => testCanCombine(t[0], t[1]));
 });
+
+describe('canSimplifyPolynomialTerms denominator in numerator', function() {
+  const tests = [
+    ['(2x + 3)/(2x + 2)', true],
```
evykassirer: can you add a bunch more tests? especially a cases that should be false

you can use some of my comments above for edge cases to test for too :)

test/simplifyExpression/breakUpNumeratorSearch/breakUpNumeratorSearch.test.js
```diff
@@ -11,6 +11,7 @@ describe('breakUpNumerator', function() {
     ['(x+3+y)/3', '(x / 3 + 3/3 + y / 3)'],
     ['(2+x)/4', '(2/4 + x / 4)'],
     ['2(x+3)/3', '2 * (x / 3 + 3/3)'],
+    ['(2x + 3)/(2x + 2)', '(1 / (2x + 2) + (2x + 2) / (2x + 2))'],
```
evykassirer: same here - try to think of edge cases. I can help you brainstorm if you dunno what to add

aliang8:
Hi @eltonlaw and @evykassirer,
I found a really cool approach to long polynomial division that I would like to share with the two of you. It's super intuitive, pedagogical, just like how we would normally do long division. 

The pseudocode is here: https://rosettacode.org/wiki/Polynomial_long_division#Python

Basically, it represents the polynomials as vectors such that the i^th element keeps the coefficient of x^i
e.g [-42 0 -12 1] => x^3 - 12 x^2 + 0x - 42
and multiplication by a monomial is a shift of the vector's element to the right
e.g [-42 0 -12 1] * x^2 => [0 0 -42 0 -12 1]
It uses a simple recursive loop to find all the coefficients of the quotient polynomial by applying the two concepts above.
Highly suggest taking a look at it since it covers most if not all of the edge cases. Let me know what yall think of it 👍

lib/checks/canFindDenominatorInNumerator.js
```diff
@@ -19,17 +20,42 @@ function canFindDenominatorInNumerator(node) {
...
+        denominator.op === '+' || numerator.op === '-')) {
+    return false;
+  }
+  if (denominator.op !== '+') {
+    return false;
+  }
+
+  let numeratorFirstTerm;
+  if (numerator.op === '+') {
+    numeratorFirstTerm = new Node.PolynomialTerm(numerator.args[0]);
```
evykassirer: I'm concerned that we can just assume it'll be a polynomial term

(what if the second arg is a polynomial term? what if there are more than 2 args?)

Sorry this got more complicated haha. If you want, we can chat on gitter or something and figure out how to attack this

lib/checks/canFindDenominatorInNumerator.js
```diff
@@ -19,17 +20,42 @@ function canFindDenominatorInNumerator(node) {
...
-  if (numerator.op !== '+') {
+  if (!(numeratorFirstTerm)) {
     return false;
   }
-  if (!('args' in denominator)) {
+  if (!(denominatorFirstTerm)) {
+    return false;
+  }
+
+  if (!(numeratorFirstTerm.getSymbolName() === 'x' && denominatorFirstTerm.getSymbolName() === 'x')) {
```
evykassirer: I think it's better to just say they have the same symbol. Then if they share the symbol `y`, it'd work too!

lib/checks/canFindDenominatorInNumerator.js
```diff
@@ -19,17 +20,42 @@ function canFindDenominatorInNumerator(node) {
   if (Node.Type.isParenthesis(denominator)) {
     denominator = denominator.content;
   }
+  if (!(numerator.op === '+' || numerator.op === '-' ||
+        denominator.op === '+' || numerator.op === '-')) {
```
eltonlaw: whoops that should be denominator.op

lib/checks/canFindDenominatorInNumerator.js
```diff
@@ -19,17 +20,42 @@ function canFindDenominatorInNumerator(node) {
...
+    return false;
+  }
+  if (denominator.op !== '+') {
```
eltonlaw: I was thinking we don't need to use this extended version on it. The original breakUpNumerator function works fine for this case. But I think you've highlighted an interesting point, anyone reading the code would be confused.

lib/checks/canFindDenominatorInNumerator.js
```diff
@@ -19,17 +20,42 @@ function canFindDenominatorInNumerator(node) {
   if (Node.Type.isParenthesis(denominator)) {
     denominator = denominator.content;
   }
+  if (!(numerator.op === '+' || numerator.op === '-' ||
```
eltonlaw: Okay

lib/checks/canFindDenominatorInNumerator.js
eltonlaw: Yeah, this is what I wanted to talk about actually. One of the main issues is that it's not sorted, so I wanted to ask you all if we should add a sorting function which will run before this. Or if you don't see any merit in that, we could add some logic to find the highest polynomial term right into the function (or outside of it if you want).

lib/checks/canFindDenominatorInNumerator.js
eltonlaw: oh true, I didn't think about the symbol being a different name. The reason I did it to equal 'x' was to restrict it to the case of x^1

test/simplifyExpression/breakUpNumeratorSearch/breakUpNumeratorSearch.test.js
```diff
@@ -11,7 +11,13 @@ describe('breakUpNumerator', function() {
     ['(x+3+y)/3', '(x / 3 + 3/3 + y / 3)'],
     ['(2+x)/4', '(2/4 + x / 4)'],
     ['2(x+3)/3', '2 * (x / 3 + 3/3)'],
-    ['(2x + 3)/(2x + 2)', '(1 / (2x + 2) + (2x + 2) / (2x + 2))'],
+    ['(2x + 3)/(2x + 2)', '((2x + 2) / (2x + 2) + 1 / (2x + 2))'],
+    ['(2x - 3)/(2x + 2)', '((2x + 2) / (2x + 2) - 5 / (2x + 2))'],
+    ['(2x + 3)/(2x)', '(2x / (2x) + 3 / (2x))'],
+    ['(3 + 2x)/(2x)', '(3 / (2x) + 2x / (2x))'],
+    ['(4x + 3)/(2x + 2)', '(2 * (2x + 2) / (2x + 2) - 1 / (2x + 2))'],
+//    ['(2x)/(3 + 2x)', '((3 + 2x) / (3 + 2x) - 3 / (3 + 2x))'],
```
eltonlaw: As mentioned earlier, I didn't know what to do without some sort of way to sort it.

lib/checks/canFindDenominatorInNumerator.js
eltonlaw: yeah sure, I'm down to chat on gitter

test/simplifyExpression/breakUpNumeratorSearch/breakUpNumeratorSearch.test.js
evykassirer: ah, cool - sorry for missing that

lib/checks/canFindDenominatorInNumerator.js
evykassirer: yeah I think just doing getSymbolName()  === getSymbolName()  would work!

lib/checks/canFindDenominatorInNumerator.js
evykassirer: ooo true. Maybe just add a comment saying where the other case is handled then!

lib/checks/canFindDenominatorInNumerator.js
evykassirer: sweet - okay I actually am busy all day tomorrow >.> (first day of work!!) 

does Wed evening work for you?

evykassirer: or can we chat asynchronously - I can try to think about solutions to this tomorrow

lib/checks/canFindDenominatorInNumerator.js
eltonlaw: wait I'm not really getting the use of getSymbolName() actually, what other values can the symbol even take, it seems like it always returns 'x'. Is this just for when there are multiple variables? Cause that seems way out of the scope for now anyways. Would it be better to check that denominatorFirstTerm.getExponentNode() === numeratorFirstTerm.getExponentNode() or check if they equal undefined?

lib/checks/canFindDenominatorInNumerator.js
eltonlaw: uhh yeah lets just chat asynchronously then, no worries take your time, good luck!

lib/checks/canFindDenominatorInNumerator.js
evykassirer: using `getSymbolName` makes sure the symbol in numerator and denominator are the same - i.e. both `x` or both `y` or whatever - you could take any example with `x`s and change them to `y`s and it should also work (actually can you add a test for that? ^_^)

`numeratorFirstTerm.getSymbolName() === denominatorFirstTerm.getSymbolName()` is what I was thinking

lib/checks/canFindDenominatorInNumerator.js
evykassirer: I think you could start with doing this rule _only_ if there are exactly one or two args, the first is always a polynomial term, and the second if it exists is always a constant

i.e. come up with very limited cases and then test for them before going forward

and then after merging this you could add some more cases if they're easy enough

what do you think?

eltonlaw: oh whoops didn't see this, sure sounds good!

eltonlaw:
I think I incorporated all your feedback @evykassirer, as things are now is this okay to get merged?

lib/checks/canFindDenominatorInNumerator.js
```diff
@@ -20,38 +20,69 @@ function canFindDenominatorInNumerator(node) {
...
     return false;
   }
-  if (denominator.op !== '+') {
+  if (!(denominator.op === '+' || denominator.op === '-')) {
     return false;
   }
-
+  // Check if numerator's second argument is a constant if numerator has two arguments
+  if (n_args_length === 2) {
+    if (!(Node.Type.isConstant(numerator.args[1]))) {
```
evykassirer: you don't need the parens around Node.Type(.....)

lib/checks/canFindDenominatorInNumerator.js
```diff
@@ -20,38 +20,69 @@ function canFindDenominatorInNumerator(node) {
...
-  }
-  if (!(denominatorFirstTerm)) {
+  // If an exponent exists (aka not x^1), return false
+  if (numeratorFirstTerm.getExponentNode() ||
+      denominatorFirstTerm.getExponentNode()) {
     return false;
   }
-
-  if (!(numeratorFirstTerm.getSymbolName() === 'x' && denominatorFirstTerm.getSymbolName() === 'x')) {
+  // Check that the symbols are the same, Ex. (x+1)/(y+1) would not pass
```
evykassirer: I love the example in the comment :D

lib/checks/canFindDenominatorInNumerator.js
```diff
@@ -20,38 +20,69 @@ function canFindDenominatorInNumerator(node) {
...
+  if ('args' in denominator) {
+    d_args_length = denominator.args.length;
+  }
+  else {
+    d_args_length = 1;
+  }
+
+  // If numerator isn't length 2 or length 1 with a polynomial return false
+  if (!(n_args_length === 2)) {
+    if (!(n_args_length === 1 || Node.Type.isConstant(numerator))) {
```
evykassirer: also you should be checking for polynomial, not constant, right?

evykassirer: did you also check that the first argument of the numerator/denominator is a polynomial term?

lib/checks/canFindDenominatorInNumerator.js
eltonlaw: oh true!

lib/checks/canFindDenominatorInNumerator.js
```diff
@@ -20,38 +20,69 @@ function canFindDenominatorInNumerator(node) {
   if (Node.Type.isParenthesis(denominator)) {
     denominator = denominator.content;
   }
-  if (!(numerator.op === '+' || numerator.op === '-' ||
-        denominator.op === '+' || numerator.op === '-')) {
+
+  let n_args_length;
+  // If numerator is '*' op, it signifies a single 'ax', should be assigned a
+  // length of 1
+  if ('args' in numerator && numerator.op !== '*') {
```
eltonlaw: ah yeah good catch, it should be ===

lib/checks/canFindDenominatorInNumerator.js
```diff
@@ -20,38 +20,69 @@ function canFindDenominatorInNumerator(node) {
...
+  }
+
+  // If numerator isn't length 2 or length 1 with a polynomial return false
+  if (!(n_args_length === 2)) {
+    if (!(n_args_length === 1 || Node.Type.isConstant(numerator))) {
+      return false;
+    }
+  }
+  // Function doesn't support denominators with args > 2
+  // Tf d_args_length < 2 the normal functionality already covers it
```
eltonlaw: normal functionality is just taking each arg of the numerator and putting a denominator under it. So (ax^2 +bx +c)/d decomposes to ax^2/d +bx/d + c/d

Yeah it's the same as 1, I'll change it

lib/checks/canFindDenominatorInNumerator.js
eltonlaw: You're right, it should be a check for polynomial, originally I wanted to put isSymbol() there, but it doesn't work on values that have a coefficient: ex. Node.Type.isSymbol('2x') evaluates to false, so i figured the logical positive would work instead: !Node.Type.isConstant('2x') evaluates to true.

Also, I don't check that the first argument of the numerator/denominator is a polynomial term because I assume that if we've already evaluated through and there's a second arg in the numerator/denominator that's a constant, the only thing left in the first args MUST be a polynomial, cause the other simplifiers would have added like terms already. Do you want me to explicitly write !Node.Type.isConstant(denominator.arg[0])?

lib/checks/canFindDenominatorInNumerator.js
evykassirer: can you give that example in the comment just so people know what you mean?

lib/checks/canFindDenominatorInNumerator.js
evykassirer: great - make sure the comment explains why (not just what - so not just saying what the if statement is looking for, but an example like what you showed me so people know why you're checking that)

lib/checks/canFindDenominatorInNumerator.js
evykassirer: there's a `isPolynomialTerm` (or something similar) defined somewhere - I think that's what you want. You could actually probably use it above when your'e trying to avoid the ax case (just make sure it's not a poly term)

`!Node.Type.isConstant('2x')` isn't rigorous enough because there are a bunch (5?) different types of node types - you can look in `Node/Type` to see them all

a counterexample to your assumption - what if it was `5/(2^x + 3)`?

evykassirer:
I'm at GoogleIO this week and can't take much time to review your code - but I replied to some comments and will actually look at the diff on the weekend

thanks for your patience! soon these rules will be easier to write hopefully - someone's working on a cleaner way to describe them right now :) (if you want to see more, check out https://github.com/semantic-math/math-rules)

lib/checks/canFindDenominatorInNumerator.js
```diff
@@ -5,8 +5,21 @@ const Node = require('../node');
 // e.g. (2x+1)/(2x+3) -> True because of the following simplification
 // (2x+1)/(2x+3) -> (2x + 3)/(2x + 3) - 2/(2x + 3) -> 1 - 2/(2x + 3)
 // e.g. (2x+1)/(2x^2 + 3) -> False
+// =========================================================================
+// CHECKS
```
evykassirer: I think it would be most clear if you explain what exactly it handles

maybe:

"conditions include being a fraction where the numerator is a sum of ...."

so you see you can include if it's a + node and what the first and second argument is etc into a much more concise language, and that's what I was hoping for.

"Check that the number of arguments in parent node is 2" is already implied in it being a fraction

e.g. several of your conditions can be combined into "the denominator has to be an addition/subtraction of a polynomial term and a constant term"

evykassirer: I love that you ahve this summary here though - it'll be so helpful!

lib/checks/canFindDenominatorInNumerator.js
```diff
@@ -20,41 +33,70 @@ function canFindDenominatorInNumerator(node) {
   if (Node.Type.isParenthesis(denominator)) {
     denominator = denominator.content;
   }
-  if (!(numerator.op === '+' || numerator.op === '-' ||
-        denominator.op === '+' || numerator.op === '-')) {
+
+  let numeratorArgsLength;
+  // If numerator is '*' op, it signifies a single 'ax', should be assigned a
+  // length of 2
```
evykassirer: I think a clearer way to do this check is to see if it's a polynomial term (and then length 1) or else make it `numerator.args.length`

also if numerator doesn't have args, you should return false, which you don't do now right?

lib/checks/canFindDenominatorInNumerator.js
eltonlaw: Yeah, you're right. Oh wow, I didn't see isPolynomialTerm. Shoot, sorry for missing that. On a kind of unrelated note, to me it would make more sense if isPolynomialTerm was in the Type file instead, what do you think?

lib/checks/canFindDenominatorInNumerator.js
eltonlaw: okay i also added another two lines of explanation to try to make it a bit more clear, let me know if you think it adds to the confusion or helps.

lib/checks/canFindDenominatorInNumerator.js
evykassirer: yeah that would be good to add there if it's easy! probably not instead, since it's nice to have the logic within polynomial term code, but if you could reference it in `Type` for now that'd be awesome

eltonlaw: uhh, actually maybe not. If we require 'PolynomialTerm' from 'Type', then they're referencing each other. So the only way to put it in there is if we just have two copies. So yeah maybe screw it.

evykassirer: ahh yeah okay I guess that's why that wasn't there then - ah well

test/simplifyExpression/breakUpNumeratorSearch/breakUpNumeratorSearch.test.js
```diff
@@ -11,6 +11,13 @@ describe('breakUpNumerator', function() {
     ['(x+3+y)/3', '(x / 3 + 3/3 + y / 3)'],
     ['(2+x)/4', '(2/4 + x / 4)'],
     ['2(x+3)/3', '2 * (x / 3 + 3/3)'],
+    ['(2x + 3)/(2x + 2)', '((2x + 2) / (2x + 2) + 1 / (2x + 2))'],
+    ['(2x - 3)/(2x + 2)', '((2x + 2) / (2x + 2) - 5 / (2x + 2))'],
+    ['(2x + 3)/(2x)', '(2x / (2x) + 3 / (2x))'],
+    ['(3 + 2x)/(2x)', '(3 / (2x) + 2x / (2x))'],
+    ['(4x + 3)/(2x + 2)', '(2 * (2x + 2) / (2x + 2) - 1 / (2x + 2))'],
+//    ['(2x)/(3 + 2x)', '((3 + 2x) / (3 + 2x) - 3 / (3 + 2x))'],
+//    ['(2x)/(2x + 3)', '((2x + 3) / (2x + 3)) - 3 / (2x + 3)'],
```
evykassirer: if these tests pass, uncomment them

otherwise leave a TODO with what we'll need to do in the future to make them pass

test/checks/checks.test.js
```diff
@@ -29,3 +29,26 @@ describe('canSimplifyPolynomialTerms addition', function() {
...
+  const tests = [
+    ['(x+1)/(x-2)', true],
+    ['(2x)/(x+4)', true],
+    ['(x)/(x+4)', true],
+    ['(x)/(2x+4)', true],
+    ['(x+3)/(x)', false], // Normal breakup function already solves this
+    ['(2x + 3)/(2x + 2)', true],
+    ['(2x+3)/(2x)', false], // Normal breakup function already solves this
+    ['(2x)/(2x + 2)', true],
+    ['(5x + 3)/(4)', false], // Normal breakup function already solves this
```
evykassirer: these comments are super helpful!

eltonlaw:
Hey @aliang8 and @evykassirer, I think the problem with 2x / (2x +3) is the following line in breakUpNumeratorSearch/index.js:  
```
if (!Node.Type.isOperator(numerator) || numerator.op !== '+') {
}
```
the second part numerator.op !== '+' evaluates to true so the no change node is returned. I was thinking about adding a part before this where it adds an op to  add a constant node '0' to the numerator in the case of a single polynomial numerator. But idk, it might end up really fragile. What do you think we should do?

aliang8:
Ah good call, I see where you're referring to. Maybe you could add in a check to see if the numerator is a single polynomial term or if it is just a constant and handle it accordingly inside that if statement. By doing this, we can restrict to only the cases that we handle.

So in practice that if statement could be: `if ((!isOperator(node) || (numerator.op != '+')) && notSinglePolynomialTerm(node)) {return}` It seems a bit clunky as of now, but hopefully you could think of a better way to handle this.

evykassirer:
What about if you check your own conditions first, then check 
`if (!Node.Type.isOperator(numerator) || numerator.op !== '+') {`
if your conditions don't pass?

eltonlaw:
hey @evykassirer @aliang8, I've been trying to debug this but I can't seem to identify the problem. I've been poking around but I just can't seem to make sense of what's going on. When I try to incorporate the new checks everything seems to blow up. Sometimes tests from addConstantFractions give out errors, other times a bunch of tests in solveEquation fail. I am baffled. Do you guys want to check it out? 

In the meantime, I just commented it out and left a TODO.

Issue #146, opened 2017-02-26
run test code inside of `it` block


kevinbarabash:
mocha runs all `describe` blocks and queues up all closures passed to `it` to be run later.  Currently, we're running test code outside of the `it` block so it runs before the `assert` which gets run at a different time.  This makes debugging tests really hard.  See https://github.com/socraticorg/mathsteps/blob/master/test/solveEquation/solveEquation.test.js#L8-L21 for an example of what I mean.

evykassirer:
omg yes I was noticing this but wasn't sure how to fix it

what do we need to do to fix it?

in this example would it be to move `it(equationString + ' -> ' + outputStr, (done) => {` to line 9?

if so, we can just make this an "easy" tagged issue, seems like a good first change!

kevinbarabash:
It should look something like this:
```
function testSolve(equationString, outputStr, debug=false) {
  it(`${equationString} -> ${outputStr}`, () => {
    const steps = solveEquation(equationString, debug);
    const lastStep = steps.length === 0
      ? NO_STEPS
      : steps[steps.length -1].newEquation.print();
    assert.equal(lastStep, outputStr);
  });
}
```

The `done` is only necessary when there's async code inside the test.

evykassirer:
yeah someone added `done` earlier and idk much about it - so let's take it out then

sounds like a good change for someone new to the codebase then :D woo

kevinbarabash:
@evykassirer can you assign this to me?  I'm fixing it as part of the math-parser work.

evykassirer:
@kevinbarabash do you want to do this separately since you closed your PR?

kevinbarabash:
@evykassirer sure... it was a pretty easy change.  Can you assign me?

PR #159, opened 2017-04-24
fixes cancelling code

aliang8:


test/simplifyExpression/.#simplify.test.js
```diff
@@ -0,0 +1 @@
+diamond@Anthonys-MBP.home.13396
```
evykassirer: hmm what's this?

lib/simplifyExpression/multiplyFractionsSearch/index.js
```diff
@@ -1,3 +1,5 @@
+const cancelLikeTerms = require('../../simplifyExpression/fractionsSearch/cancelLikeTerms.js');
+//const removeUnnecessaryParens = require('../../util/removeUnnecessaryParens.js');
```
evykassirer: why's this commented out?

lib/simplifyExpression/multiplyFractionsSearch/index.js
```diff
@@ -23,13 +25,17 @@ function multiplyFractions(node) {
...
   const atLeastOneFraction = node.args.some(
-    arg => Node.Type.isOperator(arg, '/'));
-  const hasPolynomialTerms = node.args.some(
-    arg => Node.PolynomialTerm.isPolynomialTerm(arg));
-  if (!atLeastOneFraction || hasPolynomialTerms) {
+      arg => Node.Type.isOperator(arg, '/'));
+
+  // If term is in the form of <constant fraction> <polynomial term>
+  // do not multiply in order to keep pedagogical style
+  // e.g 1/2 * x^2 -> 1/2 x^2 and not 1x^2 / 2
```
evykassirer: this is a great comment! I'm really glad you have an example, too

do you think a more clear version of the example would be:

`e.g 1/2 * x^2 remains unchanged instead of stepping to 1*x^2 / 2`

cause this code doesn't do `1/2 * x^2 -> 1/2 x^2`

lib/simplifyExpression/multiplyFractionsSearch/index.js
```diff
@@ -23,13 +25,17 @@ function multiplyFractions(node) {
...
+  const isLinearFraction = Node.PolynomialTerm.isPolynomialTerm(node.args[1]);
+  const isConstantFraction = Node.Type.isConstantFraction(node.args[0]);
```
evykassirer: what are you assuming about `node.args[0]` and `node.args[1]`? I think that `node.args` could be any length and not just 2 items long, right? you should probably just do a check for that too.

lib/simplifyExpression/multiplyFractionsSearch/index.js
```diff
@@ -23,13 +25,17 @@ function multiplyFractions(node) {
...
-    arg => Node.Type.isOperator(arg, '/'));
-  const hasPolynomialTerms = node.args.some(
-    arg => Node.PolynomialTerm.isPolynomialTerm(arg));
-  if (!atLeastOneFraction || hasPolynomialTerms) {
+      arg => Node.Type.isOperator(arg, '/'));
+
+  // If term is in the form of <constant fraction> <polynomial term>
+  // do not multiply in order to keep pedagogical style
+  // e.g 1/2 * x^2 -> 1/2 x^2 and not 1x^2 / 2
+  const isLinearFraction = Node.PolynomialTerm.isPolynomialTerm(node.args[1]);
```
evykassirer: what do you mean by `linear`?

lib/simplifyExpression/multiplyFractionsSearch/index.js
evykassirer: also what if you had `x^2 * 1/2` ? I feel like you wouldn't want to multiply it out there either

lib/simplifyExpression/multiplyFractionsSearch/index.js
```diff
@@ -43,14 +49,20 @@ function multiplyFractions(node) {
...
+  );
+
   const newDenominator = denominatorArgs.length === 1
     ? denominatorArgs[0]
     : Node.Creator.parenthesis(Node.Creator.operator('*', denominatorArgs));
 
   const newNode = Node.Creator.operator('/', [newNumerator, newDenominator]);
+
+  // Apply cancellation
+  const simplifiedNode = cancelLikeTerms(newNode)['newNode'];
```
evykassirer: convention in the codebase is `.newNode` instead of  `['newNode']`

lib/simplifyExpression/multiplyFractionsSearch/index.js
```diff
@@ -43,14 +49,20 @@ function multiplyFractions(node) {
...
+
   return Node.Status.nodeChanged(
-    ChangeTypes.MULTIPLY_FRACTIONS, node, newNode);
+    ChangeTypes.MULTIPLY_FRACTIONS, node, simplifiedNode);
```
evykassirer: there's an issue here - now even if the main change is cancelling, it'll say that the changeType is multiplying fractions. pedagogically there should always be a Node.Status with `ChangeTypes.MULTIPLY_FRACTIONS` when fractions are multiplied without cancelling, and we probably want a Node.Status with just the cancelling changeType and no multiplying fractions one if *do* cancel - does that seem right?

also I think the changegroups will be messed up here - double check that the right nodes (the ones being cancelled) are still highlighted in the `oldNode` 

this can be fixed (I think?) by just returning the status from `cancelLikeTerms(newNode)` if a change happens there

test/simplifyExpression/fractionsSearch/cancelLikeTerms.test.js
```diff
@@ -2,8 +2,8 @@ const cancelLikeTerms = require('../../../lib/simplifyExpression/fractionsSearch
 
 const TestUtil = require('../../TestUtil');
 
-function testCancelLikeTerms(exprStr, expectedStr) {
-  TestUtil.testSimplification(cancelLikeTerms, exprStr, expectedStr);
+function testCancelLikeTerms(exprString, outputStr) {
```
evykassirer: maybe use both - `expectedOutputStr`? cause it might not be the output, it's just what we hope will be the output

test/simplifyExpression/fractionsSearch/cancelLikeTerms.test.js
```diff
@@ -21,13 +21,13 @@ describe('cancel like terms', function () {
     ['(2x+5)^8 / (2x+5)^2', '(2x + 5)^(8 - (2))'],
     ['(4x^3) / (5x^2)', '(4x^(3 - (2))) / (5)'],
     ['-x / -x', '1'],
+    ['(6x)/ x', '(6)'],
+    ['(2 * x^2 * 2x) / (y * y)', '(2x^2 * 2x) / (y * y)']
```
evykassirer: maybe add a comment at the end of this test saying `// no cancelling`

test/simplifyExpression/fractionsSearch/cancelLikeTerms.test.js
```diff
@@ -21,13 +21,13 @@ describe('cancel like terms', function () {
...
   ];
 
   tests.forEach(t => {
     const before = t[0];
     const after = t[1];
-    it(before + ' -> ' + after, function () {
```
evykassirer: why'd you get rid of the `it`? the text it prints is helpful

maybe you noticed the problem in #146 ? maybe you could take that off of kevin's hands so he can focus on the parser (in a separate PR though, keep the `it` the way it is here)

test/simplifyExpression/multiplyFractionsSearch/multiplyFractionsSearch.test.js
```diff
@@ -8,9 +8,13 @@ function testMultiplyFractionsSearch(exprString, outputStr) {
 
 describe('multiplyFractions', function () {
```
evykassirer: hm it's sort of odd now that multiplyFractions also cancels fractions. you should probably rename multiplyFractions

... or you could call multiply fractions *in* the cancelling code so cancelling always happens in `cancelLikeTerms` instead of sometimes happening in `multiplyFractions` - what do you think of that? it seems a bit more intuitive to me

test/simplifyExpression/simplify.test.js
```diff
@@ -136,7 +136,7 @@ describe('absolute value support', function() {
     ['abs(5/18-abs(9/-18))', '2/9'],
     // handle parens around abs()
     ['( abs( -3) )/(3)', '1'],
-    ['- abs( -40)', '-40'],
+      ['- abs( -40)', '-40'],
```
evykassirer: you added some extra space here

test/simplifyExpression/simplify.test.js
```diff
@@ -158,7 +158,7 @@ describe('nthRoot support', function() {
 describe('handles unnecessary parens at root level', function() {
   const tests = [
     ['(x+(y))', 'x + y'],
-    ['((x+y) + ((z^3)))', 'x + y + z^3'],
+      ['((x+y) + ((z^3)))', 'x + y + z^3'],
```
evykassirer: and here

evykassirer:
oh and tests look sweet - thanks for adding so many!

lib/simplifyExpression/multiplyFractionsSearch/index.js
aliang8: never mind this, it's just to remove the parentheses after multiplication. but i don't think we need to. i will remove this line :D

test/simplifyExpression/multiplyFractionsSearch/multiplyFractionsSearch.test.js
aliang8: I disagree, I feel like it would be weird to call multiply in cancel code. It seems more intuitive to multiply and then cancel at the end. But I will rename the file :)

evykassirer: but then there's cancelling code and also... other cancelling code? do you think we could just merge them or something?

aliang8: Hmm, maybe we could merge them. I don't see any other use case for cancelling code besides multiplying fractions.

lib/simplifyExpression/multiplyFractionsSearch/index.js
aliang8: Thanks for pointing this out! Will fix.

lib/simplifyExpression/multiplyFractionsSearch/index.js
aliang8: fixed

test/simplifyExpression/simplify.test.js
aliang8: fixed

test/simplifyExpression/simplify.test.js
aliang8: fixed

test/simplifyExpression/multiplyFractionsSearch/multiplyFractionsSearch.test.js
evykassirer: well, sometimes you can cancel without having to multiply first

funny, I see it more like the only case for multiplying fractions is to cancel :p especially since that's the step that users see

test/simplifyExpression/fractionsSearch/cancelLikeTerms.test.js
aliang8: fixed

test/simplifyExpression/fractionsSearch/cancelLikeTerms.test.js
aliang8: added

evykassirer:
did you wanna keep this open and merge eventually @aliang8 or close it?

aliang8:
I wouldn't mind closing it. Actually I think it'll be better.

PR #161, opened 2017-04-28
Resetting change groups when simplifying a fraction of a polynomial

aelnaiem:
This was causing there to be two change groups in the following example
10x/5 -> 2x

The two would be a change group and the 2x would also be a change group. This is a simple fix for this, and another reason for us to look at change-groups again in the future.


evykassirer:
does this fix make 2 the changegroup or 2x the changegroup?

I feel like we shouldn't reset changegroups but should instead not have it add new changegroups (you can pass something into `Node.Status` to tell it you did the changegroups yourself)

evykassirer:
Ah, interesting. that polyterm constructor is a pain for changegroups. Seems good to me - once @aliang8 looks it over I'm happy to merge

thanks for doing this Ahmed :)

PR #167, opened 2017-06-18
Start of porting mathsteps to use math-rules

kevinbarabash:
This PR disables almost all of the existing tests except for the ones that are currently passing after being ported.  Note that the tests themselves have not been changed.

The plan build upon this foundation with a series of small PRs to port each of the various search commands and then eventually update the simplifyExpression and solveEquation.

### DIFF
--- a/lib/checks/hasUnsupportedNodes.js
+++ b/lib/checks/hasUnsupportedNodes.js
@@ -1,30 +1,26 @@
-const Node = require('../node');
+const {query} = require('math-nodes');
+
 const resolvesToConstant = require('./resolvesToConstant');
 
 function hasUnsupportedNodes(node) {
-  if (Node.Type.isParenthesis(node)) {
-    return hasUnsupportedNodes(node.content);
-  }
-  else if (Node.Type.isUnaryMinus(node)) {
-    return hasUnsupportedNodes(node.args[0]);
-  }
-  else if (Node.Type.isOperator(node)) {
+  if (query.isApply(node)) {
+    if (query.isAbs(node)) {
+      if (node.args.length !== 1) {
+        return true;
+      }
+      if (!resolvesToConstant(node.args[0])) {
+        return true;
+      }
+    }
+    else if (query.isNthRoot(node)) {
+      if (node.args.length < 1 || node.args.length > 2) {
+        return true;
+      }
+    }
     return node.args.some(hasUnsupportedNodes);
   }
-  else if (Node.Type.isSymbol(node) || Node.Type.isConstant(node)) {
+  else if (query.isIdentifier(node) || query.isNumber(node)) {
     return false;
-  }
-  else if (Node.Type.isFunction(node, 'abs')) {
-    if (node.args.length !== 1) {
-      return true;
-    }
-    if (node.args.some(hasUnsupportedNodes)) {
-      return true;
-    }
-    return !resolvesToConstant(node.args[0]);
-  }
-  else if (Node.Type.isFunction(node, 'nthRoot')) {
-    return node.args.some(hasUnsupportedNodes) || node.args.length < 1;
   }
   else {
     return true;
--- a/lib/checks/hasUnsupportedNodes.js
+++ b/lib/checks/hasUnsupportedNodes.js
@@ -8,6 +8,9 @@ function hasUnsupportedNodes(node) {
       if (node.args.length !== 1) {
         return true;
       }
+      if (hasUnsupportedNodes(node.args[0])) {
+        return true;
+      }
       if (!resolvesToConstant(node.args[0])) {
         return true;
       }
--- a/lib/checks/resolvesToConstant.js
+++ b/lib/checks/resolvesToConstant.js
@@ -1,24 +1,21 @@
-const Node = require('../node');
+const {query} = require('math-nodes');
 
 // Returns true if the node is a constant or can eventually be resolved to
 // a constant.
 // e.g. 2, 2+4, (2+4)^2 would all return true. x + 4 would return false
 function resolvesToConstant(node) {
-  if (Node.Type.isOperator(node) || Node.Type.isFunction(node)) {
+  if (query.isApply(node)) {
     return node.args.every(
       (child) => resolvesToConstant(child));
   }
-  else if (Node.Type.isParenthesis(node)) {
-    return resolvesToConstant(node.content);
-  }
-  else if (Node.Type.isConstant(node, true)) {
+  else if (query.isNumber(node)) {
     return true;
   }
-  else if (Node.Type.isSymbol(node)) {
+  else if (query.isIdentifier(node)) {
     return false;
   }
-  else if (Node.Type.isUnaryMinus(node)) {
-    return resolvesToConstant(node.args[0]);
+  else if (node.type === 'Parentheses') {
+    return resolvesToConstant(node.body);
   }
   else {
     throw Error('Unsupported node type: ' + node.type);
--- a/lib/simplifyExpression/basicsSearch/reduceMultiplicationByZero.js
+++ b/lib/simplifyExpression/basicsSearch/reduceMultiplicationByZero.js
@@ -1,27 +1,16 @@
+const {build, query} = require('math-nodes');
+
 const ChangeTypes = require('../../ChangeTypes');
 const Node = require('../../node');
 
 // If `node` is a multiplication node with 0 as one of its operands,
 // reduce the node to 0. Returns a Node.Status object.
 function reduceMultiplicationByZero(node) {
-  if (node.op !== '*') {
-    return Node.Status.noChange(node);
-  }
-  const zeroIndex = node.args.findIndex(arg => {
-    if (Node.Type.isConstant(arg) && arg.value === '0') {
-      return true;
-    }
-    if (Node.PolynomialTerm.isPolynomialTerm(arg)) {
-      const polyTerm = new Node.PolynomialTerm(arg);
-      return polyTerm.getCoeffValue() === 0;
-    }
-    return false;
-  });
-  if (zeroIndex >= 0) {
-    // reduce to just the 0 node
-    const newNode = Node.Creator.constant(0);
+  // The math-rules rule for this didn't do the right thing when the '0' was
+  // in the middle of the args list.
+  if (query.isMul(node) && node.args.some(node => query.getValue(node) === 0)) {
     return Node.Status.nodeChanged(
-      ChangeTypes.MULTIPLY_BY_ZERO, node, newNode);
+      ChangeTypes.MULTIPLY_BY_ZERO, node, build.number('0'));
   }
   else {
     return Node.Status.noChange(node);
--- a/lib/simplifyExpression/basicsSearch/reduceMultiplicationByZero.js
+++ b/lib/simplifyExpression/basicsSearch/reduceMultiplicationByZero.js
@@ -1,16 +1,23 @@
-const {build, query} = require('math-nodes');
+const {query: q} = require('math-nodes');
+const {rewriteNode} = require('math-rules');
 
 const ChangeTypes = require('../../ChangeTypes');
+const defineRuleString = require('../../util/defineRuleString');
 const Node = require('../../node');
+
+// The math-rules rule for this didn't do the right thing when the '0' was
+// in the middle of the args list.
+const REDUCE_MULTIPLY_BY_ZERO = defineRuleString('#a', '0', {
+  a: node => q.isMul(node) && node.args.some(arg => q.getValue(arg) === 0),
+});
 
 // If `node` is a multiplication node with 0 as one of its operands,
 // reduce the node to 0. Returns a Node.Status object.
 function reduceMultiplicationByZero(node) {
-  // The math-rules rule for this didn't do the right thing when the '0' was
-  // in the middle of the args list.
-  if (query.isMul(node) && node.args.some(node => query.getValue(node) === 0)) {
+  const newNode = rewriteNode(REDUCE_MULTIPLY_BY_ZERO, node);
+  if (newNode) {
     return Node.Status.nodeChanged(
-      ChangeTypes.MULTIPLY_BY_ZERO, node, build.number('0'));
+      ChangeTypes.MULTIPLY_BY_ZERO, node, newNode);
   }
   else {
     return Node.Status.noChange(node);
--- a/lib/simplifyExpression/basicsSearch/removeAdditionOfZero.js
+++ b/lib/simplifyExpression/basicsSearch/removeAdditionOfZero.js
@@ -1,30 +1,25 @@
-const clone = require('../../util/clone');
-
 const ChangeTypes = require('../../ChangeTypes');
 const Node = require('../../node');
+
+const {rules, canApplyRule, applyRule} = require('math-rules');
 
 // If `node` is an addition node with 0 as one of its operands,
 // remove 0 from the operands list. Returns a Node.Status object.
 function removeAdditionOfZero(node) {
-  if (node.op !== '+') {
-    return Node.Status.noChange(node);
-  }
-  const zeroIndex = node.args.findIndex(arg => {
-    return Node.Type.isConstant(arg) && arg.value === '0';
-  });
-  let newNode = clone(node);
-  if (zeroIndex >= 0) {
-    // remove the 0 node
-    newNode.args.splice(zeroIndex, 1);
-    // if there's only one operand left, there's nothing left to add it to,
-    // so move it up the tree
-    if (newNode.args.length === 1) {
-      newNode = newNode.args[0];
-    }
+  // TODO(kevinb): merge these two rules together
+  if (canApplyRule(rules.REMOVE_ADDING_ZERO, node)) {
+    const newNode = applyRule(rules.REMOVE_ADDING_ZERO, node);
     return Node.Status.nodeChanged(
       ChangeTypes.REMOVE_ADDING_ZERO, node, newNode);
   }
-  return Node.Status.noChange(node);
+  else if (canApplyRule(rules.REMOVE_ADDING_ZERO_REVERSE, node)) {
+    const newNode = applyRule(rules.REMOVE_ADDING_ZERO_REVERSE, node);
+    return Node.Status.nodeChanged(
+      ChangeTypes.REMOVE_ADDING_ZERO, node, newNode);
+  }
+  else {
+    return Node.Status.noChange(node);
+  }
 }
 
 module.exports = removeAdditionOfZero;
--- a/lib/simplifyExpression/basicsSearch/removeAdditionOfZero.js
+++ b/lib/simplifyExpression/basicsSearch/removeAdditionOfZero.js
@@ -1,4 +1,4 @@
-const {canApplyRule, applyRule} = require('math-rules');
+const {rewriteNode} = require('math-rules');
 
 const ChangeTypes = require('../../ChangeTypes');
 const defineRuleString = require('../../util/defineRuleString');
@@ -13,14 +13,12 @@ const REMOVE_ADDING_ZERO_REVERSE = defineRuleString('0 + #a', '#a');
 // If `node` is an addition node with 0 as one of its operands,
 // remove 0 from the operands list. Returns a Node.Status object.
 function removeAdditionOfZero(node) {
-  // TODO(kevinb): merge these two rules together
-  if (canApplyRule(REMOVE_ADDING_ZERO, node)) {
-    const newNode = applyRule(REMOVE_ADDING_ZERO, node);
-    return Node.Status.nodeChanged(
-      ChangeTypes.REMOVE_ADDING_ZERO, node, newNode);
+  let newNode = null;
+  newNode = rewriteNode(REMOVE_ADDING_ZERO, node);
+  if (!newNode) {
+    newNode = rewriteNode(REMOVE_ADDING_ZERO_REVERSE, node);
   }
-  else if (canApplyRule(REMOVE_ADDING_ZERO_REVERSE, node)) {
-    const newNode = applyRule(REMOVE_ADDING_ZERO_REVERSE, node);
+  if (newNode) {
     return Node.Status.nodeChanged(
       ChangeTypes.REMOVE_ADDING_ZERO, node, newNode);
   }
--- a/lib/simplifyExpression/basicsSearch/removeAdditionOfZero.js
+++ b/lib/simplifyExpression/basicsSearch/removeAdditionOfZero.js
@@ -1,19 +1,26 @@
+const {canApplyRule, applyRule} = require('math-rules');
+
 const ChangeTypes = require('../../ChangeTypes');
+const defineRuleString = require('../../util/defineRuleString');
 const Node = require('../../node');
 
-const {rules, canApplyRule, applyRule} = require('math-rules');
+// e.g. 2 + 0 -> 2
+const REMOVE_ADDING_ZERO = defineRuleString('#a + 0', '#a');
+
+// e.g. 0 + 2 -> 2
+const REMOVE_ADDING_ZERO_REVERSE = defineRuleString('0 + #a', '#a');
 
 // If `node` is an addition node with 0 as one of its operands,
 // remove 0 from the operands list. Returns a Node.Status object.
 function removeAdditionOfZero(node) {
   // TODO(kevinb): merge these two rules together
-  if (canApplyRule(rules.REMOVE_ADDING_ZERO, node)) {
-    const newNode = applyRule(rules.REMOVE_ADDING_ZERO, node);
+  if (canApplyRule(REMOVE_ADDING_ZERO, node)) {
+    const newNode = applyRule(REMOVE_ADDING_ZERO, node);
     return Node.Status.nodeChanged(
       ChangeTypes.REMOVE_ADDING_ZERO, node, newNode);
   }
-  else if (canApplyRule(rules.REMOVE_ADDING_ZERO_REVERSE, node)) {
-    const newNode = applyRule(rules.REMOVE_ADDING_ZERO_REVERSE, node);
+  else if (canApplyRule(REMOVE_ADDING_ZERO_REVERSE, node)) {
+    const newNode = applyRule(REMOVE_ADDING_ZERO_REVERSE, node);
     return Node.Status.nodeChanged(
       ChangeTypes.REMOVE_ADDING_ZERO, node, newNode);
   }
--- a/lib/simplifyExpression/simplify.js
+++ b/lib/simplifyExpression/simplify.js
@@ -1,8 +1,5 @@
-const math = require('mathjs');
-
 const checks = require('../checks');
 const flattenOperands = require('../util/flattenOperands');
-const print = require('../util/print');
 const removeUnnecessaryParens = require('../util/removeUnnecessaryParens');
 const stepThrough = require('./stepThrough');
 
@@ -23,14 +20,7 @@ function simplify(node, debug=false) {
     // removing parens isn't counted as a step, so try it here
     simplifiedNode = removeUnnecessaryParens(flattenOperands(node), true);
   }
-  // unflatten the node.
-  return unflatten(simplifiedNode);
-}
-
-// Unflattens a node so it is in the math.js style, by printing and parsing it
-// again
-function unflatten(node) {
-  return math.parse(print(node));
+  return simplifiedNode;
 }
 
 
--- a/lib/util/clone.js
+++ b/lib/util/clone.js
@@ -1,18 +1,5 @@
-// Simple clone function, which creates a deep copy of the given node
-// And recurses on the children (due to the shallow nature of the mathjs node
-// clone)
-function clone(node) {
-  const copy = node.clone();
-  copy.changeGroup = node.changeGroup;
-  if (node.args) {
-    node.args.forEach((child, i) => {
-      copy.args[i] = clone(child);
-    });
-  }
-  else if (node.content) {
-    copy.content = clone(node.content);
-  }
-  return copy;
+function clone (node) {
+  return JSON.parse(JSON.stringify(node));
 }
 
 module.exports = clone;
--- a/lib/util/defineRuleString.js
+++ b/lib/util/defineRuleString.js
@@ -1,0 +1,24 @@
+const {definePatternRule} = require('math-rules');
+const {parse} = require('math-parser');
+const {traverse} = require('math-traverse');
+
+const defineRuleString = (matchPattern, rewritePattern, constraints) => {
+  const matchAST = parse(matchPattern);
+  const rewriteAST = parse(rewritePattern);
+
+  traverse(matchAST, {
+    leave(node) {
+      delete node.loc;
+    }
+  });
+
+  traverse(rewriteAST, {
+    leave(node) {
+      delete node.loc;
+    }
+  });
+
+  return definePatternRule(matchAST, rewriteAST, constraints);
+};
+
+module.exports = defineRuleString;
--- a/lib/util/flattenOperands.js
+++ b/lib/util/flattenOperands.js
@@ -1,6 +1,7 @@
-const evaluate = require('./evaluate');
-const Negative = require('../Negative');
-const Node = require('../node');
+const {build, query} = require('math-nodes');
+const {replace} = require('math-traverse');
+
+const clone = require('../util/clone');
 
 /*
 Background:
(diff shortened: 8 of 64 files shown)


test/simplifyExpression/oneStep.test.js
```diff
@@ -6,28 +6,28 @@ const ChangeTypes = require('../../lib/ChangeTypes');
...
-  const nodeStatus = steps[0];
-  if (debug) {
-    if (!nodeStatus.changeType) {
-      throw Error('missing or bad change type');
-    }
-    // eslint-disable-next-line
-    console.log(nodeStatus.changeType);
-    // eslint-disable-next-line
-    console.log(print(nodeStatus.newNode));
-  }
```
kevinbarabash: Code that's outside of an `it` block will get run regardless so whether the `describe` is disabled (via .skip) or not.

test/solveEquation/solveEquation.test.js
```diff
@@ -6,21 +6,21 @@ const solveEquation = require('../../lib/solveEquation');
...
 
 function testSolve(equationString, outputStr, debug=false) {
-  const steps = solveEquation(equationString, debug);
-  let lastStep;
-  if (steps.length === 0) {
-    lastStep = NO_STEPS;
-  }
-  else {
-    lastStep = steps[steps.length -1].newEquation.print();
-  }
```
kevinbarabash: Same thing as above.  All code should go in and `it`... or `beforeEach` or `afterEach`.

lib/simplifyExpression/basicsSearch/reduceMultiplicationByZero.js
```diff
@@ -1,27 +1,16 @@
...
-      return polyTerm.getCoeffValue() === 0;
-    }
-    return false;
-  });
-  if (zeroIndex >= 0) {
-    // reduce to just the 0 node
-    const newNode = Node.Creator.constant(0);
+  // The math-rules rule for this didn't do the right thing when the '0' was
+  // in the middle of the args list.
+  if (query.isMul(node) && node.args.some(node => query.getValue(node) === 0)) {
```
kevinbarabash: This doesn't do the same thing as the original, but the tests pass so maybe it's good enough.

aliang8:
Thanks for the comments! I will look over this PR soon.

lib/simplifyExpression/basicsSearch/removeAdditionOfZero.js
```diff
@@ -1,30 +1,32 @@
-const clone = require('../../util/clone');
+const {canApplyRule, applyRule} = require('math-rules');
```
aliang8: Do you think we should keep the `require` syntax or move to babel's import syntax on this new branch?

aliang8: It would be tedious to go back and change everything at the end.

kevinbarabash: It doesn't have to be tedious.  We can use codemods to do it automatically for us.  I'd like to minimize the number of changes in this branch so that it's easier to review and easier to merge back into `master.

aliang8: oh, codemods exist. wow!

lib/simplifyExpression/basicsSearch/reduceMultiplicationByZero.js
```diff
@@ -1,27 +1,16 @@
...
-      const polyTerm = new Node.PolynomialTerm(arg);
-      return polyTerm.getCoeffValue() === 0;
-    }
-    return false;
-  });
-  if (zeroIndex >= 0) {
-    // reduce to just the 0 node
-    const newNode = Node.Creator.constant(0);
+  // The math-rules rule for this didn't do the right thing when the '0' was
+  // in the middle of the args list.
```
aliang8: So, `2 * 0 * 2` would not give you `0 * 2`?

lib/simplifyExpression/basicsSearch/removeAdditionOfZero.js
```diff
@@ -1,30 +1,32 @@
...
+
+// e.g. 0 + 2 -> 2
+const REMOVE_ADDING_ZERO_REVERSE = defineRuleString('0 + #a', '#a');
+
 // If `node` is an addition node with 0 as one of its operands,
 // remove 0 from the operands list. Returns a Node.Status object.
 function removeAdditionOfZero(node) {
-  if (node.op !== '+') {
-    return Node.Status.noChange(node);
+  // TODO(kevinb): merge these two rules together
```
aliang8: should we keep these todos here or do we already have this in `math-rules` for reference?

lib/simplifyExpression/basicsSearch/reduceMultiplicationByZero.js
kevinbarabash: There's a test in https://github.com/socraticorg/mathsteps/blob/master/test/simplifyExpression/basicsSearch/reduceMutliplicationByZero.test.js that checks that `2 * 0 * z^2 --> 0`.

lib/simplifyExpression/basicsSearch/removeAdditionOfZero.js
kevinbarabash: I can remove this TODO, we already have this captured in the `math-rules` github issues.

lib/simplifyExpression/basicsSearch/reduceMultiplicationByZero.js
aliang8: Oh I was just curious as to why : 
```
// e.g. x * 0 -> 0
export const MULTIPLY_BY_ZERO = defineRuleString('#a * 0', '0')

// e.g. 0 * x -> 0
export const MULTIPLY_BY_ZERO_REVERSE = defineRuleString('0 * #a', '0')
``` 

you didn't use these.

aliang8: I guess the new rule you wrote works better as it simplifies the entire expression to 0 as opposed to term by term.

test/simplifyExpression/oneStep.test.js
aliang8: Good to know!

lib/simplifyExpression/basicsSearch/reduceMultiplicationByZero.js
kevinbarabash: I suppose we could run one and then run the other one again, but it still doesn't help with situations like `a * b * c * 0 * e * f * g` b/c running each once would result in `a * b * 0 * f * g` so we'd have to run them multiple times and I believe the intent of reduceMultiplicationByZero is that we transform a product with at least one factor that is `0` to `0`.

aliang8: Mhmm agreed.

aliang8:
Great! This PR looks good to merge. Thanks for leading the way Kevin!

lib/simplifyExpression/basicsSearch/reduceMultiplicationByZero.js
kevinbarabash: We could write a rule like this:
```
defineRuleString('#a', '0', { a: node => node.args.some(arg => query.getValue(arg) === 0) });
```
and then we could use `canApplyRule` and `applyRule`.

aliang8: Yup. To keep the structure uniform.

lib/simplifyExpression/basicsSearch/removeAdditionOfZero.js
```diff
@@ -1,30 +1,32 @@
...
-  if (zeroIndex >= 0) {
-    // remove the 0 node
-    newNode.args.splice(zeroIndex, 1);
-    // if there's only one operand left, there's nothing left to add it to,
-    // so move it up the tree
-    if (newNode.args.length === 1) {
-      newNode = newNode.args[0];
-    }
+  else if (canApplyRule(REMOVE_ADDING_ZERO_REVERSE, node)) {
+    const newNode = applyRule(REMOVE_ADDING_ZERO_REVERSE, node);
```
kevinbarabash: `canApplyRule` and `applyRule` both do their own searches so for more complicated expressions they'd return the root whereas the previous implementation would return false until `TreeSearch` finds the node that can actually be transformed in which case it returns that node.  I'm going to look at swapping out `TreeSearch` and figure out how to return the correct node.

kevinbarabash:
> Great! This PR looks good to merge. Thanks for leading the way Kevin!

Please don't merge it yet.  There's still some unresolved issues that I'd like to take care of first.

aliang8:
I wasn't going to haha, sorry if I was a bit rushed in saying that.

kevinbarabash:
Sorry for make this diff so big.  I want to make sure `simplifyExpression` was working for at least a couple of situations.  I wanted to minimize the changes to `simplify.js` and `stepThrough.js` so I ended up having to update `removeUnnecessaryParens` and `flattenOperands` while making sure their tests pass.

aliang8:
All good. I'll take a look at this PR tomorrow at work!

kevinbarabash:
There are 40 passing tests and 462 pending.

kevinbarabash:
You may find reviewing each commit easier since they're more self contained.

lib/checks/hasUnsupportedNodes.js
```diff
@@ -1,30 +1,26 @@
...
-  }
-  else if (Node.Type.isSymbol(node) || Node.Type.isConstant(node)) {
-    return false;
-  }
-  else if (Node.Type.isFunction(node, 'abs')) {
-    if (node.args.length !== 1) {
-      return true;
+  if (query.isApply(node)) {
+    if (query.isAbs(node)) {
+      if (node.args.length !== 1) {
```
aliang8: Maybe you could do `return node.args.length !== 1 || !resolvesToConstant(node.args[0])`. But I guess for now we should keep the format and minimize the changes.

lib/checks/hasUnsupportedNodes.js
```diff
@@ -1,30 +1,26 @@
...
+        return true;
+      }
+      if (!resolvesToConstant(node.args[0])) {
+        return true;
+      }
     }
-    if (node.args.some(hasUnsupportedNodes)) {
-      return true;
+    else if (query.isNthRoot(node)) {
```
aliang8: Not entirely sure if this file has any use. Our parser already catches and throws an error for incorrect parameters.

lib/checks/hasUnsupportedNodes.js
```diff
@@ -1,30 +1,26 @@
...
-  }
-  else if (Node.Type.isFunction(node, 'abs')) {
-    if (node.args.length !== 1) {
-      return true;
+  if (query.isApply(node)) {
+    if (query.isAbs(node)) {
+      if (node.args.length !== 1) {
+        return true;
+      }
+      if (!resolvesToConstant(node.args[0])) {
```
aliang8: This is saying that an expression like `|-(x+4)|` `hasUnsupportedNodes` and will not be evaluated?

lib/util/clone.js
```diff
@@ -1,18 +1,5 @@
...
-    node.args.forEach((child, i) => {
-      copy.args[i] = clone(child);
-    });
-  }
-  else if (node.content) {
-    copy.content = clone(node.content);
-  }
-  return copy;
+function clone (node) {
+  return JSON.parse(JSON.stringify(node));
```
aliang8: Yay!

lib/util/flattenOperands.js
```diff
@@ -34,293 +35,69 @@ interested in how that works
...
+          }, [])
+        );
       }
-      else {
-        parentOp = node.op;
+      else if (query.isMul(node)) {
+        if (isPolynomialTerm(node) && query.isNumber(node.args[0])
+            && node.args.slice(1).every(arg => !query.isNumber(arg))) {
+          const newNode = clone(node);
+          newNode.implicit = true;
```
aliang8: I like this. It would be nice if you could add comments above each of these cases. For this one you could say: `e.g.  2 * x^2 * y ->  2 x^2 y`.

lib/util/flattenOperands.js
```diff
@@ -34,293 +35,69 @@ interested in how that works
...
+          return newNode;
+        }
+        else if (query.isNumber(node.args[0]) && query.isDiv(node.args[1])
+            && query.isIdentifier(node.args[1].args[0]) && node.implicit) {
```
aliang8: You could change this to `getNumerator` after we merge it in `math-nodes`.

lib/util/flattenOperands.js
```diff
@@ -34,293 +35,69 @@ interested in how that works
...
-      else {
-        parentOp = node.op;
+      else if (query.isMul(node)) {
+        if (isPolynomialTerm(node) && query.isNumber(node.args[0])
+            && node.args.slice(1).every(arg => !query.isNumber(arg))) {
+          const newNode = clone(node);
+          newNode.implicit = true;
+          return newNode;
+        }
+        else if (query.isNumber(node.args[0]) && query.isDiv(node.args[1])
```
aliang8: Is this case: `2 * x / 3 -> 2x / 3`?

lib/util/flattenOperands.js
aliang8: Instead of `query.isIdentifier...` shouldn't it be `query.isPolynomialTerm`, since we would also want to transform a case like: `2 (x + 1) / 3  -> 2(x+1) / 3`

lib/checks/hasUnsupportedNodes.js
kevinbarabash: It's probably worth having some sort of validation function to check that ASTs constructed manually are valid.

lib/checks/hasUnsupportedNodes.js
kevinbarabash: This comes from line 24.  Not sure what the original intent was.

lib/checks/hasUnsupportedNodes.js
kevinbarabash: I missed the call to `hasUnsupportedNodes` that checks if `node.args` has any unsupported nodes.  I'm going to keep formatting similar to what was there originally so that it's easier to review.

lib/checks/hasUnsupportedNodes.js
```diff
@@ -1,30 +1,26 @@
...
-  else if (Node.Type.isOperator(node)) {
-    return node.args.some(hasUnsupportedNodes);
-  }
-  else if (Node.Type.isSymbol(node) || Node.Type.isConstant(node)) {
-    return false;
-  }
-  else if (Node.Type.isFunction(node, 'abs')) {
-    if (node.args.length !== 1) {
-      return true;
+  if (query.isApply(node)) {
```
kevinbarabash: This covers `Node.Type.isUnaryMinus`, `Node.Type.isOperator`, and `Node.Type.isFunction`.

lib/checks/hasUnsupportedNodes.js
```diff
@@ -1,30 +1,29 @@
...
+    if (query.isAbs(node)) {
+      if (node.args.length !== 1) {
+        return true;
+      }
+      if (hasUnsupportedNodes(node.args[0])) {
+        return true;
+      }
```
kevinbarabash: I've added the call to `hasUnsupportedNodes` back.

aliang8:
Kevin, can you fix the linter issue? Then I can merge the PR and start porting other rules.

evykassirer:
would it be okay for me to check out this PR before merging? I could probably look tomorrow

or since it's still on a feature branch I could just look after merging if you want to get started on stuff soon

evykassirer:
sorry I've been so inactive - I've been wanting to check this out for a while! :) 

finally figured out my internship plans for next term and have a bit more time now

kevinbarabash:
I think the reason why the pre-commit hook for lint isn't working is that it makes some assumptions that aren't always true:
```
"setup-hooks": "ln -s ../../scripts/git-hooks/pre-commit.sh .git/hooks/pre-commit"
```
I've been quite pleased with https://www.npmjs.com/package/pre-commit.  Maybe we can use it here as well.

lib/checks/hasUnsupportedNodes.js
evykassirer: I believe the existing absolute value code didn't support input with variables, since it would need to be split out into two cases and we have no way to support that kind of branching in mathsteps right now - we want to make sure to return early and just say it's not supported instead of getting stuck there

lib/checks/hasUnsupportedNodes.js
```diff
@@ -1,30 +1,29 @@
...
+      if (!resolvesToConstant(node.args[0])) {
```
evykassirer: would you mind adding a comment here to explain why it's not supported? I can also make a note to add that later :)

lib/checks/hasUnsupportedNodes.js
```diff
@@ -1,30 +1,29 @@
...
-  }
-  else if (Node.Type.isFunction(node, 'abs')) {
-    if (node.args.length !== 1) {
-      return true;
+  if (query.isApply(node)) {
+    if (query.isAbs(node)) {
+      if (node.args.length !== 1) {
+        return true;
+      }
+      if (hasUnsupportedNodes(node.args[0])) {
```
evykassirer: since all the `Apply` nodes are caught in the same if branch and all check `hasUnsupportedNodes`, we could just check that in the beginning for all of them

lib/checks/hasUnsupportedNodes.js
```diff
@@ -1,30 +1,29 @@
-const Node = require('../node');
+const {query} = require('math-nodes');
+
 const resolvesToConstant = require('./resolvesToConstant');
 
 function hasUnsupportedNodes(node) {
-  if (Node.Type.isParenthesis(node)) {
```
evykassirer: have we totally gotten rid of parenthesis nodes? :o cool

what happens in the case where we want to use them for pedagogical reasons like in grouping like terms?

lib/checks/hasUnsupportedNodes.js
evykassirer: I wonder if math-rules could do the check if the args.length is 1? And just return false if not, since then it wouldn't be a valid abs()

Maybe this is more confusing though

lib/checks/resolvesToConstant.js
```diff
@@ -1,24 +1,21 @@
...
+  else if (query.isNumber(node)) {
     return true;
   }
-  else if (Node.Type.isSymbol(node)) {
+  else if (query.isIdentifier(node)) {
     return false;
   }
-  else if (Node.Type.isUnaryMinus(node)) {
-    return resolvesToConstant(node.args[0]);
+  else if (node.type === 'Parentheses') {
```
evykassirer: can this be a function somewhere? I don't really like typing out 'Parentheses' every time

also why are you assuming there are no parenthesis nodes in some functions like `hasUnsupportedNodes` but have the check here?

lib/simplifyExpression/basicsSearch/reduceMultiplicationByZero.js
```diff
@@ -1,25 +1,21 @@
+const {query: q} = require('math-nodes');
```
evykassirer: should we be consistent with if we use query or q?

lib/simplifyExpression/basicsSearch/reduceMultiplicationByZero.js
```diff
@@ -1,25 +1,21 @@
...
+const {rewriteNode} = require('math-rules');
+
 const ChangeTypes = require('../../ChangeTypes');
+const defineRuleString = require('../../util/defineRuleString');
 const Node = require('../../node');
 
+// The math-rules rule for this didn't do the right thing when the '0' was
+// in the middle of the args list.
```
evykassirer: I'd actually prefer to define as many rules in mathsteps as possible

You could remove this comment?

lib/simplifyExpression/basicsSearch/reduceMultiplicationByZero.js
```diff
@@ -1,25 +1,21 @@
...
+const REDUCE_MULTIPLY_BY_ZERO = defineRuleString('#a', '0', {
+  a: node => q.isMul(node) && node.args.some(arg => q.getValue(arg) === 0),
```
evykassirer: I wonder if this would be easier to read multi-line (I get it now, but I also know what the rule is supposed to do)

lib/simplifyExpression/basicsSearch/removeAdditionOfZero.js
```diff
@@ -1,30 +1,30 @@
...
 const ChangeTypes = require('../../ChangeTypes');
+const defineRuleString = require('../../util/defineRuleString');
 const Node = require('../../node');
 
+// e.g. 2 + 0 -> 2
+const REMOVE_ADDING_ZERO = defineRuleString('#a + 0', '#a');
+
+// e.g. 0 + 2 -> 2
+const REMOVE_ADDING_ZERO_REVERSE = defineRuleString('0 + #a', '#a');
+
```
evykassirer: what happens if 0 is in the middle? or do you need to do this to match the end cases - would be worth adding a comment to explain why you have both

lib/simplifyExpression/simplify.js
```diff
@@ -23,14 +20,7 @@ function simplify(node, debug=false) {
     // removing parens isn't counted as a step, so try it here
     simplifiedNode = removeUnnecessaryParens(flattenOperands(node), true);
   }
-  // unflatten the node.
-  return unflatten(simplifiedNode);
-}
-
-// Unflattens a node so it is in the math.js style, by printing and parsing it
-// again
-function unflatten(node) {
```
evykassirer: yay for not needing to do things for mathjs :D .

lib/util/defineRuleString.js
```diff
@@ -0,0 +1,24 @@
+const {definePatternRule} = require('math-rules');
+const {parse} = require('math-parser');
+const {traverse} = require('math-traverse');
+
+const defineRuleString = (matchPattern, rewritePattern, constraints) => {
+  const matchAST = parse(matchPattern);
+  const rewriteAST = parse(rewritePattern);
+
+  traverse(matchAST, {
```
evykassirer: Why do you delete `loc` when defining the rule? It seems like it'd make more sense to do this just once at the beginning of a step before trying all the rule matches.

Or maybe I don't understand where `loc` gets set

evykassirer: (`loc` is like changeGroups right? Maybe I need to do some research on this)

lib/util/flattenOperands.js
```diff
@@ -34,293 +35,80 @@ interested in how that works
...
+      if (query.isAdd(node)) {
+        return build.add(
+          ...node.args.reduce((accum, arg) => {
+            if (query.isAdd(arg)) {
+              return accum.concat(arg.args);
+            }
+            else {
+              return accum.concat(arg);
+            }
+          }, [])
```
evykassirer: semi-colon here I think? linter should catch this though once that's working

lib/util/flattenOperands.js
```diff
@@ -34,293 +35,80 @@ interested in how that works
...
-        // that after collecting the operands, they'll be children args of *
-        parentOp = '*';
-      }
-      else if (node.op === '-') {
-        // Subtraction is flattened in partner with addition, This means that
-        // after collecting the operands, they'll be children args of +
-        parentOp = '+';
+function flattenOperands(root) {
+  return replace(root, {
+    leave(node) {
```
evykassirer: this replace/leave/reduce syntax is really pretty 😍 I'm learning stuff, thanks!

lib/util/flattenOperands.js
```diff
@@ -34,293 +35,80 @@ interested in how that works
...
+      if (query.isAdd(node)) {
+        return build.add(
+          ...node.args.reduce((accum, arg) => {
+            if (query.isAdd(arg)) {
+              return accum.concat(arg.args);
```
evykassirer: does this recurse on `arg.args`? or does it only accumulate 2 layers deep (do the tests still test this with your new parser?)

lib/util/defineRuleString.js
kevinbarabash: There parser used add `loc` props on each node describing substring range within the original source string that maps to the node.  The parser no longer does that.  I forgot to update this code.  I'll remove the code that delete `loc` as it's no longer necessary.

lib/util/flattenOperands.js
kevinbarabash: travis runs the linter so I'm not sure why it didn't catch this.

lib/util/flattenOperands.js
kevinbarabash: I just copied what estraverse does.  ^_^

lib/util/flattenOperands.js
kevinbarabash: `leave` is only called on the way back of the tree so `arg.args` will have already been processed by this point.

lib/simplifyExpression/basicsSearch/removeAdditionOfZero.js
kevinbarabash: I'll add a comment.  `math-rules` treats addition and multiplication in a special way.  Since an expression like `a + b + 0 + d + e` is a single `add` node, we'd don't want to have to describe rules for all of the different possible sums that could be expressed.  When `math-rules` is matching an `add` or `mul` pattern it will compare each sub-array of node.args against the array of args in the pattern.  It will compare `0 + #a` against `a + b`, `b + 0`, `0 + d`, etc. until it finds a match.  In this case it finds `0 + d` which is replaces with `d`.

lib/util/flattenOperands.js
evykassirer: oo nice - so elegant!

lib/util/flattenOperands.js
evykassirer: hmm might be worth looking into that - maybe it's missing other stuff too?

lib/simplifyExpression/basicsSearch/removeAdditionOfZero.js
kevinbarabash: I haven't had time to document `math-rules` outside of some comments in the codebase itself.  It would be good if there were docs so that people using it in `mathsteps` have something to refer to when working with it.

lib/simplifyExpression/basicsSearch/reduceMultiplicationByZero.js
kevinbarabash: That's the plan.  Will do.  It's kind of a useless comment.

lib/simplifyExpression/basicsSearch/reduceMultiplicationByZero.js
kevinbarabash: I used `q` so that I didn't have to break as many lines.  I can change it to `query` if you'd like.

lib/simplifyExpression/simplify.js
kevinbarabash: If `unflatten` was to output something for consumption by `mathjs` is it safe to say that `flatten` was only used for input something from `mathjs`?

lib/checks/resolvesToConstant.js
kevinbarabash: I can add a `isParens` to `query` from `math-nodes`.

I was following what was on lines 11-13.  We can leave this out if we remove unnecessary parens from user input before doing any processing on it.

lib/util/flattenOperands.js
```diff
@@ -34,293 +35,80 @@ interested in how that works
...
+          }, [])
+        );
       }
-      else {
-        parentOp = node.op;
+      else if (query.isMul(node)) {
+        // Rewrite explicit multiplication to be implicit if the factors look
+        // like those of a polynomial.  The first factor must be a number and
+        // the following factors must not be numbers.
+        // e.g. 2 * x^2 * y -> 2 x^2 y
```
evykassirer: do you know if 2 * x^2 * 2 becomes 2x^2 * 2? I forget what behaviour I wanted there anyways, but curious

lib/util/flattenOperands.js
```diff
@@ -34,293 +35,80 @@ interested in how that works
...
+              return accum.concat(arg);
+            }
+          }, [])
+        );
       }
-      else {
-        parentOp = node.op;
+      else if (query.isMul(node)) {
+        // Rewrite explicit multiplication to be implicit if the factors look
+        // like those of a polynomial.  The first factor must be a number and
```
evykassirer: there's an extra space before The

lib/util/flattenOperands.js
```diff
@@ -34,293 +35,80 @@ interested in how that works
...
+        // the following factors must not be numbers.
+        // e.g. 2 * x^2 * y -> 2 x^2 y
+        if (isPolynomialTerm(node) && query.isNumber(node.args[0])
```
evykassirer: `isPolynomialTerm` is defined more globally right? I think it's used in a lot of places. maybe not as much now though? still need to read through the rest of this heh

lib/checks/hasUnsupportedNodes.js
kevinbarabash: I messed up.  I should add this back in.  We do still have parenthesis nodes, but they probably won't be that useful.  Things like `(1 + 2) + 3` are represented as:
```
{ 
   type: 'Apply',
   op: 'add',
   args: [
      { type: 'Apply', op: 'add', args: [ 1, 2 ] },  // full structure omitted
      3,
   ]
}
```
`math-parser`'s `print` function will output `(1 + 2) + 3` when passed this AST.  The only time a `paren` node is require is when it's presences can't otherwise be deduced from the structure of the AST.  One example would be `1 + (x)`.

lib/checks/hasUnsupportedNodes.js
kevinbarabash: TBH, I don't know why this is the case.  I was copying what was on line 24.  Personally I'm okay with having non-constant nodes inside of an absolute.  The only thing is that if they are there we can't evaluate it, but it seems like we'd probably want a separate function called `canEvaluate` to check for that.  I'll get rid of this line if you're okay with that.

lib/util/flattenOperands.js
evykassirer: or like x^2 * 2 * x -> x^2 * 2x

lib/util/flattenOperands.js
evykassirer: oh this is defined differently - that might get confusing, if we still keep the old polynomialTerm concept

lib/util/flattenOperands.js
kevinbarabash: Sorry.  I was taught to put two spaces after periods.  I can change.

lib/simplifyExpression/basicsSearch/removeAdditionOfZero.js
aliang8: I started one and I think I shared it with you through Google Docs to take a look at. Did you get the invite or should I resend the link?

lib/util/flattenOperands.js
kevinbarabash: It do anything to `2 * x^2 * 2` or `x^2 * 2 * x` as written b/c it checks that everything after the first factor is not a number.

lib/util/flattenOperands.js
evykassirer: oh! interesting. I'd be open to keeping that, makes it easier to parse sentences from each other

lib/util/flattenOperands.js
kevinbarabash: I left the old one b/c I wanted to keep the scope of this diff limited.  As we convert more of the code to use `math-rules` we'll either rewrite the original one or use this one.  It's possible that we may want two different behaviors.  If that is that case we'll have to come up with new name(s).  I'll add a comment by the definition of this new one.

lib/simplifyExpression/basicsSearch/removeAdditionOfZero.js
kevinbarabash: @aliang8 can you resend the link?  Sorry for missing it.

lib/util/flattenOperands.js
```diff
@@ -34,293 +35,80 @@ interested in how that works
...
+            && node.args.slice(1).every(arg => !query.isNumber(arg))) {
+          const newNode = clone(node);
+          newNode.implicit = true;
+          return newNode;
+        }
+        // Rearrange multiplication of fractions involving variables in the
+        // the numerator.
+        // e.g. 2 x / 3 -> 2x / 3
```
evykassirer: this example is confusing to me
is the first one 2 x/3 and the second 2x / 3? (with the space showing the groupings?)

aliang8: You can think of the first one as 2 (x / 3) and the second as (2x) / 3.

lib/util/flattenOperands.js
```diff
@@ -34,293 +35,80 @@ interested in how that works
...
+        else if (query.isNumber(node.args[0]) && query.isDiv(node.args[1])
```
evykassirer: I've tried to set these things up so I can name things instead of just doing args[..].args[..]

i.e. something like

```
fraction = node.args[1]
numerator = fraction.args[0]
query.isIdentifier(numerator)
```

cause it can be pretty hard to read otherwise

I see how it could get messy here because it's all in the if statement, but I wonder if you could make a helper function? would that be harder to read? just throwing it out there

lib/util/flattenOperands.js
```diff
@@ -34,293 +35,80 @@ interested in how that works
...
+            && query.isIdentifier(node.args[1].args[0]) && node.implicit) {
+          return build.div(
+            build.implicitMul(node.args[0], node.args[1].args[0]),
```
evykassirer: that kind of thing would make this line a lot easier to read too

lib/util/flattenOperands.js
```diff
@@ -34,293 +35,80 @@ interested in how that works
...
+            node.args[1].args[1]);
+        }
+        // Flatten multiplication nest mul nodes as long as they're implicit
```
evykassirer: "they're all" ?

lib/util/flattenOperands.js
```diff
@@ -34,293 +35,80 @@ interested in how that works
...
-  // We recurse on the left side of the tree to find operands so far
-  // Flattening division is always considered part of a bigger picture
-  // of multiplication, so we get operands with '*'
-  let operands = getOperands(node.args[0], '*');
-
-  if (operands.length === 1) {
-    node.args[0] = operands.pop();
-    node.args[1] = flattenOperands(node.args[1]);
-    operands = [node];
+function isPolynomialTerm(node) {
```
evykassirer: add a docstring with a few examples?

lib/util/flattenOperands.js
```diff
@@ -34,293 +35,80 @@ interested in how that works
...
-// Returns true if there is a * node nested in some division, with no other
-// operators or parentheses between them.
-// e.g. returns true: 2*3/4, 2 / 5 / 6 * 7 / 8
-// e.g. returns false: 3/4/5, ((3*2) - 5) / 7, (2*5)/6
-function hasMultiplicationBesideDivision(node) {
-  if (!Node.Type.isOperator(node)) {
-    return false;
+  else if (query.isPow(node)) {
+    const [base, exponent] = node.args;
+    return query.isIdentifier(base) && isPolynomialTerm(exponent);
```
evykassirer: does that mean x^(x^2) counts as a polynomial term here? idk how you want to define this

I would probably keep to the exponent having to be a constant, but it does feel a bit arbitrary

lib/util/flattenOperands.js
```diff
@@ -34,293 +35,80 @@ interested in how that works
...
-    return true;
+  else if (query.isNeg(node)) {
+    return isPolynomialTerm(node.args[0]);
   }
-  // we ony recurse through division
-  if (node.op !== '/') {
-    return false;
+  else if (query.isMul(node)) {
+    return node.args.every(isPolynomialTerm);
   }
```
evykassirer: Do you not need something to catch the rest of the operators/functions and return false by default?

lib/util/removeUnnecessaryParens.js
```diff
@@ -1,174 +1,53 @@
...
-  // e.g. (x+4) + 12 -> x+4 + 12
-  if (node.op === '+') {
-    node.args.forEach((child, i) => {
-      if (Node.Type.isParenthesis(child) &&
-          !canCollectOrCombine(child.content)) {
-        // remove the parens by replacing the child node (in its args list)
-        // with its content
-        node.args[i] = child.content;
+// Note: this function does more than remove unnecessary parens, it also
+// flattens nested expressions, e.g. (1 + x) + (2 + y), sets wasMinus = true on
```
evykassirer: this example doesn't actually explain what flattening means

also good catch on the name being not great - we should deal with that at some point, this function in general is kinda ugly XD

lib/util/removeUnnecessaryParens.js
```diff
@@ -1,174 +1,53 @@
...
+              return accum.concat(arg.args);
+            }
+            else {
+              return accum.concat(arg);
+            }
+          }, [])
+        );
+      }
+      else if (query.isNeg(node)) {
+        const newNode = clone(node);
```
evykassirer: why do you clone?

lib/util/removeUnnecessaryParens.js
```diff
@@ -1,174 +1,53 @@
...
+// flattens nested expressions, e.g. (1 + x) + (2 + y), sets wasMinus = true on
+// negation nodes (this converts x + -y to x - y).
+// TODO(kevinb): rewrite rules should decide whether or not do these conversions
+function removeUnnecessaryParens(root) {
+  return replace(root, {
+    leave(node) {
+      if (node.type === 'Parentheses') {
+        return node.body;
+      }
+      else if (query.isAdd(node)) {
```
evykassirer: wait why is it flattening here too? is this one of those things where flattening isn't working with unnecessary parens and unnecessary parens doesn't work with flattening? maybe we can put them in the same function and have a flag for whether we want to remove the parens or not

lib/util/removeUnnecessaryParens.js
```diff
@@ -1,174 +1,53 @@
...
-  if (node.op === '+') {
-    node.args.forEach((child, i) => {
-      if (Node.Type.isParenthesis(child) &&
-          !canCollectOrCombine(child.content)) {
-        // remove the parens by replacing the child node (in its args list)
-        // with its content
-        node.args[i] = child.content;
+// Note: this function does more than remove unnecessary parens, it also
+// flattens nested expressions, e.g. (1 + x) + (2 + y), sets wasMinus = true on
+// negation nodes (this converts x + -y to x - y).
```
evykassirer: it doesn't actually change the tree structure right? it just changes it so it prints cleaner? it wasn't clear to me at first what was going on there, might be worth explaining differently here or adding a comment below or something

package.json
```diff
@@ -4,6 +4,10 @@
   "description": "Step by step math solutions",
   "main": "index.js",
   "dependencies": {
+    "math-nodes": "^0.1.7",
```
evykassirer: wheee

test/Negative.test.js
```diff
@@ -1,18 +1,19 @@
...
 const print = require('../lib/util/print');
 
 const Negative = require('../lib/Negative');
 
-const TestUtil = require('./TestUtil');
-
 function testNegate(exprString, outputStr) {
-  const inputStr = Negative.negate(flatten(math.parse(exprString)));
-  TestUtil.testFunctionOutput(print, inputStr, outputStr);
+  it(exprString + ' -> ' + outputStr,  () => {
```
evykassirer: thanks so much for fixing all these it things, I've been wanting to get this done since you mentioned it a while ago!

test/TestUtil.js
```diff
@@ -17,7 +17,8 @@ TestUtil.testFunctionOutput = function (fn, input, output) {
 // tests a function that takes in a node and returns a boolean value
 TestUtil.testBooleanFunction = function (simplifier, exprString, expectedBooleanValue) {
   it(exprString + ' ' + expectedBooleanValue, () => {
-    const inputNode = flatten(math.parse(exprString));
+    // const inputNode = flatten(parse(exprString));
```
evykassirer: you can get rid of this now right?

evykassirer:
wow kevin this is amazing I'm so excited

thanks for all the work you've been putting into this!!

I gotta do some other things before I sleep, but I'll try to reply to your stuff tomorrow or latest the weekend. Excited to be thinking about these things again!

kevinbarabash:
Eek... I guess I'll work all the feedback into a followup PR.

evykassirer:
yeah sorry I would have preferred this stay open too, but it was hard for Anthony to work off of your stuff because it was in another fork (maybe we can make you a contributor, or there's some other better solution, idk)

evykassirer:
we could also reopen this PR, revert the change, and anthony can rebase later :p

lib/checks/hasUnsupportedNodes.js
evykassirer: ah okay, makes sense. Yeah we have lots of other stuff we can't actually evaluate but we keep it in, we should let this through too!

lib/checks/hasUnsupportedNodes.js
evykassirer: I'm worried that if you don't explicitly have parens, calling flatten will flatten it - should we be worried about that?

lib/checks/resolvesToConstant.js
evykassirer: depends on if we need parenthesis nodes at all - if we find a way to not have them at all then I guess it'd be fine to remove them. depends on the flatten issue above

lib/simplifyExpression/basicsSearch/reduceMultiplicationByZero.js
evykassirer: if it's significantly fewer line breaks I'm ok to keep it like this

lib/simplifyExpression/basicsSearch/removeAdditionOfZero.js
evykassirer: makes sense. would be good to draft something up in google docs (as is being done) and then make a wiki website or something (where the demo lives!)

lib/simplifyExpression/simplify.js
evykassirer: if we allow people to create their own trees, I think we might sometimes need to flatten - this is also why I flatten after every step using mathjs, even though we flatten at the beginning anyways and keep passing the same node around

it's possible we don't need it, but that's why my intuition is that we do

lib/util/flattenOperands.js
evykassirer: would be worth making it more clear in the comment

lib/checks/hasUnsupportedNodes.js
kevinbarabash: I don't really know how `hasUnsupportedNodes` is used.  My assumption from the name of the function is that it's some sort of validator that checks for malformed ASTs.  @evykassirer could you describe what it's used for in the codebase?

lib/checks/resolvesToConstant.js
kevinbarabash: The parser will add parens nodes for things like `(x) + 1`.  The parens aren't needed in this expression b/c it can be written as `x + 1`.  I don't think there's a reason to keep things like `(x)` around past the first step.  I think I mentioned this somewhere else (too many comments), but the parser keeps the structure of things like `(x + y) + z` without having explicit parens nodes.

lib/simplifyExpression/simplify.js
kevinbarabash: Let's keep flatten around until we port everything over and then we can start experimenting.

lib/util/flattenOperands.js
kevinbarabash: I figured it would be nice to be able to describe things like `x^a * x^b --> x^(a+b)` where `a` and `b` are actually `a` and `b`.

lib/util/flattenOperands.js
kevinbarabash: This whole thing needs to be made more robust b/c it will only do this rewriting for the first two factors of the multiplication now.  :\

kevinbarabash: I'll add a TODO when I update the comment.

lib/util/flattenOperands.js
kevinbarabash: I'm not sure what the helper function would look like.

lib/util/flattenOperands.js
kevinbarabash: Explicitly return false would be a good thing to do.

lib/util/removeUnnecessaryParens.js
kevinbarabash: :(  I should've wrote `(1 + x) + (2 + y) --> 1 + x + 2 + y`... I'll update this.

lib/util/removeUnnecessaryParens.js
kevinbarabash: There is no flattening in `mathsteps`' original definition of flattening going on here.  That definition being taking a binary tree and converting it to an n-ary tree.  `math-parser` produces an n-ary tree to start with.  It nests nodes to describe order of operations.  Sometimes that order of operations requires parens, e.g. `(1 + 2 + 3) + 4 + 5 + 6` which results in `(+ (+ 1 2 3) 4 5 6)`.  `removeUnnecessaryParens` is all about removing order of operations nesting which doesn't matter.  This means converting `(+ (+ 1 2 3) 4 5 6)` to `(+ 1 2 3 4 5 6)` which looks suspiciously like the original implementation of `flatten`.

lib/util/removeUnnecessaryParens.js
kevinbarabash: It's probably not necessary here since we're exiting the node.  There are some situations where if you don't clone you can end up with an infinite loop or corrupted AST.

test/util/removeUnnecessaryParens.test.js
```diff
@@ -1,29 +1,30 @@
...
     ['x + (y)', 'x + y'],
     ['x + -(y)', 'x - y'],
     ['((3 - 5)) * x', '(3 - 5) * x'],
     ['((3 - 5)) * x', '(3 - 5) * x'],
     ['(((-5)))', '-5'],
-    ['((4+5)) + ((2^3))', '(4 + 5) + 2^3'],
-    ['(2x^6 + -50 x^2) - (x^4)', '2x^6 - 50x^2 - x^4'],
+    ['((4+5)) + ((2^3))', '4 + 5 + 2^3'],
+    // TODO(kevinb): fix this test
+    // ['(2x^6 + -50 x^2) - (x^4)', '2 x^6 - 50 x^2 - x^4'],
```
kevinbarabash: The output of this is `2 x^6 + -50 x^2 - x^4` which I think is probably better especially since the function is called `removeUnnecessaryParens` not `convertAdditiveInversesToSubtraction`.  😛

lib/checks/hasUnsupportedNodes.js
evykassirer: Oh - good question. It's used to return early and just not do anything.

We send any user input from Socratic to mathjs to try to parse it. If it fails to parse, we don't do anything. If it parses but has mathjs nodes we don't support we also don't do anything. And then we assume that all the mathjs trees we're working with only have one of function/parenthesis/constant/etc nodes (the ones you see around the codebase)

Let me know if I can make that more clear

lib/checks/resolvesToConstant.js
evykassirer: how do you differentiate between cases where you could flatten x + y + x, and cases where it should be kept as (x + y) + z for pedagogical reasons? Or do you think we could make it work by parsing the tree to be flat and then not trying to flatten the tree at all in the process of solving the problem?

lib/simplifyExpression/simplify.js
evykassirer: sounds good! I think it'd be good to keep parenthesis nodes around too then?

lib/util/flattenOperands.js
evykassirer: ah dang, okay cool

yeah flattening the tree is surprisingly (maybe not surprisingly?) kinda messy/edge-casey

lib/util/flattenOperands.js
evykassirer: just a function at the bottom of the file that checks this case

```
function canRearrangeFraction(node) {
  if (!node.args) return false;
  if (!query.isDiv(node.args[1])) return false;
  fraction = node.args[1];
  numerator = fraction.args[0];
  return query.isNumber(node.args[0]) && query.isIdentifier(numerator);
}
````

something like that maybe? just makes it more readable when you nest so deep in the tree

lib/util/flattenOperands.js
evykassirer: ah okay, makes sense! 👍 some logic will change in mathsteps because of this, but I like it

lib/util/removeUnnecessaryParens.js
evykassirer: Ahh yeah, makes sense. Technically that is flattening I guess? Seems like there's some code sharing here, might be cleaner to have a function that "flattens addition/multiplication" like this that both places use

test/util/removeUnnecessaryParens.test.js
evykassirer: hm yeah fair enough

lib/checks/hasUnsupportedNodes.js
kevinbarabash: Thanks!  Sounds like we should be able to get rid of it then since it's specific to `mathjs`.

lib/checks/resolvesToConstant.js
kevinbarabash: TBH, we should stop flattening b/c there's no reason to now.  `math-parser` produces an n-ary AST by default and `math-rules` returns n-ary ASTs.  Thus any nesting of nodes of the same type should be meaningful nesting.

lib/util/flattenOperands.js
kevinbarabash: This should be separate out of the flatten function as we want this behavior but we don't want to always be removing meaningful nesting.

lib/util/flattenOperands.js
kevinbarabash: That makes sense.

kevinbarabash:
> we could also reopen this PR, revert the change, and anthony can rebase later :p

I don't want to hold things up.  Maybe @aliang8 could do some of changes we've identified as wanting to make.

aliang8:
Yup, of course. I'll make the changes Evy requested.

lib/checks/resolvesToConstant.js
evykassirer: makes sense!

lib/checks/hasUnsupportedNodes.js
evykassirer: Yup! If there are valid trees handled by math-parser that aren't handled in mathsteps we might want to keep something like this, but hopefully we can always just default to not doing anything when it's a node we don't recognize

lib/util/removeUnnecessaryParens.js
aliang8: It doesn't change the tree structure. The function adds an extra `wasMinus` property to distinguish between `-3` and `2 - 3`.

lib/util/flattenOperands.js
aliang8: Moved function to `math-nodes`.

lib/checks/hasUnsupportedNodes.js
evykassirer: @aliang8 what do you think?

PR #171, opened 2017-06-22
Porting Arithmetic Search

aliang8:


test/simplifyExpression/arithmeticSearch/arithmeticSearch.test.js
```diff
@@ -6,11 +6,11 @@ function testArithmeticSearch(exprStr, outputStr) {
...
 }
 
-describe.skip('evaluate arithmeticSearch', function () {
+describe('evaluate arithmeticSearch', function () {
   const tests = [
     ['2+2', '4'],
     ['2*3*5', '30'],
-    ['9/4', '9/4'], //  does not divide
+    ['9/4', '9 / 4'], //  does not divide
   ];
```
kevinbarabash: I'm surprised this is all there is for tests.  It might be worth adding some additional tests.  In particular the code says that it prioritizes deep nodes, but the tests don't really show that.  It'd be good to have a nested expression and show the most deeply nested node gets evaluated.  Also, does only a single node get evaluated at a time?  There's lots of expressions this doesn't cover like unary minus, nthRoot, absolute value, and implicit multiplication.

lib/simplifyExpression/arithmeticSearch/index.js
```diff
@@ -2,35 +2,40 @@ const ChangeTypes = require('../../ChangeTypes');
...
 const TreeSearch = require('../../TreeSearch');
+const {build, query} = require('math-nodes');
 
 // Searches through the tree, prioritizing deeper nodes, and evaluates
 // arithmetic (e.g. 2+2 or 3*5*2) on an operation node if possible.
 // Returns a Node.Status object.
 const search = TreeSearch.postOrder(arithmetic);
 
+
+const isIntegerFraction = (node) => query.isNeg(node) ? isIntegerFraction(node.args[0]) : query.isDiv(node) && node.args.every(arg => Number.isInteger(query.getValue(arg)));
```
kevinbarabash: These prime candidates for `math-nodes`.  I would probably also add a `isInteger` helper as well.

lib/simplifyExpression/arithmeticSearch/index.js
```diff
@@ -2,35 +2,40 @@ const ChangeTypes = require('../../ChangeTypes');
...
+
+  if (!node.args.every(child => query.isNumber(child))) {
     return Node.Status.noChange(node);
   }
 
   // we want to eval each arg so unary minuses around constant nodes become
   // constant nodes with negative values
   node.args.forEach((arg, i) => {
-    node.args[i] = Node.Creator.constant(evaluate(arg));
+    node.args[i] = build.number(evaluate(arg));
```
kevinbarabash: Wouldn't `evaluate` need to be updated to handle `math-ast` nodes?  Could also use `math-evaluator` for this.

kevinbarabash: TBH, I don't understand how the tests are passing.  :confused:

test/simplifyExpression/arithmeticSearch/arithmeticSearch.test.js
```diff
@@ -6,11 +6,11 @@ function testArithmeticSearch(exprStr, outputStr) {
   TestUtil.testSimplification(arithmeticSearch, exprStr, outputStr);
```
kevinbarabash: This still uses `mathjs.parse`, see
https://github.com/socraticorg/mathsteps/blob/master/test/TestUtil.js#L26-L33

lib/simplifyExpression/arithmeticSearch/index.js
kevinbarabash: The implementation of `evaluate` is:
```
function evaluate(node) {
  // TODO: once we swap in math-parser, call its evaluate function instead
  return node.eval();
}
```
but there isn't supposed to be an `eval` method on the nodes.  In fact there should be no methods on the nodes.

kevinbarabash:
You could probably enable the `arithmetic stepping` tests as part of this diff.

lib/simplifyExpression/arithmeticSearch/index.js
aliang8: Yeah, I know what you mean. Haha. I'm not sure either, but I changed the evaluate function to use the one from `math-parser`.

lib/simplifyExpression/arithmeticSearch/index.js
evykassirer: try not to have functions like this that goes over multiple lines with ternary operators, cause it's hard to read. just split it into an `if`/`else`

evykassirer: having a helper like this would be helpful yeah - I thought we already had an `isIntegerFraction` exported from somewhere, but yeah that should live somewhere that can be used in multiple places

test/simplifyExpression/arithmeticSearch/arithmeticSearch.test.js
evykassirer: Ah, yeah there's probably fewer tests because arithmetic was the first thing we did and most of those tests are in simplifyExpression from before we had more modular tests. I agree we should add some more here - would be easier to understand what this treeSearch does

- unary minus and implicit multiplication: what kind of things are you thinking of?
- applying roots are handled in a separate root-specific code, and similar for absolute value (since they're function nodes and not operator nodes))

lib/util/evaluate.js
```diff
@@ -1,10 +1,11 @@
+const evaluate = require('math-evaluator').default;
+
 // Evaluates a node to a numerical value
 // e.g. the tree representing (2 + 2) * 5 would be evaluated to the number 20
 // it's important that `node` does not contain any symbol nodes
 
-function evaluate(node) {
-  // TODO: once we swap in math-parser, call its evaluate function instead
```
evykassirer: 🎉

lib/simplifyExpression/arithmeticSearch/index.js
```diff
@@ -2,35 +2,40 @@ const ChangeTypes = require('../../ChangeTypes');
...
   });
 
   // Only resolve division of integers if we get an integer result.
   // Note that a fraction of decimals will be divided out.
-  if (Node.Type.isIntegerFraction(node)) {
+  if (isIntegerFraction(node)) {
```
evykassirer: yeah, put this function wherever the other Node.Type functions are coming from :) (I think Kevin already suggested that)

evykassirer:
talked with Anthony on chat about how we want to use math-rules like in Kevin's PR, not just replace the `type` and `creator` functions

aliang8:
Will close this PR, want to redo these changes

PR #172, opened 2017-06-25
Porting Basic Rules

aliang8:
- Combined basic rules into a single file (`basicRules.js`)
- Combined basic rules test into a single file (`basicTest.js`)
- Changed `index.js` to import from `basicRules.js`


evykassirer:
btw you can run the linter locally before committing! There are also hooks you can set up that don't even let you commit without passing the linter, if that fits with your workflow. Can also set up to stop you from pushing without passing the linter. What you did here (just fixing it after) is also totes, fine - just throwing it out there if it's easier for ya

lib/simplifyExpression/basicsSearch/basicRules.js
```diff
@@ -0,0 +1,224 @@
+const {build, query} = require('math-nodes');
+const {rewriteNode} = require('math-rules');
+
+const ChangeTypes = require('../../ChangeTypes');
+const checks = require('../../checks');
+const defineRuleString = require('../../util/defineRuleString');
+const Node = require('../../node');
+
+// e.g. -(-3) -> 3
+const NEGATION = defineRuleString('--#a', '#a');
```
evykassirer: I wonder if this should deal with chains all at once 

e.g. should -----5 go to ---5 or -5 ?

evykassirer: (we could also just collapse multiple of these steps in a row somewhere else)

alsooo I'd call this RESOLVE_DOUBLE_NEGATION because negation sounds like you're negating it

lib/simplifyExpression/basicsSearch/basicRules.js
```diff
@@ -0,0 +1,224 @@
...
+  if (newNode) {
+    return Node.Status.nodeChanged(
+      ChangeTypes.NEGATION, node, newNode);
+  }
+  else {
+    return Node.Status.noChange(node);
+  }
+}
+
+//const REARRANGE_COEFF = defineRuleString('#b * #a', '#a #b', {a: query.isNumber, b: isPolynomialTerm})
```
evykassirer: why's this commented out?

lib/simplifyExpression/basicsSearch/basicRules.js
```diff
@@ -0,0 +1,224 @@
...
+// denominator. Returns a Node.Status object.
+function removeDivisionByOne(node) {
+  let newNode = null;
+  newNode = rewriteNode(DIVISION_BY_ONE, node);
+  if (newNode) {
+    return Node.Status.nodeChanged(
+      ChangeTypes.DIVISION_BY_ONE, node, newNode);
+  }
+
+  const numerator = node.args[0];
```
evykassirer: you can't assume a node will have `args`, and if `args` is null, `args[0] `will throw an error

lib/simplifyExpression/basicsSearch/basicRules.js
```diff
@@ -0,0 +1,224 @@
...
+  if (query.isNeg(numerator)) {
+    return negation(build.neg(numerator));
```
evykassirer: here you're returning something that's not a NodeStatus, which shouldn't happen (unless I'm missing something)

lib/simplifyExpression/basicsSearch/basicRules.js
```diff
@@ -0,0 +1,224 @@
...
+  }
+
+  newNode = rewriteNode(DIVISION_BY_NEGATIVE_ONE, node);
+
+  if (newNode) {
+    return Node.Status.nodeChanged(
+      ChangeTypes.DIVISION_BY_NEGATIVE_ONE, node, newNode);
+  }
+  else {
+    return Node.Status.noChange(node);
```
evykassirer: seems like a lot of code pattern reuse - what about having a function that took a rewriteRuleString and a change type, something approx like this function sketch:

```
function [insert some good name here](node) {
   const newNode = rewriteNode(rewriteRule, node);
   if (newNode) {
       changeType, node, newNode);
   }
   else {
   }
 }
```
that'd work for most of these right?

lib/simplifyExpression/basicsSearch/basicRules.js
```diff
@@ -0,0 +1,224 @@
...
+  }
+}
+
+const MULTIPLY_BY_ZERO = defineRuleString('#a', '0', {
+  a: node => query.isMul(node)
```
evykassirer: I'd rather you only use `=>` notation if the function is one line, otherwise use braces

lib/simplifyExpression/basicsSearch/basicRules.js
```diff
@@ -0,0 +1,224 @@
...
+}
+
+// e.g. 1 ^ x -> 1
+const REMOVE_EXPONENT_BASE_ONE = defineRuleString('1 ^ #a', '1');
+
+// If `node` is of the form 1^x, reduces it to a node of the form 1.
+// Returns a Node.Status object.
+function removeExponentBaseOne(node) {
+  const newNode = rewriteNode(REMOVE_EXPONENT_BASE_ONE, node);
+  if (newNode && checks.resolvesToConstant(node.args[1])) {
```
evykassirer: we should add a comment to explain the `checks.resolvesToConstant(node.args[1]` part

lib/simplifyExpression/basicsSearch/basicRules.js
```diff
@@ -0,0 +1,224 @@
...
+}
+
+// e.g. x * -1 -> -x
+const REMOVE_MULTIPLYING_BY_NEGATIVE_ONE = defineRuleString('#a * -1', '-#a');
+
+// e.g. -1 * x -> -x
+const REMOVE_MULTIPLYING_BY_NEGATIVE_ONE_REVERSE = defineRuleString('-1 * #a', '-#a');
+
+// If `node` is a multiplication node with -1 as one of its operands,
+// and a non constant as the next operand, remove -1 from the operands
```
evykassirer: "If `node` is a multiplication node with -1 as one of its operands,
 // and a non constant as the next operand,"

doesn't look like you're following this condition - are there tests for this case?

lib/simplifyExpression/basicsSearch/basicRules.js
evykassirer: (can you add a test that fails because of this?)

lib/simplifyExpression/basicsSearch/basicRules.js
evykassirer: we could also just remove the condition, I forget why it's here :p maybe if it's a constant we'd use simplifyArithmetic or resolveDoubleMinus instead? because they're more specific. yeah, that'd be nice

lib/simplifyExpression/basicsSearch/basicRules.js
```diff
@@ -0,0 +1,224 @@
...
+  else {
+    return Node.Status.noChange(node);
+  }
+}
+
+// e.g. x * 1 -> x
+const REMOVE_MULTIPLYING_BY_ONE = defineRuleString('#a * 1', '#a');
+
+// e.g. 1 * x -> x
+const REMOVE_MULTIPLYING_BY_ONE_REVERSE = defineRuleString('1 * #a', '#a');
```
evykassirer: do you think math-rules could implement something that did this reverse thing? 

e.g. defineRuleStringWithReverse('#a * 1', '1 * #a', '#a')

I think that'd be a bit cleaner :) looks like it might come up a lot

lib/simplifyExpression/basicsSearch/index.js
```diff
@@ -3,60 +3,18 @@
...
-  // If this is a + node and one of the operands is 0, get rid of the 0
-  removeAdditionOfZero,
-];
+const SIMPLIFICATION_FUNCTIONS = Object.keys(basic_rules);
 
 const search = TreeSearch.preOrder(basics);
 
 // Look for basic step(s) to perform on a node. Returns a Node.Status object.
 function basics(node) {
   for (let i = 0; i < SIMPLIFICATION_FUNCTIONS.length; i++) {
```
evykassirer: you can just iterate through the keys here, I don't think you need to define `SIMPLIFICATION_FUNCTIONS` anymore

test/simplifyExpression/basicsSearch/basicTest.js
```diff
@@ -0,0 +1,148 @@
...
+      ['x ^ 1', 'x'],
+      ['(x + 1) ^ 1', 'x + 1'],
+    ];
+    tests.forEach(t => testSimplify(t[0], t[1], basic_rules.removeExponentByOne));
+  });
+
+  describe('remove exponent by base one', function() {
+    const tests = [
+        ['1 ^ 2', '1'],
+        //['1 ^ x', '1'],
```
evykassirer: why are these commented out?

btw I don't think the exponent needs to resolve to a constant? Or are there cases where for certain values of `x`, 1^x would be undefined? I guess if the exponent was something complicated and undefined we'd want to catch that? hmmm

test/simplifyExpression/basicsSearch/rearrangeCoefficient.test.js
```diff
@@ -1,11 +0,0 @@
-const rearrangeCoefficient = require('../../../lib/simplifyExpression/basicsSearch/rearrangeCoefficient');
-
-const testSimplify = require('./testSimplify');
-
-describe.skip('rearrangeCoefficient', function() {
-  const tests = [
-    ['2 * x^2', '2x^2'],
```
evykassirer: have these tests been moved *to* somewhere?

test/simplifyExpression/basicsSearch/basicTest.js
```diff
@@ -0,0 +1,148 @@
...
+      ['0 * x', '0'],
+      ['0 x', '0'],
+      ['0 * (x + 1)', '0'],
+    ];
+    tests.forEach(t => testSimplify(t[0], t[1], basic_rules.removeMultiplicationByZero));
+  });
+
+  describe('reduce exponent by zero', function() {
+    const tests = [
+      ['2 ^ 0', '1'],
```
evykassirer: interesting that you now have spaces, I find it more readable to have just 2^0 without spaces. No strong preference though

....but it would be parsed correctly either way right?

lib/simplifyExpression/basicsSearch/basicRules.js
kevinbarabash: See https://github.com/semantic-math/math-rules/issues/10.  The API should probably be:
```
defineRuleString(['#a * 1', '1 * #a'], '#a')
```
Also, I was thinking of making `math-rules` smart enough to do this automatically for binary operators.

test/simplifyExpression/basicsSearch/basicTest.js
kevinbarabash: It would be parsed correctly either way.  Also, `math-parser`'s `print` function doesn't add spaces around the exponent operator, see https://github.com/semantic-math/math-parser/blob/master/lib/__test__/print.test.js#L53-L64.

lib/simplifyExpression/basicsSearch/basicRules.js
aliang8: Good call. I would go with your second suggestion of collapsing multiple steps where we deal with pedagogy.

lib/simplifyExpression/basicsSearch/basicRules.js
aliang8: That addresses this: https://github.com/socraticorg/mathsteps/blob/master/lib/simplifyExpression/basicsSearch/removeDivisionByOne.js#L27-L31

lib/simplifyExpression/basicsSearch/basicRules.js
aliang8: I would keep the rule as it is and change the comment. I pulled them straight from the old rules haha.

lib/simplifyExpression/basicsSearch/index.js
aliang8: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/keys

It does preserve the order.

test/simplifyExpression/basicsSearch/basicTest.js
aliang8: Yeah that condition was unnecessary. I will remove it.

lib/simplifyExpression/basicsSearch/basicRules.js
aliang8: Yup, having an array of match functions would be nice.

evykassirer: sweeetness let's do that

lib/simplifyExpression/basicsSearch/index.js
evykassirer: sweet, good to know

test/simplifyExpression/basicsSearch/basicTest.js
evykassirer: maybe we should parse it the way it'd be printed then

lib/ChangeTypes.js
```diff
@@ -21,9 +21,9 @@ module.exports = {
   // e.g. x * 2 -> 2x
   REARRANGE_COEFF: 'REARRANGE_COEFF',
   // e.g. x ^ 0 -> 1
-  REDUCE_EXPONENT_BY_ZERO: 'REDUCE_EXPONENT_BY_ZERO',
+  REMOVE_EXPONENT_BY_ZERO: 'REDUCE_EXPONENT_BY_ZERO',
```
evykassirer: odd that you only renamed one of these? (same below) 

let's keep the wording consistent - I could go either way

lib/simplifyExpression/basicsSearch/basicRules.js
```diff
@@ -6,22 +6,25 @@ const checks = require('../../checks');
...
 const Node = require('../../node');
 
-// e.g. -(-3) -> 3
-const NEGATION = defineRuleString('--#a', '#a');
-
-// Simplifies two unary minuses in a row by removing both of them.
-// e.g. -(- 4) --> 4
-function negation(node) {
-  const newNode = rewriteNode(NEGATION, node);
+// Helper function to apply a rule or its reverse to the inputNode
```
evykassirer: I have a feeling we'll be using this outside of the basic rules, maybe put it at the top level somewhere? or we can wait until you start on more rules and then figure it out then

lib/simplifyExpression/basicsSearch/basicRules.js
evykassirer: @aliang8  ?

test/simplifyExpression/basicsSearch/.#testSimplify.js
```diff
@@ -1 +0,0 @@
-diamond@Anthonys-MBP.home.10195
```
evykassirer: what's this? can you add it to your gitignore?

test/simplifyExpression/basicsSearch/basicTest.js
```diff
@@ -2,16 +2,19 @@ const assert = require('assert');
...
+  describe('resolve double negation', function() {
     const tests = [
       ['--1','1'],
       ['--x','x'],
       ['--(x + 1)', 'x + 1'],
+      ['----5','--5'],
     ];
-    tests.forEach(t => testSimplify(t[0], t[1], basic_rules.negation));
+
+    tests.forEach(t => testSimplify(t[0], t[1], basics.RESOLVE_DOUBLE_NEGATION, ChangeTypes.RESOLVE_DOUBLE_NEGATION));
```
evykassirer: line's too long, can you put each argument on one line?

test/simplifyExpression/basicsSearch/basicTest.js
```diff
@@ -21,25 +24,20 @@ describe('basic rules', function() {
         // TODO: handle this case better
         //['3x^2 * 5', '5 (3 x^2)']
     ];
-    tests.forEach(t => testSimplify(t[0], t[1], basic_rules.REARRANGE_COEFF));
```
evykassirer: as a side note (since you've already changed it) use camel case and not snake case for naming stuffs

test/simplifyExpression/basicsSearch/basicTest.js
```diff
@@ -87,7 +85,7 @@ describe('basic rules', function() {
       ['0 + x', 'x'],
       ['0 + (x + 1)', 'x + 1'],
     ];
-    tests.forEach(t => testSimplify(t[0], t[1], basic_rules.removeAdditionOfZero));
+    tests.forEach(t => testSimplify(t[0], t[1], basics.REMOVE_ADDING_ZERO, ChangeTypes.REMOVE_ADDING_ZERO, basics.REMOVE_ADDING_ZERO_REVERSE));
```
evykassirer: I find it weird that you're passing in the reverse - how soon can you make the change in math-rules? can you do it before finishing this PR?

otherwise I'd rather you pass in an array than tacking a new argument at the end - it's kinda hard to know what each argument is

lib/simplifyExpression/basicsSearch/basicRules.js
evykassirer: (this still needs to be addressed - wait are you not done going over comments? let me know when you are ^_^)

lib/ChangeTypes.js
aliang8: Oh woops, I meant to rename both to make it uniform. I don't want to switch between `REDUCE` and `REMOVE`.

lib/simplifyExpression/basicsSearch/basicRules.js
aliang8: I was thinking to move it into a separate file as a `util`. Where do you think I should put it?

test/simplifyExpression/basicsSearch/basicTest.js
aliang8: Yeah, I'm not too satisfied with it myself. We can make it cleaner once we have the array parameter working.

evykassirer: do you think you could do that tomorrow? or will it take a while

test/simplifyExpression/basicsSearch/basicTest.js
```diff
@@ -14,15 +14,16 @@ describe('basic rules', function() {
...
 
   describe.skip('rearrange coefficient', function() {
     const tests = [
-        ['y^3 * 5', '5 y^3'],
-        ['yz * 3', '3 yz'],
-        // TODO: handle this case better
-        //['3x^2 * 5', '5 (3 x^2)']
+      ['y^3 * 5', '5 y^3'],
+      ['yz * 3', '3 yz'],
+      // TODO: handle this case better
```
evykassirer: what's up with this?

test/simplifyExpression/basicsSearch/basicTest.js
```diff
@@ -94,16 +102,18 @@ describe('basic rules', function() {
...
+                                    ChangeTypes.REMOVE_EXPONENT_BY_ONE));
   });
 
   describe('remove exponent by base one', function() {
     const tests = [
-        ['1 ^ 2', '1'],
-        //['1 ^ x', '1'],
-        //['1 ^ (x + 1)', '1'],
+      ['1 ^ 2', '1'],
+      //['1 ^ x', '1'],
```
evykassirer: why are these commented out?

test/simplifyExpression/basicsSearch/basicTest.js
aliang8: Waiting for the `math-node` query functions PR to be merged since `rearrangeCoefficient` requires one of the methods.

evykassirer: cool - let me know when this PR is ready to take a look at again :)

PR #173, opened 2017-06-26
Arithmetic Search

aliang8:
- Move `applyRule` to separate file 
-`getRanges` won't import ?? 
- added more tests to `arithmeticSearch.test.js`

### DIFF
--- a/lib/node/Status.js
+++ b/lib/node/Status.js
@@ -1,4 +1,5 @@
 const clone = require('../util/clone');
+const {query} = require('math-nodes');
 const {traverse} = require('math-traverse');
 
 const ChangeTypes = require('../ChangeTypes');
@@ -81,7 +82,7 @@ Status.childChanged = function(node, childStatus, childArgIndex=null) {
     return substeps;
   }
 
-  if (Type.isParenthesis(node)) {
+  if (node.type === 'Parentheses') {
     oldNode.content = childStatus.oldNode;
     newNode.content = childStatus.newNode;
     substeps = updateSubsteps(substeps, (step) => {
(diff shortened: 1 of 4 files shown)


lib/node/Status.js
```diff
@@ -81,7 +81,7 @@ Status.childChanged = function(node, childStatus, childArgIndex=null) {
     return substeps;
   }
 
-  if (Type.isParenthesis(node)) {
+  if (node.type === 'Parentheses') {
```
evykassirer: can you add parenthesis to query? don't want to check it against plain strings like that

lib/simplifyExpression/arithmeticSearch/index.js
```diff
@@ -1,8 +1,96 @@
...
+          }
+        }
+      }
+      else if (node.args.every(query.isNumber)) {
+        return {node};
+      }
+    }
+    return null;
+  },
+  // TODO: replace this with '#eval(#a)'
```
evykassirer: is this a TODO for later or did you wanna do it before code review?

kevinbarabash: This would require some work to handle evaluation of `1 + 2` within `x + 1 + 2 + y`.

lib/node/Status.js
kevinbarabash: @aliang8  is adding `isParens` to `math-nodes` in https://github.com/semantic-math/math-nodes/pull/4/files.

lib/simplifyExpression/arithmeticSearch/index.js
aliang8: Wouldn't that be handled within the `TreeSearch`? @kevinbarabash

evykassirer: that seems like a pretty important case - I don't think TreeSearch won't handle because it searches a node and then each of its children separately, not a subset of its children together

kevinbarabash: `rewiteNode` and `applyRule` will handle the example above when the pattern being matched is `#a + #b` or `#a * #b`, but will not handle matching `#a` which could be anything.  When I added the TODO, it was me being optimistic that we could figure out a way to handle this, but it seems complicated so I'd rather just remove the TODO.  Note that for individual operators we can write rules using the pattern strings, see https://github.com/semantic-math/math-rules/blob/master/lib/rule-list.js#L69-L79.

PR #178, opened 2017-07-05
Basic rules

aliang8:
- added parser support for basic rules
- changed `Creator` function to use `build` functions from `math-nodes`, similarly for `Type` to `query`


lib/Negative.js
```diff
@@ -1,55 +1,31 @@
-const Node = require('./node');
+const clone = require('./util/clone');
+const {build, query} = require('math-nodes');
 
 const Negative = {};
 
-// Returns if the given node is negative. Treats a unary minus as a negative,
```
ldworkin: Why all this code deletion? Is this logic getting handled somewhere else now?

aliang8: IsNegative is now a helper function in `math-nodes`. https://github.com/semantic-math/math-nodes/blob/master/lib/query.js#L23

lib/simplifyExpression/basicsSearch/index.js
```diff
@@ -6,49 +6,42 @@
...
 
-// const rearrangeCoefficient = require('./rearrangeCoefficient');
-// const reduceExponentByZero = require('./reduceExponentByZero');
+const rearrangeCoefficient = require('./rearrangeCoefficient');
+const reduceExponentByZero = require('./reduceExponentByZero');
 const reduceMultiplicationByZero = require('./reduceMultiplicationByZero');
-// const reduceZeroDividedByAnything = require('./reduceZeroDividedByAnything');
+const reduceZeroDividedByAnything = require('./reduceZeroDividedByAnything');
 const removeAdditionOfZero = require('./removeAdditionOfZero');
-// const removeDivisionByOne = require('./removeDivisionByOne');
```
ldworkin: I'm in favor of deleting unused code instead of commenting it out. It's always in git if we need it.

lib/Negative.js
ldworkin: 👍

lib/simplifyExpression/basicsSearch/simplifyDoubleUnaryMinus.js
```diff
@@ -2,36 +2,28 @@ const clone = require('../../util/clone');
...
   // e.g. - -4, -4 could be a constant with negative value
-  else if (Node.Type.isConstant(unaryArg) && parseFloat(unaryArg.value) < 0) {
-    const newNode = Node.Creator.constant(parseFloat(unaryArg.value) * -1);
+  else if (query.isNumber(unaryArg) && parseFloat(val) < 0) {
+    const newNode = build.number(parseFloat(val) * -1);
     return Node.Status.nodeChanged(
       ChangeTypes.RESOLVE_DOUBLE_MINUS, node, newNode);
   }
-  // e.g. -(-(5+2))
-  else if (Node.Type.isParenthesis(unaryArg)) {
```
ldworkin: Why is this block deleted?

lib/util/flattenOperands.js
```diff
@@ -92,23 +88,13 @@ function flattenOperands(root) {
...
-    const [base, exponent] = node.args;
-    return query.isIdentifier(base) && isPolynomialTerm(exponent);
-  }
-  else if (query.isNeg(node)) {
-    return isPolynomialTerm(node.args[0]);
-  }
-  else if (query.isMul(node)) {
-    return node.args.every(isPolynomialTerm);
-  }
+function canRearrangeFraction(node) {
```
ldworkin: Add a docstring here

lib/simplifyExpression/basicsSearch/removeAdditionOfZero.js
```diff
@@ -1,30 +1,32 @@
...
-
-// e.g. 0 + 2 -> 2
-const REMOVE_ADDING_ZERO_REVERSE = defineRuleString('0 + #a', '#a');
+const {query} = require('math-nodes');
 
 // If `node` is an addition node with 0 as one of its operands,
 // remove 0 from the operands list. Returns a Node.Status object.
 function removeAdditionOfZero(node) {
-  let newNode = null;
-  newNode = rewriteNode(REMOVE_ADDING_ZERO, node);
```
ldworkin: Are we not using the rewriteNode function anymore?

lib/util/flattenOperands.js
```diff
@@ -92,23 +88,13 @@ function flattenOperands(root) {
   });
 }
 
-function isPolynomialTerm(node) {
```
ldworkin: Does math-parser have a function for this now or something?

ldworkin:
Looks good! Just left questions about some things I was curious about 👍

lib/simplifyExpression/basicsSearch/simplifyDoubleUnaryMinus.js
```diff
@@ -2,36 +2,28 @@ const clone = require('../../util/clone');
 
 const ChangeTypes = require('../../ChangeTypes');
 const Node = require('../../node');
+const {build, query} = require('math-nodes');
 
 // Simplifies two unary minuses in a row by removing both of them.
 // e.g. -(- 4) --> 4
 function simplifyDoubleUnaryMinus(node) {
-  if (!Node.Type.isUnaryMinus(node)) {
```
kevinbarabash: One way to speed up porting is to change the implementation of `Node.Type.isUnaryMinus` to call `query.isNeg` and do the same for the rest of the functions in `Node.Type`.

lib/util/flattenOperands.js
```diff
@@ -64,13 +60,13 @@ function flattenOperands(root) {
         // Rearrange multiplication of fractions involving variables in the
         // the numerator.
         // e.g. 2 x / 3 -> 2x / 3
-        else if (query.isNumber(node.args[0]) && query.isDiv(node.args[1])
-            && query.isIdentifier(node.args[1].args[0]) && node.implicit) {
+        // TODO (kevinb): only rewrites first two factors of multiplication
```
kevinbarabash: It might be good to give an example of where this fails and describe what should happen in that case.

lib/simplifyExpression/basicsSearch/index.js
kevinbarabash: The reason why I commented these out as opposed to deleting is that I knew for sure that these tests were coming back for sure.

lib/simplifyExpression/basicsSearch/removeAdditionOfZero.js
kevinbarabash: I think there was a desire to introduce the use of `math-rules` as a separate step.

lib/simplifyExpression/basicsSearch/simplifyDoubleUnaryMinus.js
kevinbarabash: The only time the new parser emits a `parentheses` node is when they aren't necessary, e.g `1 + (2)`.  We remove all unnecessary parentheses from user input before proceeding so we should never run into a `parentheses` node here.  In this case `-(-(5+2))` is represented as follows:
```
{
   type: 'Apply',
   op: 'neg',
   args: [{
      type: 'Apply',
      op: 'neg',
      args: [{
         type: 'Apply',
         op: 'add',
         args: [{type: 'Number', value: '5'}, {type: 'Number', value: '2'}],
      }],
   }],
}
```
If you're curious about what different input looks like, there's a demo of the parser at https://semantic-math.github.io/math-parser/.

lib/util/flattenOperands.js
kevinbarabash: There's another repo called `math-nodes` which contains `query` and `build` helpers for querying properties of nodes and building new nodes (and trees) respectively.  `isPolynomialTerm` is defined on `query`.

kevinbarabash:
I think some (or all of) of the tests in the `'subtraction support'` suite in test/simplifyExpression/simplify.test.js could be enabled.

PR #180, opened 2017-07-06
fixes bug #153

aliang8:
https://github.com/socraticorg/mathsteps/commit/1083124f163a196d0686677a63a264df41ea5b84#commitcomment-22937523

### DIFF
--- a/lib/checks/canMultiplyLikeTermConstantNodes.js
+++ b/lib/checks/canMultiplyLikeTermConstantNodes.js
@@ -9,7 +9,8 @@ function canMultiplyLikeTermConstantNodes(node) {
     return false;
   }
   const args = node.args;
-  if (!args.every(n => ConstantOrPowerTerm.isConstantOrConstantPower(n))) {
+  if (!args.every(n => ConstantOrPowerTerm.isConstantOrConstantPower(n)
+                  && ConstantOrPowerTerm.getExponentNode(n).value === 1)) {
     return false;
   }
 
--- a/lib/checks/canMultiplyLikeTermConstantNodes.js
+++ b/lib/checks/canMultiplyLikeTermConstantNodes.js
@@ -9,8 +9,11 @@ function canMultiplyLikeTermConstantNodes(node) {
     return false;
   }
   const args = node.args;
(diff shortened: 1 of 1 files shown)


aliang8:
Ahhh never mind, it introduced some newer bugs.

ldworkin:
👍

lib/checks/canMultiplyLikeTermConstantNodes.js
```diff
@@ -13,6 +13,10 @@ function canMultiplyLikeTermConstantNodes(node) {
     return false;
   }
 
+  if (args.every(arg => ConstantOrPowerTerm.getExponentNode(arg).value === 1)) {
```
evykassirer: actually instead of this I think it'd be better to check that all of them don't have exponents

 i.e. `args.every(arg => !Node.Type.isOperator(arg, '^'))` since they'd all be constants and not power terms. I think it's a bit easier to understand what's going on

what do you think?

evykassirer:
@aliang8 @ldworkin in the future can you do squash and merge instead of merge? makes the commit history cleaner :) that's what Ahmed and I did and what in general we've been doing so far, but feel free to make an argument for changing it to just merging

lib/checks/canMultiplyLikeTermConstantNodes.js
ldworkin: @evykassirer not sure I understand, does this change the functionality at all, or is it just a cosmetic change? cc @aliang8

ldworkin:
@evykassirer yup will squash and merge in the future.

lib/checks/canMultiplyLikeTermConstantNodes.js
evykassirer: mostly cosmetic, but if you had `6^1 * 6^1` it wouldn't add the exponents, it'd just ignore them. I was going to comment on that, but I think that'd never actually happen because of the `_^1 => _` rule that gets applied first

ldworkin: Cool, I see, makes sense!

PR #181, opened 2017-07-06
better fix for multiplying like term constant nodes

ldworkin:
https://github.com/socraticorg/mathsteps/pull/180#discussion_r125961158

### DIFF
--- a/lib/checks/canMultiplyLikeTermConstantNodes.js
+++ b/lib/checks/canMultiplyLikeTermConstantNodes.js
@@ -13,7 +13,9 @@ function canMultiplyLikeTermConstantNodes(node) {
     return false;
   }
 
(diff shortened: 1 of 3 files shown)


evykassirer:
thanks for updating this!

also can you add a test somewhere that failed before and doesn't anymore? tests in this repo are usually fairly quick to add, should just be one line of asciimath input/out example.

ldworkin:
What would that test look like @evykassirer ? Cause the simplification was still "correct" before, as in the final answer was fine, but the steps it took to get there were excessive.

evykassirer:
Oh sorry - true. I'd recommend using `TestUtil.testSubsteps`

evykassirer:
oo maybe `simplifyExpression/oneStep.test.js`

aliang8:
Thanks! This looks cleaner.

evykassirer:
sweet! glad that was easy

PR #183, opened 2017-07-06
Simplify roots of any degree (fixes cube root bug)

aliang8:
Cube root grouping terms under the radicand is incorrect. In the image below, note there are groups of 2 2's instead of groups of 3 since it is a cube root.

https://cl.ly/0u0q3V0r3q2n


evykassirer:
the image was really helpful for knowing what's going on here! can you also give a little description? it wasn't super clear to me just from the pic

evykassirer:
also (this is a Socratic convention that I also use at Joyable!) when you're making branches, preface them with your initials so we know it's a branch you were working on

e.g. `al_cube_root`

lib/simplifyExpression/functionsSearch/nthRoot.js
```diff
@@ -197,46 +198,61 @@ function getFactorNodes(node) {
...
-      children = children.concat(radicandNode.args.slice(i, j));
-    }
-    i = j;
-  }
+  const terms = radicandNode.args.map(arg => print(arg));
+
+  // Collect groups when there are duplicate terms in the radicand
+  if ([...new Set(terms)].length !== terms.length) {
+    // dictionary storing the number of times a constant appears
+    // e.g. 2 * 2 * 2 => {'2': 3} , 2 appears 3 times
```
evykassirer: 👌 example

lib/simplifyExpression/functionsSearch/nthRoot.js
```diff
@@ -197,46 +198,61 @@ function getFactorNodes(node) {
...
-    }
-    else {
-      children = children.concat(radicandNode.args.slice(i, j));
-    }
-    i = j;
-  }
+  const terms = radicandNode.args.map(arg => print(arg));
+
+  // Collect groups when there are duplicate terms in the radicand
+  if ([...new Set(terms)].length !== terms.length) {
```
evykassirer: I'm a fan of returning early instead of nesting ifs

so you could do 

```
// there is nothing to simplify when there are no duplicates of factors in the radicand
if ([...new Set(terms)].length === terms.length) {
}
```

and then the rest of the code is less indented

lib/simplifyExpression/functionsSearch/nthRoot.js
```diff
@@ -197,46 +198,61 @@ function getFactorNodes(node) {
...
-      const groupedNode = Node.Creator.parenthesis(
-        Node.Creator.operator('*', radicandNode.args.slice(i, j)));
-      children.push(groupedNode);
-    }
-    else {
-      children = children.concat(radicandNode.args.slice(i, j));
-    }
-    i = j;
-  }
+  const terms = radicandNode.args.map(arg => print(arg));
```
evykassirer: can you call this `termStrings` so it's clear that they're the string representation?

lib/simplifyExpression/functionsSearch/nthRoot.js
```diff
@@ -197,46 +198,61 @@ function getFactorNodes(node) {
...
+
+  // Collect groups when there are duplicate terms in the radicand
+  if ([...new Set(terms)].length !== terms.length) {
+    // dictionary storing the number of times a constant appears
+    // e.g. 2 * 2 * 2 => {'2': 3} , 2 appears 3 times
+    const count = {};
+
+    radicandNode.args.forEach(arg => {
+      arg = print(arg);
```
evykassirer: `argString = print(arg)`

lib/simplifyExpression/functionsSearch/nthRoot.js
```diff
@@ -197,46 +198,61 @@ function getFactorNodes(node) {
...
+      count[arg] ? count[arg]++ : count[arg] = 1;
+    });
+
+    const flatten = arr => arr.reduce(
```
evykassirer: since you only use this in one place (and it's not clear at this point what it's used for) I think it'd be cleaner to just use the definition instead of giving it a name - I'll explain below

aliang8:
Cube root grouping terms under the radicand is incorrect. In the case of the image, there are groups of 2 2's instead of 3 groups since it is a cube root.

lib/simplifyExpression/functionsSearch/nthRoot.js
```diff
@@ -197,46 +198,61 @@ function getFactorNodes(node) {
...
+      (acc, val) => acc.concat(val), []
+    );
+
+    newNode = Node.Creator.nthRoot(
+      Node.Creator.operator(
+        '*',
+        flatten(Object.keys(count).map(key => {
+          let leftover = count[key];
```
evykassirer: why is this called leftover?

lib/simplifyExpression/functionsSearch/nthRoot.js
```diff
@@ -197,46 +198,61 @@ function getFactorNodes(node) {
...
+          const term = math.parse(key);
+          const times = rootValue;
+
+          const args = [];
+
+          while (leftover - times > 0) {
+            leftover -= times;
+            args.push(Node.Creator.parenthesis(
+              Node.Creator.operator('*', Array(times).fill(term))));
```
evykassirer: > Array(times).fill(term)

cool I didn't know you could do that with javascript :D

lib/simplifyExpression/functionsSearch/nthRoot.js
```diff
@@ -197,46 +198,61 @@ function getFactorNodes(node) {
...
+      Node.Creator.operator(
+        '*',
+        flatten(Object.keys(count).map(key => {
+          let leftover = count[key];
+          const term = math.parse(key);
+          const times = rootValue;
+
+          const args = [];
+
+          while (leftover - times > 0) {
```
evykassirer: this while loop should probably have a comment explaining what it's doing

lib/simplifyExpression/functionsSearch/nthRoot.js
```diff
@@ -197,46 +198,61 @@ function getFactorNodes(node) {
...
+            leftover -= times;
+            args.push(Node.Creator.parenthesis(
+              Node.Creator.operator('*', Array(times).fill(term))));
+          }
+          const arg = leftover === 1
```
evykassirer: this seems like a special last arg, not just a generic `arg`  - can you give it a more specific name?

lib/simplifyExpression/functionsSearch/nthRoot.js
```diff
@@ -197,46 +198,61 @@ function getFactorNodes(node) {
...
+    radicandNode.args.forEach(arg => {
+      arg = print(arg);
+      count[arg] ? count[arg]++ : count[arg] = 1;
+    });
+
+    const flatten = arr => arr.reduce(
+      (acc, val) => acc.concat(val), []
+    );
+
+    newNode = Node.Creator.nthRoot(
```
evykassirer: Ahmed taught me a way of writing code that I really like where you name things as you go, instead of nesting

e.g. here's a sketch of this code

```
const termStrings = Object.keys(count);
newTermGroups = termStrings.map(termString => .... )
newTerms = flatten(newTermGroups)  ---- by the way I still don't get what this flattening stuff is for - is it an array of arrays otherwise?
newBase = Node.Creator.operator('*', newTerms)
newNode = Node.Creator.nthRoot(___, rootNode)
```

then you can see how the code works as you read it, instead of having to look deeper into the code and read it from the middle out - also you have nice variable names to explain what each part is

what do you think?

lib/simplifyExpression/functionsSearch/nthRoot.js
evykassirer: also what if this while loop never runs? (i.e. times > leftover to begin with) - it should just return no change right?

lib/simplifyExpression/functionsSearch/nthRoot.js
```diff
@@ -463,11 +479,10 @@ function isMultiplicationOfEqualNodes(node) {
     return false;
   }
 
-  // return if they are all equal nodes
-  return node.args.reduce((a, b) => {
-    return a.equals(b);
-  });
+  const terms = node.args.map(print);
```
evykassirer: `termStrings`

lib/simplifyExpression/functionsSearch/nthRoot.js
```diff
@@ -463,11 +479,10 @@ function isMultiplicationOfEqualNodes(node) {
...
 
+  // return if they are all equal nodes
+  return [...new Set(terms)].length === 1;
```
evykassirer: Neat! I'm wondering if it's less readable - probably okay but Lili can maybe confirm

test/simplifyExpression/functionsSearch/nthRoot.test.js
```diff
@@ -40,23 +40,23 @@ describe('nthRoot steps', function () {
   const tests = [
     ['nthRoot(12)',
       ['nthRoot(2 * 2 * 3)',
-        'nthRoot((2 * 2) * 3)',
-        'nthRoot(2 ^ 2 * 3)',
+        'nthRoot((2 * 2) * 3, 2)',
+        'nthRoot(2 ^ 2 * 3, 2)',
```
ldworkin: I think it defaults to base 2 if none is specified, right?

lib/simplifyExpression/functionsSearch/nthRoot.js
```diff
@@ -41,7 +43,7 @@ function nthRootExponent(node) {
   const rootNode = getRootNode(node);
   const baseNode = radicandNode.args[0];
   const exponentNode = Node.Type.isParenthesis(radicandNode.args[1]) ?
-    radicandNode.args[1].content : radicandNode.args[1];
+        radicandNode.args[1].content : radicandNode.args[1];
```
ldworkin: extra space?

lib/simplifyExpression/functionsSearch/nthRoot.js
ldworkin: I like it, I've used len(set(x))==1 to check for equality before. @aliang8 I'm wondering why this change was made though? Did the old way not work for a particular case?

lib/simplifyExpression/functionsSearch/nthRoot.js
aliang8: If it's easier to decode, I'll go with it. :D

lib/simplifyExpression/functionsSearch/nthRoot.js
aliang8: It's like how many terms are leftover that haven't been grouped.

lib/simplifyExpression/functionsSearch/nthRoot.js
aliang8: Yes

aliang8: This test case should remove the need for that:
`['nthRoot(4, 3)', 'nthRoot(4, 3)'],`

lib/simplifyExpression/functionsSearch/nthRoot.js
aliang8: Mhmhm, a.equals(b) is not a function (I think it works for objects only).

aliang8:
Made the request changes. :D Let me know if there's anything else.

lib/simplifyExpression/functionsSearch/nthRoot.js
```diff
@@ -205,54 +206,62 @@ function groupTermsByRoot(node) {
...
+  // dictionary storing the number of times a constant appears
+  // e.g. 2 * 2 * 2 => {'2': 3} , 2 appears 3 times
+  const count = {};
+
+  radicandNode.args.forEach(arg => {
+    const argString = print(arg);
+    count[argString] ? count[argString]++ : count[argString] = 1;
+  });
+
+  const flatten = arr => arr.reduce(
```
evykassirer: defining it this far before you use it is kinda confusing

can't you just do this below?

```
// Compress array of arrays
const newTerms = newTermGroups.reduce((acc, val) => acc.concat(val), []);
```

lib/simplifyExpression/functionsSearch/nthRoot.js
```diff
@@ -205,54 +206,62 @@ function groupTermsByRoot(node) {
...
+    (acc, val) => acc.concat(val), []
+  );
+
+  const keys = Object.keys(count);
```
evykassirer: I think naming this something more detailed than `keys` would be good

lib/simplifyExpression/functionsSearch/nthRoot.js
```diff
@@ -205,54 +206,62 @@ function groupTermsByRoot(node) {
...
+
+  const newTermGroups = keys.map(key => {
+    let leftover = count[key];
```
evykassirer: leftover is all of the terms though, right? what was taken away that makes this leftover?

evykassirer: oh I see you decrement it as you go... makes sense

lib/simplifyExpression/functionsSearch/nthRoot.js
```diff
@@ -205,54 +206,62 @@ function groupTermsByRoot(node) {
...
+  });
+
+  // Compress array of arrays
+  const newTerms = flatten(newTermGroups);
+
+  const newBase = Node.Creator.operator('*', newTerms);
+
+  newNode = Node.Creator.nthRoot(newBase, rootNode);
+
+  return Node.Status.nodeChanged(
```
evykassirer: this is all a lot cleaner I think ^_^  thanks! Hope you like it too

lib/simplifyExpression/functionsSearch/nthRoot.js
```diff
@@ -222,14 +222,10 @@ function groupTermsByRoot(node) {
     const argString = print(arg);
     count[argString] ? count[argString]++ : count[argString] = 1;
   });
+  
+  const termString = Object.keys(count);
```
evykassirer: mmm this should have an `s` on the end of it probably but oh well :p

evykassirer:
@ldworkin reminder please squash and merge in da future! then we don't have things like `trailing spaces whut` in master 🤣

evykassirer:
once you do it once, it should keep that as the default then on

PR #184, opened 2017-07-07
Factoring fix

aliang8:
fixes: https://cl.ly/0n2k3P1v1K3m vs https://cl.ly/3P0o0m33202m

and https://s3.amazonaws.com/socratic-ocr-uploads/prod/2fe54324-0e6e-4af6-860f-b22ba8ae4d11/raw.jpeg


lib/solveEquation/stepThrough.js
```diff
@@ -68,7 +71,42 @@ function stepThrough(leftNode, rightNode, comparator, debug=false) {
     }
 
     try {
```
evykassirer: this is all in a try block - I think it'll catch a lot more errors than you're thinking for a lot of cases you're not considering. Maybe you want to do that for now, but it seems pretty not great to me - any error could happen, even bugs in your code, and you wouldn't know

lib/solveEquation/stepThrough.js
```diff
@@ -68,7 +71,42 @@ function stepThrough(leftNode, rightNode, comparator, debug=false) {
...
+          const rightNode = Node.Creator.constant(0);
+
+          const lastStep = stepThrough(leftNode, rightNode, '=').slice(-1)[0];
+
+          symbol = lastStep.newEquation.leftNode;
+          // Append to a list of roots
+          solutions.push(lastStep.newEquation.rightNode);
+        }
+
+        const roots = new Equation(symbol, solutions[0], '=');
```
evykassirer: maybe add another comment here pointing out it's just one solution

lib/solveEquation/stepThrough.js
```diff
@@ -68,7 +71,42 @@ function stepThrough(leftNode, rightNode, comparator, debug=false) {
...
+
+        // Return because we are done at this point
+        return steps;
+      }
+      else {
+        equationStatus = step(equation, symbolName);
+      }
     }
     catch (e) {
       // This error happens for some math that we don't support
```
evykassirer: if you decide to keep this hacky code, update this comment at least noting that it will catch a whole bunch of errors, not just "some math we don't support"

edit: oh, I just expanded the catch block and it bubbles up other errors - which is good - but it's gonna be throwing a *lot* of errors for things that don't match exactly what you want (for all the things I commented on above)

lib/solveEquation/stepThrough.js
```diff
@@ -183,11 +221,18 @@ function addSimplificationSteps(steps, equation, debug=false) {
...
 
   const leftSteps = simplifyExpressionNode(equation.leftNode, false);
+  const leftFactorSteps = factor(equation.leftNode, false);
   const leftSubSteps = [];
   for (let i = 0; i < leftSteps.length; i++) {
     const step = leftSteps[i];
     leftSubSteps.push(EquationStatus.addLeftStep(equation, step));
   }
+
+  for (let i = 0; i < leftFactorSteps.length; i++) {
```
evykassirer: `leftFactorSteps.each(step => leftSubSteps.push(EquationStatus.addLeftStep(equation, step)))`

I'm a bit confused - why are there left/right Steps as well as left/right FactorSteps? What's an example that would have steps coming from both?

lib/solveEquation/stepThrough.js
```diff
@@ -68,7 +71,42 @@ function stepThrough(leftNode, rightNode, comparator, debug=false) {
...
+
+        let symbol;
+        // For each factor, solve the equation factor = 0 for the symbol
+        for (var f in factors) {
+          const leftNode = factors[f].content;
+          const rightNode = Node.Creator.constant(0);
+
+          const lastStep = stepThrough(leftNode, rightNode, '=').slice(-1)[0];
+
+          symbol = lastStep.newEquation.leftNode;
```
evykassirer: it might get partway to a solution but not end up with a symbol on the left side (pretty sure)

lib/solveEquation/stepThrough.js
```diff
@@ -68,7 +71,42 @@ function stepThrough(leftNode, rightNode, comparator, debug=false) {
...
+
+        const solutions = [];
+
+        let symbol;
+        // For each factor, solve the equation factor = 0 for the symbol
+        for (var f in factors) {
+          const leftNode = factors[f].content;
+          const rightNode = Node.Creator.constant(0);
+
+          const lastStep = stepThrough(leftNode, rightNode, '=').slice(-1)[0];
```
evykassirer: how do you know this will return any steps? maybe there will be no possible steps and the list will be empty

lib/solveEquation/stepThrough.js
```diff
@@ -68,7 +71,42 @@ function stepThrough(leftNode, rightNode, comparator, debug=false) {
...
+          && (Node.Type.isConstant(equation.rightNode) && parseFloat(equation.rightNode.value) === 0)) {
+        // Gets all factors that aren't a constant
+        const factors = equation.leftNode.args.filter(arg => !Node.Type.isConstant(arg));
+
+        const solutions = [];
+
+        let symbol;
+        // For each factor, solve the equation factor = 0 for the symbol
+        for (var f in factors) {
+          const leftNode = factors[f].content;
```
evykassirer: here you're assuming it's a parenthesis node, which it might not be

e.g. (x-2)*x = 0

lib/solveEquation/stepThrough.js
```diff
@@ -68,7 +71,42 @@ function stepThrough(leftNode, rightNode, comparator, debug=false) {
...
+      // Returns: An equation in the form of symbol = root
+      if (equation.leftNode.op === '*'
+          && (Node.Type.isConstant(equation.rightNode) && parseFloat(equation.rightNode.value) === 0)) {
+        // Gets all factors that aren't a constant
+        const factors = equation.leftNode.args.filter(arg => !Node.Type.isConstant(arg));
+
+        const solutions = [];
+
+        let symbol;
+        // For each factor, solve the equation factor = 0 for the symbol
```
evykassirer: are you assuming there'll always be a solution? there will be a lot of cases where there is not, since `factor` can be anything that's not a constant

e.g. 2^7, 5^x, 33+89, x+y, etc.

lib/solveEquation/stepThrough.js
```diff
@@ -68,7 +71,42 @@ function stepThrough(leftNode, rightNode, comparator, debug=false) {
...
     try {
-      equationStatus = step(equation, symbolName);
+      // Find roots of a equation
+      // Conditions: left side is a mul node, right side is 0
+      // TODO: move polynomial to left if is on the right
+      // TODO: assuming a single symbol in equation
+      // TODO: display all solutions
+      // Returns: An equation in the form of symbol = root
+      if (equation.leftNode.op === '*'
+          && (Node.Type.isConstant(equation.rightNode) && parseFloat(equation.rightNode.value) === 0)) {
```
evykassirer: might be worth adding something to `Node.Type.isConstant` that takes an optional parameter of the value we're looking for - would be cleaner and likely helpful elsewhere!

ldworkin:
@evykassirer Curious if you could elaborate on why you think this is hacky, and what you think a better approach would be? In general I think we might have to go with hacky solutions for now given how crunched for time we are, but I do want to understand what the "right" way to do it would be!

lib/solveEquation/stepThrough.js
```diff
@@ -68,7 +71,42 @@ function stepThrough(leftNode, rightNode, comparator, debug=false) {
     }
 
     try {
-      equationStatus = step(equation, symbolName);
+      // Find roots of a equation
+      // Conditions: left side is a mul node, right side is 0
+      // TODO: move polynomial to left if is on the right
+      // TODO: assuming a single symbol in equation
+      // TODO: display all solutions
```
aliang8: Yes !

lib/solveEquation/stepThrough.js
aliang8: I could use resolvesToConstant, would that take care of some of these?

aliang8: In the cases of 2^7 or 33 + 89, we could use `resolvesToConstant`. What do you propose we do in the other scenarios?

lib/solveEquation/stepThrough.js
aliang8: I have no idea. I was trying to mimic what was already in place. I'm not sure what the original intentions were.

evykassirer:
@ldworkin hacky because of all the edge cases it misses - it will probably throw a lot of errors (for reasons I pointed out in comments). I think you could maybe cover up those errors with a try catch (where the catch catches more), but that'd feel gross to me

ldworkin:
@evykassirer Gotcha, thanks! Again I think we might have to settle for hacky/gross solutions right now, but we can still try to add explanatory comments and clean up the code as much as possible. @aliang8 can you add comments including several examples of cases that are supported and that aren't? Right now it's unclear to me what is covered and what isn't.

lib/ChangeTypes.js
```diff
@@ -167,7 +167,7 @@ module.exports = {
   SUBTRACT_FROM_BOTH_SIDES: 'SUBTRACT_FROM_BOTH_SIDES',
   // e.g. 2 = x -> x = 2
   SWAP_SIDES: 'SWAP_SIDES',
-  // e.g. (x - 2) (x + 2) = 0
+  // e.g. (x - 2) (x + 2) = 0 => x = 2, -2
```
evykassirer: so the right side will be the node "2, -2"? right now it'd only show one factor right? don't lie here :p

lib/solveEquation/stepThrough.js
```diff
@@ -72,31 +71,45 @@ function stepThrough(leftNode, rightNode, comparator, debug=false) {
...
+      // Conditions: left side is a multiplication node, right side is 0
       // TODO: move polynomial to left if is on the right
       // TODO: assuming a single symbol in equation
       // TODO: display all solutions
+      // TODO: add optional param for Node.Type.isConstant
       // Returns: An equation in the form of symbol = root
       if (equation.leftNode.op === '*'
-          && (Node.Type.isConstant(equation.rightNode) && parseFloat(equation.rightNode.value) === 0)) {
+          && (Node.Type.isConstant(equation.rightNode)
+              && parseFloat(equation.rightNode.value) === 0)) {
```
evykassirer: the `&&`'s should line up, you don't need to nest the second one

lib/solveEquation/stepThrough.js
evykassirer: we should do nothing, since there are no factors - but this shouldn't crash if it gets input like that

lib/solveEquation/stepThrough.js
```diff
@@ -72,31 +71,45 @@ function stepThrough(leftNode, rightNode, comparator, debug=false) {
...
 
-          symbol = lastStep.newEquation.leftNode;
-          // Append to a list of roots
-          solutions.push(lastStep.newEquation.rightNode);
+            symbol = lastStep.newEquation.leftNode;
+            // Append to a list of roots
+            solutions.push(lastStep.newEquation.rightNode);
+          }
+          else {
+            // e.g. x (x - 2) = 0, x = 0
```
evykassirer: why is this the case for whenever the else catches? what if it was 2^x = 0 --> there'd be no steps, but 0 isn't a root

test/solveEquation/solveEquation.test.js
```diff
@@ -57,6 +57,13 @@ describe('solveEquation for =', function () {
     ['2(x+3)/3 = 2', 'x = 0'],
     ['( u )/( 0.3) = 4u + 6.28', 'u = -9.42'],
     ['- q - 4.36= ( 2.2q )/( 1.8)', 'q = -1.962'],
+    // Returns only one solution
+    //['5x^2 - 5x - 30 = 0', 'x = -2'],
```
evykassirer: why are these commented out?

test/solveEquation/solveEquation.test.js
```diff
@@ -58,12 +58,15 @@ describe('solveEquation for =', function () {
...
     // Returns only one solution
-    //['5x^2 - 5x - 30 = 0', 'x = -2'],
-    //['x^2 + 3x + 2 = 0', 'x = -1'],
+    ['5x^2 - 5x - 30 = 0', 'x = -2'],
+    ['x^2 + 3x + 2 = 0', 'x = -1'],
     ['x^2 - x = 0', 'x = 0'],
     ['x^2 + 2x - 15 = 0', 'x = 3'],
     ['x^2 + 2x = 0', 'x = 0'],
     ['x^2 - 4 = 0', 'x = -2'],
+    // TODO: no steps because we need to up our factoring support
```
evykassirer: hehe

lib/ChangeTypes.js
```diff
@@ -167,6 +167,8 @@ module.exports = {
   SUBTRACT_FROM_BOTH_SIDES: 'SUBTRACT_FROM_BOTH_SIDES',
   // e.g. 2 = x -> x = 2
   SWAP_SIDES: 'SWAP_SIDES',
+  // e.g. (x - 2) (x + 2) = 0 => x = 2, -2
```
evykassirer: so the right side will be the node "2, -2"? right now it'd only show one factor right? don't lie here :p

lib/solveEquation/stepThrough.js
```diff
@@ -68,6 +70,52 @@ function stepThrough(leftNode, rightNode, comparator, debug=false) {
     }
 
     try {
+      if (equation.leftNode.op === '*'
+          && (Node.Type.isConstant(equation.rightNode)
+              && parseFloat(equation.rightNode.value) === 0)) {
```
evykassirer: the &&'s should line up, you don't need to nest the second one

lib/ChangeTypes.js
aliang8: Yeah, not sure how we want to represent the right side.

lib/solveEquation/stepThrough.js
```diff
@@ -68,6 +70,52 @@ function stepThrough(leftNode, rightNode, comparator, debug=false) {
...
+        const roots = new Equation(symbol, solutions[0], '=');
+        steps.push(new EquationStatus(
+          ChangeTypes.FIND_ROOTS, equation, roots));
+
+        // Return because we are done at this point
+        return steps;
+      }
+    }
+    catch (e) {
+      throw e; // bubble up
```
evykassirer: if you always bubble up, you don't have to try / catch (it bubbles up by default)

lib/solveEquation/stepThrough.js
```diff
@@ -68,6 +70,52 @@ function stepThrough(leftNode, rightNode, comparator, debug=false) {
...
+          const leftNode = Node.Type.isParenthesis(factors[f])
+                ? factors[f].content
+                : factors[f];
+
+          const rightNode = Node.Creator.constant(0);
+
+          const allSteps = stepThrough(leftNode, rightNode, '=');
+
+          if (allSteps.length > 0) {
+            const lastStep = stepThrough(leftNode, rightNode, '=').slice(-1)[0];
```
evykassirer: you don't have to step through twice, you can just do `allSteps.slice(-1)[0]`

lib/solveEquation/stepThrough.js
aliang8: I was leaving the catch in case we want to address some errors in there.

lib/solveEquation/stepThrough.js
```diff
@@ -68,6 +70,52 @@ function stepThrough(leftNode, rightNode, comparator, debug=false) {
...
+
+            symbol = lastStep.newEquation.leftNode;
```
evykassirer: you still haven't addressed:

> it might get partway to a solution but not end up with a symbol on the left side (pretty sure)

lib/solveEquation/stepThrough.js
evykassirer: actually I think this works. But you should dig in a bit and understand how it works and add comments explaining it - especially if you keep working in this area of the code you should know what's going on!

lib/solveEquation/stepThrough.js
aliang8: Can you give me an example of this?

lib/solveEquation/stepThrough.js
aliang8: Oh yeah, forgot to bring that out.

aelnaiem:
Don't mean to slow this down so leaving this as a future point, but I feel that this logic should live in the factoring module, so maybe a future TODO would be to move it into it's specific module

aliang8:
Hey Ahmed! I don't think this logic can go into the factoring module. We have to first recognize that we are dealing with an equation that is `= 0` before we can find the roots.

ldworkin:
@evykassirer is this good to approve?

lib/checks/canFindRoots.js
```diff
@@ -0,0 +1,33 @@
...
+const resolvesToConstant = require('./resolvesToConstant.js');
+
+function canFindRoots(equation) {
+
+  // Equation should be in the form of factor * factor = 0, factor^power = 0, or identifier = 0
+  // e.g (x - 2)^2 = 0, x(x + 2)(x - 2) = 0, x = 0
+  const left = equation.leftNode;
+  const right = equation.rightNode;
+
+  const zeroRightSide = Node.Type.isConstant(right)
```
evykassirer: it'd be nice to have a helper function where you can just do `Node.Type.isConstant(right, 0)` but I guess you don't need that for now (would be really nice though ^^)

lib/checks/canFindRoots.js
```diff
@@ -0,0 +1,33 @@
...
+        && parseFloat(right.value) === 0;
+
+  const hasRoots = Node.Type.isOperator(left, '*') || Node.Type.isOperator(left, '^');
```
evykassirer: I feel like this isn't a good name for this variable, because this condition could be true without it having roots

I'd just add this into the if below if you can't think of a better name

though think about this: what if the left side is parens e.g. `((x-2)^2)` will it still work (maybe that gets handled elsewhere)?

lib/checks/canFindRoots.js
```diff
@@ -0,0 +1,33 @@
...
+
+  if (!(zeroRightSide && hasRoots)) {
+    return false;
+  }
+
+  if (Node.Type.isOperator(left, '*')) {
+    const factors = left.args.filter(arg => !resolvesToConstant(arg));
```
evykassirer: add comments explaining why you do this check, same below

lib/checks/canFindRoots.js
```diff
@@ -0,0 +1,33 @@
+const Node = require('../node');
+const resolvesToConstant = require('./resolvesToConstant.js');
+
+function canFindRoots(equation) {
+
+  // Equation should be in the form of factor * factor = 0, factor^power = 0, or identifier = 0
+  // e.g (x - 2)^2 = 0, x(x + 2)(x - 2) = 0, x = 0
```
evykassirer: does it return true for `identifier = 0` though? It seems like you only let * and ^ operator nodes through

lib/solveEquation/stepThrough.js
```diff
@@ -67,6 +71,27 @@ function stepThrough(leftNode, rightNode, comparator, debug=false) {
...
 
+    if (checks.canFindRoots(equation)) {
+      const [solutions, symbol] = getSolutionsAndSymbol(equation);
+
+      // Returning a single solution for now.
+      const roots = new Equation(symbol, solutions[0], '=');
+      steps.push(new EquationStatus(
+        ChangeTypes.FIND_ROOTS, equation, roots));
+
+      // Return because we are done at this point
```
evykassirer: I don't think you need this comment

lib/solveEquation/stepThrough.js
```diff
@@ -67,6 +71,27 @@ function stepThrough(leftNode, rightNode, comparator, debug=false) {
...
+      return steps;
+    }
+    else if (checks.canFindRoots(originalEquation)) {
```
evykassirer: Why not just do this check at the beginning instead of cloning `originalEquation`?

Also why can't you only do it there? why do you also have to do it after other updates? Generally we only make one change at a time unless we have substeps, if it doesn't catch it this time it loops, it'll catch it the next time, right?

lib/solveEquation/stepThrough.js
```diff
@@ -107,6 +132,48 @@ function stepThrough(leftNode, rightNode, comparator, debug=false) {
   return steps;
 }
 
+/*
+  Helper function that returns the roots and symbol of an input equation
+  that has roots to be found.
+  For every factor on the left hand side, solve a new equation that is factor = 0
+  for the symbol.
+  TODO: handle multiple variable solutions e.g (x + 2) (y + 2) = 0
```
evykassirer: what happens in those cases right now? does it handle it safely or does it throw errors or return the wrong answer?

lib/solveEquation/stepThrough.js
```diff
@@ -107,6 +132,48 @@ function stepThrough(leftNode, rightNode, comparator, debug=false) {
...
+*/
+
+function getSolutionsAndSymbol (equation) {
+  // Gets all factors that aren't a constant
```
evykassirer: I don't think this comment is super descriptive - the name of the function is actually more accurate

maybe call the variable `factorsWithSymbols`?

lib/solveEquation/stepThrough.js
```diff
@@ -107,6 +132,48 @@ function stepThrough(leftNode, rightNode, comparator, debug=false) {
...
+    const allSteps = stepThrough(leftNode, rightNode, '=');
+
+    if (allSteps.length > 0) {
+      const lastStep = allSteps.slice(-1)[0];
+
+      symbol = lastStep.newEquation.leftNode;
+      // Append to a list of roots
+      solutions.push(lastStep.newEquation.rightNode);
+    }
+    else {
```
evykassirer: I really don't like that you assume 0 is a solution if there are no steps - can you explicitly check for that, and otherwise don't push anything to `solutions`, and check at the end to make sure solutions has length > 0 (and if not, return something that shows nothing was found or changed) - that'd make me feel a lot safer about all this, that we're handling missing cases

lib/solveEquation/stepThrough.js
```diff
@@ -107,6 +132,48 @@ function stepThrough(leftNode, rightNode, comparator, debug=false) {
...
+          : factors[f];
+
+    const rightNode = Node.Creator.constant(0);
+
+    const allSteps = stepThrough(leftNode, rightNode, '=');
+
+    if (allSteps.length > 0) {
+      const lastStep = allSteps.slice(-1)[0];
+
+      symbol = lastStep.newEquation.leftNode;
```
evykassirer: You still haven't addressed that this might not be a symbol. The equation code is far from perfect and finished, and so there can be steps without ending up with `symbol = solution` at the end (at least I think so - if you're pretty sure that's wrong feel free to show me the code that guarantees that!)

 I think you should explicitly check that and if it's not a symbol, don't push to `solutions`

test/solveEquation/solveEquation.test.js
```diff
@@ -44,8 +44,8 @@ describe('solveEquation for =', function () {
     ['2 - 4x = x', 'x = 2/5'],
     ['9x + 4 - 3 = 2x', 'x = -1/7'],
     ['9x + 4 - 3 = -2x', 'x = -1/11'],
-    ['(2x^2 - 1)(x^2 - 5)(x^2 + 5) = 0', '2x^6 - x^4 - 50x^2 = -25'],
-    ['(-x ^ 2 - 4x + 2)(-3x^2 - 6x + 3) = 0', '3x^4 + 18x^3 + 15x^2 - 24x = -6'],
+    //['(2x^2 - 1)(x^2 - 5)(x^2 + 5) = 0', '2x^6 - x^4 - 50x^2 = -25'],
```
evykassirer: we shouldn't merge commented out tests - why are these commented out?

lib/solveEquation/stepThrough.js
```diff
@@ -211,19 +289,26 @@ function addSimplificationSteps(steps, equation, debug=false) {
...
   const rightSubSteps = [];
   for (let i = 0; i < rightSteps.length; i++) {
     const step = rightSteps[i];
     rightSubSteps.push(EquationStatus.addRightStep(equation, step));
   }
+
+  for (let i = 0; i < rightFactorSteps.length; i++) {
+    const step = rightFactorSteps[i];
+    rightSubSteps.push(EquationStatus.addRightStep(equation, step));
+  }
```
evykassirer: I take it back - this is really weird

essentially you're collecting steps for `simplifyExpressionNode` on an equation, and then collecting steps for calling `factor` on the same original equation and then combining the steps together. You should only take steps from one of them - I believe I commented on this earlier too.

lib/solveEquation/stepThrough.js
evykassirer: I'd like to see more tests showing things we don't support, and how they're handled right now (show that it doesn't change or whatever)

just cause there's a lot of that for factoring

lib/solveEquation/stepThrough.js
```diff
@@ -155,6 +132,48 @@ function stepThrough(leftNode, rightNode, comparator, debug=false) {
...
+
+  // For each factor, solve the equation factor = 0 for the symbol
+  for (var f in factors) {
+    const leftNode = Node.Type.isParenthesis(factors[f])
+          ? factors[f].content
+          : factors[f];
+
+    const rightNode = Node.Creator.constant(0);
+
+    const allSteps = stepThrough(leftNode, rightNode, '=');
```
evykassirer: curious - why do you call this allSteps instead of steps?

lib/checks/canFindRoots.js
aliang8: I mean when it's identifier = 0, there is no change to be made.

lib/checks/canFindRoots.js
aliang8: Yup, will add that in this PR.

lib/solveEquation/stepThrough.js
aliang8: You got it.

lib/checks/canFindRoots.js
evykassirer: maybe don't add it to the examples for this function then?

lib/checks/canFindRoots.js
evykassirer: ✨ ✨ ✨ ✨ ✨ ✨

ldworkin:
No worries if review takes a long time, especially for something as complicated as this. I'll read this over carefully today, and then maybe @aliang8 we could pair on cleaning it up.

lib/solveEquation/stepThrough.js
aliang8: `simplifyingExpression` is separate from `factor`

test/solveEquation/solveEquation.test.js
aliang8: We can either find roots for these or expand them out. But the current system is confused as to which one to do.

ldworkin: hmm I feel like we should find the roots? tricky

ldworkin: what is a student most likely to be looking for? I'd imagine the values of x

lib/checks/canFindRoots.js
```diff
@@ -0,0 +1,39 @@
+const Node = require('../node');
+const resolvesToConstant = require('./resolvesToConstant.js');
+
+function canFindRoots(equation) {
```
ldworkin: Can you move your comment below to the top of the function instead, as a docstring? Make it be like "Returns true if the equation is of the form ..."

lib/checks/canFindRoots.js
```diff
@@ -0,0 +1,39 @@
...
+        && parseFloat(right.value) === 0;
+
+  const isMulOrPower = Node.Type.isOperator(left, '*') || Node.Type.isOperator(left, '^');
+
+  if (!(zeroRightSide && isMulOrPower)) {
+    return false;
+  }
+
+  // If the left side of the equation is multiplication, filter out all the factors
+  // that do not evaluate to constants because they do not have roots. If the
```
ldworkin: This comment is confusing me -- do you mean "filter out the factors that _do_ evaluate to constants"?

lib/checks/canFindRoots.js
```diff
@@ -0,0 +1,39 @@
...
+  // resulting array is empty, there is no roots to be found. Do a similiar check
+  // for when the left side is a power node.
+  // e.g 2^7, 2, (33 + 89)
+
+  if (Node.Type.isOperator(left, '*')) {
+    const factors = left.args.filter(arg => !resolvesToConstant(arg));
+    if (factors.length === 0) {
```
ldworkin: Instead you could just `return factors.length > 1`, and then you wouldn't need the return at the bottom of the function.

test/solveEquation/solveEquation.test.js
ldworkin: @aliang8 can you do something about these tests? If they don't work anymore, can you move them to a TODO section below with the other factoring errors, and add an explanation of what happens and why?

lib/solveEquation/stepThrough.js
```diff
@@ -50,9 +51,23 @@ function stepThrough(leftNode, rightNode, comparator, debug=false) {
   const MAX_STEP_COUNT = 20;
   let iters = 0;
 
+  // Checks if there are roots in the original equation before we
```
ldworkin: This code is repeated here and below, why? Can we break it out in a helper function?

lib/solveEquation/stepThrough.js
```diff
@@ -107,6 +133,76 @@ function stepThrough(leftNode, rightNode, comparator, debug=false) {
...
+  }
+
+  // For each factor, solve the equation factor = 0 for the symbol
+  // If the factor is a power node, solve the equation base = 0 for the symbol
+  for (var f in factorsWithSymbols) {
+    const factor = factorsWithSymbols[f];
+
+    let leftNode;
+
+    if (Node.Type.isOperator(factor, '^')) {
```
ldworkin: You could clean up this section by changing the logic as follows:
1. if exponent, then factor = factor.args[0]
2. now check for parenthesis on factor
That way the parenthesis check only happens once

lib/solveEquation/stepThrough.js
```diff
@@ -107,6 +133,76 @@ function stepThrough(leftNode, rightNode, comparator, debug=false) {
   return steps;
 }
 
+/*
+  Helper function that returns the roots and symbol of an input equation
+  that has roots to be found.
+  For every factor on the left hand side, solve a new equation that is factor = 0
+  for the symbol.
```
ldworkin: Great docstring 👍

lib/solveEquation/stepThrough.js
```diff
@@ -183,11 +279,25 @@ function addSimplificationSteps(steps, equation, debug=false) {
   let oldEquation = equation.clone();
 
   const leftSteps = simplifyExpressionNode(equation.leftNode, false);
+  const simplifiedLeftNode = leftSteps.length !== 0
```
ldworkin: As per our in person convo :) can you add some comments in this part of the code, explaining what you're doing, and maybe stepping through an example?

ldworkin:
Ok, I just looked through this and got comfy with it. Left a few comments that should be easy to address. Once you fix those, hopefully @evykassirer can take one more quick look and then we can merge.

I'm adding an Asana task to come back and address the issue where it only returns one of the roots. (e.g. (x-2)(x+2) = 0 will only give x=2). That concerns me, because it's kinda giving the wrong answer, so I'll try to fix soon! But I think that can go in another PR.

lib/solveEquation/stepThrough.js
ldworkin: And if it's easy to remove some of the redundancy via a helper function, do so, otherwise not a huge deal.

lib/checks/canFindRoots.js
ldworkin: @aliang8 ^

lib/checks/canFindRoots.js
ldworkin: @aliang8 ^

lib/solveEquation/stepThrough.js
ldworkin: @aliang8 , @evykassirer says "I’m pretty sure you don’t have to make the second call, it’ll just catch next time it loops or looks for steps" -- does this makes sense to you? Can you address this?

lib/solveEquation/stepThrough.js
```diff
@@ -107,6 +136,70 @@ function stepThrough(leftNode, rightNode, comparator, debug=false) {
...
+
+      symbol = lastStep.newEquation.leftNode;
+
+      // Append to a list of roots
+      if (Node.Type.isSymbol(symbol)) {
+        solutions.push(lastStep.newEquation.rightNode);
+      }
+    }
+    else {
+      // e.g. x = 0
```
ldworkin: Will we ever be in this case? Because `canFindRoots` returns false for `x=0` (as per our other conversations), and this function only gets called if `canFindRoots` is true, right?

lib/solveEquation/stepThrough.js
ldworkin: @aliang8 has this been addressed? if it's something we don't support is it in a TODO in the test cases?

lib/solveEquation/stepThrough.js
aliang8: I do have to make a second call because if the equation isn't factored yet (e.g. `x^2 + 3x + 2 =0`), we have to first factor and then look for roots in the `do ... while` loop. If the original equation is already factored (e.g. `(x+2) (x-2) = 0`), we don't have to go through any of the simplification steps and find the roots immediately.

lib/checks/canFindRoots.js
```diff
@@ -0,0 +1,35 @@
...
+const resolvesToConstant = require('./resolvesToConstant.js');
+
+function canFindRoots(equation) {
+
+  // Equation should be in the form of factor * factor = 0, factor^power = 0, or identifier = 0
+  // e.g (x - 2)^2 = 0, x(x + 2)(x - 2) = 0
+  const left = equation.leftNode;
+  const right = equation.rightNode;
+
+  const zeroRightSide = Node.Type.isConstant(right)
```
ldworkin: @aliang8 I think @evykassirer wanted some sort of helper function here, do you remember?

lib/solveEquation/stepThrough.js
ldworkin: @evykassirer does that make sense to you? If that's the case, @aliang8 can you make it a helper function?

lib/solveEquation/stepThrough.js
aliang8: Yup, this will be called. The condition for that code block is when the number of steps when factoring is === 0, namely when you have symbol = 0.

lib/checks/canFindRoots.js
aliang8: Ahhh yes! I tried adding that helper, but ran into some issues. I added an extra param for validating a constant, but it seems like somewhere in mathsteps we are already calling `Node.Type.isConstant` with an extra param. Haven't fully investigated, but Evy said it would be fine if we push that fix into a separate PR.

ldworkin: Sounds good!

lib/solveEquation/stepThrough.js
ldworkin: Ah got it!

lib/solveEquation/stepThrough.js
```diff
@@ -50,9 +51,34 @@ function stepThrough(leftNode, rightNode, comparator, debug=false) {
...
+  // Checks if there are roots in the original equation before we
+  // do any simplification.
+  // E.g. (33 + 89) (x - 99) = 0
+  if (checks.canFindRoots(equation)) {
+    const [solutions, symbol] = getSolutionsAndSymbol(equation);
+
+    let allSolutions;
+
+    if (solutions.length > 1) {
+      const flatten = [];
```
evykassirer: `flattenedSolutionList` is more descriptive :)

lib/solveEquation/stepThrough.js
```diff
@@ -107,6 +157,69 @@ function stepThrough(leftNode, rightNode, comparator, debug=false) {
...
+function getSolutionsAndSymbol (equation) {
+  const left = equation.leftNode;
+  const rightNode = Node.Creator.constant(0);
+
+  const solutions = [];
+
+  let symbol, steps, factorsWithSymbols;
+
+  // If left hand side is a power node and it does not resolve to a constant (e.g. 2^7)
+  // then it is a factor. If left hand side is a multiplication node, return a list of all the valid factors.
```
evykassirer: small thing, but can you make these lines max 80 chars?

and `(e.g. 2^7 resolves to a constant so it is not a factor, but x^2 would have factors x=0)` would be more clear, as I was reading this I thought you meant 2^7 is a power node that doesn't resolve to a constant which confused me

lib/solveEquation/stepThrough.js
```diff
@@ -107,6 +157,69 @@ function stepThrough(leftNode, rightNode, comparator, debug=false) {
...
+  Helper function that returns the roots and symbol of an input equation
+  that has roots to be found.
+  For every factor on the left hand side, solve a new equation that is factor = 0
+  for the symbol.
+  TODO: handle multiple variable solutions e.g (x + 2) (y + 2) = 0
+*/
+
+function getSolutionsAndSymbol (equation) {
+  const left = equation.leftNode;
+  const rightNode = Node.Creator.constant(0);
```
evykassirer: why do you set `rightNode = Node.Creator.constant(0);`? Isn't `rightNode` always 0 here? And if it wasn't, why would you make this variable 0 then?

lib/solveEquation/stepThrough.js
```diff
@@ -107,6 +157,69 @@ function stepThrough(leftNode, rightNode, comparator, debug=false) {
...
+  // then it is a factor. If left hand side is a multiplication node, return a list of all the valid factors.
+  if (Node.Type.isOperator(left, '^') && !checks.resolvesToConstant(left)){
+    factorsWithSymbols = [left];
+  }
+  else {
+    factorsWithSymbols = equation.leftNode.args.filter(arg => !checks.resolvesToConstant(arg));
+  }
+
+  // For each factor, solve the equation factor = 0 for the symbol
+  // If the factor is a power node, solve the equation base = 0 for the symbol
```
evykassirer: this stuff is pretty complex here but your comments are really helping - thanks!

lib/solveEquation/stepThrough.js
evykassirer: edit: I came back and kinda see what you're doing - but can you define this right before you use it instead of here?

lib/solveEquation/stepThrough.js
```diff
@@ -107,6 +157,69 @@ function stepThrough(leftNode, rightNode, comparator, debug=false) {
...
+/*
+  Helper function that returns the roots and symbol of an input equation
+  that has roots to be found.
+  For every factor on the left hand side, solve a new equation that is factor = 0
+  for the symbol.
+  TODO: handle multiple variable solutions e.g (x + 2) (y + 2) = 0
+*/
+
+function getSolutionsAndSymbol (equation) {
+  const left = equation.leftNode;
```
evykassirer: `leftNode`

lib/solveEquation/stepThrough.js
```diff
@@ -107,6 +157,69 @@ function stepThrough(leftNode, rightNode, comparator, debug=false) {
...
+    }
+
+    const leftNode = Node.Type.isParenthesis(factor)
+      ? factor.content
+      : factor;
+
+    steps = stepThrough(leftNode, rightNode, '=');
+
+    if (steps.length > 0) {
+      // e.g. x - 2 = 0
```
evykassirer: I don't think this comment/example helps explain what you're doing here (if you keep it, explain why it's relevant)

lib/solveEquation/stepThrough.js
```diff
@@ -107,6 +157,69 @@ function stepThrough(leftNode, rightNode, comparator, debug=false) {
...
+      const lastStep = steps.slice(-1)[0];
+
+      symbol = lastStep.newEquation.leftNode;
```
evykassirer: I should stress again that it's possible that the equation you're recursing on doesn't end with a solution, since our equation code is incomplete

(right? or maybe I'm missing something and you made sure earlier that it's in exactly a form we can solve)

so you'd want to check it's in this form instead of assuming

lib/solveEquation/stepThrough.js
evykassirer: this is really small and not required, but based on Lili's comments and the way I read it too, it might be easier to read if you do 

`if (steps.length === 0)` and put this code first, and then do the !== 0 case as the else afterwards, so that the if/else are closer (totally optional)

lib/solveEquation/stepThrough.js
```diff
@@ -182,12 +295,35 @@ function step(equation, symbolName) {
...
   let oldEquation = equation.clone();
 
+  /*
+    Collects substeps for factoring and simplifying the expression separately
+    because their logics are located in different places. Collect factoring
+    substeps first in case the original expression is factorable. If no factoring
+    substeps, proceed to collect simplifying substeps.
+    Some examples of when this comes into play:
+    e.g. x * x + 3x + 2 = 0 <- factor the left side
+    e.g. x + 4 + 2 = 0 <- simplify the left side
```
evykassirer: this docstring is wonderful

lib/solveEquation/stepThrough.js
```diff
@@ -182,12 +295,35 @@ function step(equation, symbolName) {
...
   const leftSubSteps = [];
-  for (let i = 0; i < leftSteps.length; i++) {
-    const step = leftSteps[i];
+
+  for (let i = 0; i < leftFactorSteps.length; i++) {
+    const step = leftFactorSteps[i];
     leftSubSteps.push(EquationStatus.addLeftStep(equation, step));
   }
+
+  if (leftSubSteps.length === 0) {
```
evykassirer: I'd move this down below - you don't want to mix left simplify steps and right factor steps

maybe even pull it out into helper functions, but try all of one first, then if it works return

if it doesn't get any steps, then try all of the other one (the code for them shouldn't be interspersed)

aliang8: How am I mixing left simplify steps and right factor steps? I'm confused on that part.

lib/solveEquation/stepThrough.js
```diff
@@ -50,9 +51,34 @@ function stepThrough(leftNode, rightNode, comparator, debug=false) {
...
+
+    let allSolutions;
+
+    if (solutions.length > 1) {
+      const flatten = [];
+      solutions.forEach(s => s.items ? flatten.push(...s.items) : flatten.push(s));
+      allSolutions =  Node.Creator.list(flatten);
+    }
+    else {
+      allSolutions = solutions[0];
```
aliang8: When will this case occur? There is guaranteed a solution because there is a check (`canFindRoots`).

lib/solveEquation/stepThrough.js
```diff
@@ -182,12 +297,35 @@ function step(equation, symbolName) {
...
+    Collects substeps for factoring and simplifying the expression separately
+    because their logics are located in different places. Collect factoring
+    substeps first in case the original expression is factorable. If no factoring
+    substeps, proceed to collect simplifying substeps.
+    Some examples of when this comes into play:
+    e.g. x * x + 3x + 2 = 0 <- factor the left side
+    e.g. x + 4 + 2 = 0 <- simplify the left side
+  */
   const leftSteps = simplifyExpressionNode(equation.leftNode, false);
+  const simplifiedLeftNode = leftSteps.length !== 0
```
ldworkin: Here you call `factor` on `simplifiedLeftNode` instead of `equation.leftNode`, but in the right hand side section, you call `factor` on `equation.RightNode`. Is there a reason for that difference? This is why I think there should be a helper function!!

lib/solveEquation/stepThrough.js
```diff
@@ -182,12 +297,35 @@ function step(equation, symbolName) {
...
+  /*
+    Collects substeps for factoring and simplifying the expression separately
+    because their logics are located in different places. Collect factoring
+    substeps first in case the original expression is factorable. If no factoring
+    substeps, proceed to collect simplifying substeps.
+    Some examples of when this comes into play:
+    e.g. x * x + 3x + 2 = 0 <- factor the left side
+    e.g. x + 4 + 2 = 0 <- simplify the left side
+  */
   const leftSteps = simplifyExpressionNode(equation.leftNode, false);
```
ldworkin: For clarity, can you call this `leftSimplifySteps` (and same for the right hand part)

lib/solveEquation/stepThrough.js
ldworkin: Hey @evykassirer , discussed this with Anthony. We're not totally sure we understand your comment, but here's what we came up with.
1) On the left side, we should first simplify, and add all those simplify substeps to a list of `leftSteps.`
2) Then we should factor the simplified equation, and add all those factoring substeps to  the `leftSteps` list.
3) On the right side, there should be no need to factor, because we always move everything other to the left side first (@aliang8 is going to double check this)

evykassirer: ah, yeah I think that makes sense. add some comments to make that structure really clear?

lib/solveEquation/stepThrough.js
evykassirer: I'm pretty sure canFindRoots won't always tell the truth - e.g. 2^x will pass those qualifications but won't have roots, right? And what about some complicate thing that doesn't resolve to constant but still we can't solve it? unless we can solve everything, that'll show up right?

evykassirer: Maybe also rename that function to be more clear it doesn't always mean we'll be able to find roots. Just think through all the cases we can't factor yet and either catch them there (which seems hard) or just gracefully handle when we don't end up finding factors (which I think would be easier)

lib/solveEquation/stepThrough.js
```diff
@@ -158,64 +126,110 @@ function stepThrough(leftNode, rightNode, comparator, debug=false) {
...
-    steps = stepThrough(leftNode, rightNode, '=');
+      steps = stepThrough(leftNode, rightNode, '=');
 
-    if (steps.length > 0) {
-      // e.g. x - 2 = 0
-      const lastStep = steps.slice(-1)[0];
+      if (steps.length === 0 && Node.Type.isSymbol(leftNode)) {
+        // e.g. x = 0
+        symbol = leftNode;
+        solutions.push(...Array(exponent).fill(rightNode));
```
evykassirer: ohh I was confused by this for a while until I realized what it meant

can you add a comment about how/why pushing the solution `exponent` times?

lib/solveEquation/stepThrough.js
```diff
@@ -158,64 +126,110 @@ function stepThrough(leftNode, rightNode, comparator, debug=false) {
...
-      factor = factor.args[0];
-    }
+      const leftNode = Node.Type.isParenthesis(factor)
+            ? factor.content
+            : factor;
 
-    const leftNode = Node.Type.isParenthesis(factor)
-      ? factor.content
-      : factor;
+      const rightNode = Node.Creator.constant(0);
```
evykassirer: this will always be `equation.rightNode`, I think it's more clear to just do

`steps = stepThrough(leftNode, equation.rightNode, '=');`

test/solveEquation/solveEquation.test.js
```diff
@@ -93,7 +97,8 @@ describe('solveEquation for =', function () {
     // ['(x^2 + 2x + 1) (x^2 + 3x + 2) = 0', ''],
     // ['(2x^2 - 1)(x^2 - 5)(x^2 + 5) = 0', ''],
     // ['(-x ^ 2 - 4x + 2)(-3x^2 - 6x + 3) = 0', ''],
-
+    // ['x^2 = -2x - 1', 'x = -1'],
+    ['(x+2)^2 -x^2 = 4(x+1)', '4 = 4'],
```
evykassirer: why is this one not commented out?

test/solveEquation/solveEquation.test.js
```diff
@@ -56,33 +55,38 @@ describe('solveEquation for =', function () {
...
-    ['x * x + 12x + 36 = 0', 'x = -6'],
-    ['x * x - 2x + 1 = 0', 'x = 1'],
+    ['x * x + 12x + 36 = 0', 'x = [-6, -6]'],
+    ['x * x - 2x + 1 = 0', 'x = [1, 1]'],
+    ['0 = x^2 + 3x + 2', 'x = [-1, -2]'],
+    ['0 = x * x + 3x + 2', 'x = [-1, -2]'],
+    ['x * x + (x + x) + 1 = 0', 'x = [-1, -1]'],
+    ['0 = x * x + (x + x) + 1', 'x = [-1, -1]'],
+    ['(x^3 / x) + (3x - x) + 1 = 0', 'x = [-1, -1]'],
+    ['0 = (x^3 / x) + (3x - x) + 1', 'x = [-1, -1]'],
```
evykassirer: beautiful tests!

lib/solveEquation/stepThrough.js
```diff
@@ -107,6 +125,115 @@ function stepThrough(leftNode, rightNode, comparator, debug=false) {
...
+        }
+      }
+    }
+  }
+  catch (e) {
+    throw new Error(e);
+  }
+
+
+
```
evykassirer: way too many new lines here :p just one is good!

test/solveEquation/solveEquation.test.js
aliang8: This one works properly.

evykassirer: it's confusing that it's in the middle of a bunch of TODO tests

aliang8: Oh okay, i'll move it up.

aliang8:
Merging !

Issue #137, opened 2017-02-22
multiply fractions with parenthesis


evykassirer:
the `multiplyFractions` function doesn't handle cases like this where a fraction is in parenthesis:

- `(2 / x) * x` -> `2x / x`
- `(2 / (4x)) * 4x` -> `(2 * 4x)/4x` 

this makes it harder for things to cancel, which means that we can't get #88 working

this would be very helpful to get done soon and I don't think it's too much work 😸

evykassirer:
another possibility is just getting rid of those parenthesis in `removeUnnecessaryParentheses` because I don't think they're needed around a fraction

ldworkin:
Fixed by https://github.com/socraticorg/mathsteps/commit/344d15b5d5e25251cef7d6976d6a73df16661d7d

PR #185, opened 2017-07-07
multiply fractions with parenthesis

ldworkin:
https://github.com/socraticorg/mathsteps/issues/137


lib/node/Type.js
```diff
@@ -96,5 +96,32 @@ NodeType.isIntegerFraction = function(node, allowUnaryMinus=false) {
           Number.isInteger(parseFloat(denominator.value)));
 };
 
+NodeType.isFraction = function(node, allowUnaryMinus=true, allowParens=true) {
```
ldworkin: Not sure if this `isFraction` / `getFraction` pattern is okay, or if there's a better way to accomplish this? I'm sure we've run into similar situations elsewhere, not sure how it is typically handled?

ldworkin:
I know in the issue you suggested that one potential fix was to modify `removeUnnecessaryParentheses`, but that made me nervous ...

aliang8:
This looks good to me.

lib/node/Type.js
evykassirer: oo that's a good question. I can't think of anything off the top of my head, but if you come across something later we could generalize them together. I think this is fine for now

I could imagine something general that takes in a function that returns true for the base condition (like the `isOperator('/')` here) and does the unaryMinus/Parenthesis checking more generally no matter what the base case is

... does that make sense? We could wait until it comes up to generalize though

lib/simplifyExpression/multiplyFractionsSearch/index.js
```diff
@@ -23,7 +23,7 @@ function multiplyFractions(node) {
     return Node.Status.noChange(node);
   }
   const atLeastOneFraction = node.args.some(
-    arg => Node.Type.isOperator(arg, '/'));
+    arg => Node.Type.isFraction(arg));
```
evykassirer: oh I now know how to make this cleaner (if only past me knew!)

`node.args.some(Node.Type.isFraction)`

evykassirer: also LOVE this hack (it's not that hacky imo, it's just not an 'in depth' fix)

lib/simplifyExpression/multiplyFractionsSearch/index.js
```diff
@@ -33,9 +33,10 @@ function multiplyFractions(node) {
   const numeratorArgs = [];
   const denominatorArgs = [];
   node.args.forEach(operand => {
-    if (Node.Type.isOperator(operand, '/')) {
-      numeratorArgs.push(operand.args[0]);
-      denominatorArgs.push(operand.args[1]);
+    if (Node.Type.isFraction(operand)) {
+      const fraction = Node.Type.getFraction(operand);
```
evykassirer: the getFraction will remove the unaryminus which breaks stuff - you probably want to move that unary minus into the numerator (I feel like there's code somewhere that does this already but I don't remember where 

maybe in flatten? It'd be nice to have it there to always turn -(a/b) into -a / b then you won't have to deal with it here 

what do you think?

test/simplifyExpression/multiplyFractionsSearch/multiplyFractionsSearch.test.js
```diff
@@ -10,7 +10,9 @@ describe('multiplyFractions', function () {
   const tests = [
     ['3 * 1/5 * 5/9', '(3 * 1 * 5) / (5 * 9)'],
     ['3/7 * 10/11', '(3 * 10) / (7 * 11)'],
-    ['2 * 5/x', '(2 * 5) / x']
+    ['2 * 5/x', '(2 * 5) / x'],
+    ['2 * (5/x)', '(2 * 5) / x'],
+    ['(5/x) * (2/x)', '(5 * 2) / (x * x)'],
```
evykassirer: add some tests with unary minus

lib/node/Type.js
ldworkin: I liked this idea a lot so went ahead and did it now!

lib/simplifyExpression/multiplyFractionsSearch/index.js
ldworkin: That is cleaner, cool! Fixed

lib/simplifyExpression/multiplyFractionsSearch/index.js
ldworkin: Omg good catch! * facepalm *

I modified the`isFraction`/`getFraction` functions to _not_ allow unary minus for now, since now that you've pointed this out, it seems dangerous.

It looks like a simple case like -1/2 works fine, because minus does go on the numerator. The case that doesn't work is -(1/2). I thought for a while but couldn't think of a clean way to handle this right now, so I'm punting for the moment. Added a TODO comment in the tests. Let me know if you have any ideas though.

test/simplifyExpression/multiplyFractionsSearch/multiplyFractionsSearch.test.js
ldworkin: Yup good call, added (except for the case I don't support yet, re: comment above)

lib/simplifyExpression/multiplyFractionsSearch/index.js
ldworkin: ACTUALLY this only works if the function you're passing to `some` only takes one argument. `some` actually passes in 3 arguments to the function you give it: the element, the index, and the array. If your function expects just one argument, the element, that's fine. But if your function can take other optional arguments (like isFraction can take allowParens) then it gets confused because it thinks the index = allowParens. Ugh this is hard to explain but basically it turns out you have to be careful if you use that shortcut!

ldworkin: more info here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some

evykassirer: oh cool! good to know :D learned something new about javascript today ~~ thanks!

lib/simplifyExpression/multiplyFractionsSearch/index.js
evykassirer: can you do, for the case with unary minus, call getFraction recursively then negate the numerator before returning? I think that should work - how's that sound?

lib/node/Type.js
```diff
@@ -96,32 +96,44 @@ NodeType.isIntegerFraction = function(node, allowUnaryMinus=false) {
           Number.isInteger(parseFloat(denominator.value)));
 };
 
-NodeType.isFraction = function(node, allowUnaryMinus=true, allowParens=true) {
-  if (NodeType.isOperator(node, '/')) {
+// Returns true if `node` belongs to the type specified by boolean `isTypeFunc`.
+// If `allowParens` is true, we allow for the node to be nested within parenthesis.
```
evykassirer: love it! I think you could add the unaryminus case back in if what I said works

lib/node/Type.js
```diff
@@ -96,32 +96,44 @@ NodeType.isIntegerFraction = function(node, allowUnaryMinus=false) {
...
   }
 
-  throw Error('Expected fraction');
+  throw Error('`getType` called on a node that does not belong to specified type');
+};
+
+NodeType.isFraction = function(node, allowParens=true) {
+  return NodeType.isType(
+    node,
+    function(node) { return NodeType.isOperator(node, '/'); },
```
evykassirer: in es6 I think this is generally done `(node) => NodeType.isOperator(node, '/')`

lib/simplifyExpression/multiplyFractionsSearch/index.js
```diff
@@ -22,10 +22,12 @@ function multiplyFractions(node) {
   if (!Node.Type.isOperator(node) || node.op !== '*') {
     return Node.Status.noChange(node);
   }
+
+  // we need to use the verbose syntax for `some` here because isFraction
+  // can take more than one parameter
```
evykassirer: omg I love that you added this comment - I was just thinking about how someone might come back to this later and change it, so good call

test/Node/Type.test.js
```diff
@@ -108,3 +109,38 @@ describe('isIntegerFraction', function () {
       false);
   });
 });
+
+describe('isFraction', function () {
```
evykassirer: tests 😍

test/TestUtil.js
```diff
@@ -53,4 +53,11 @@ TestUtil.testSubsteps = function (fn, exprString, outputList,
   });
 };
 
+// Remove some property used in mathjs that we don't need and prevents node
+// equality checks from passing
+TestUtil.removeComments = function(node) {
+  node.filter(node => node.comment !== undefined).forEach(
+    node => delete node.comment);
+};
```
evykassirer: oh interesting is this a new mathjs thing? I think there's something else we remove somewhere too (unless this is it)

evykassirer: oh it is the same thing! (she says after reading further)

I like it here, thanks!! ❤️

evykassirer:
looks good!! I think you can add the unary minus back in :)

lib/simplifyExpression/multiplyFractionsSearch/index.js
ldworkin: SMART -- love it, will try that out

ldworkin: Okay so I did this (https://github.com/socraticorg/mathsteps/pull/185/commits/47e5f33eb2d86c4414246dcc8000338588aaf052) , but it was a bit more complicated than expected 😂  First there was an import cycle (NodeType -> Negative -> NodeType) so I had to move some stuff out of NodeType into a new NodeHelper file. (Open to a better name if you have suggestions). And then I had to do something a little wonky with a `moveUnaryMinus` function parameter to get the unary minus stuff to work for the generalized isType/getType functions. Anyway let me know what you think, haha.

evykassirer: omg import cycles annoy me so much when working on this project haha T_T

lib/node/Helper.js
```diff
@@ -0,0 +1,78 @@
+const Negative = require('../Negative');
+const NodeCreator = require('./Creator');
+const NodeType = require('./Type');
+
+const NodeHelper = {};
```
evykassirer: hmm trying to think of a more specific name than "helper" since that could be anything

NodeGeneralType? or maybe NodeCustomType?

lib/node/Helper.js
```diff
@@ -0,0 +1,78 @@
...
+  node, isTypeFunc, allowUnaryMinus=true, allowParens=true, moveUnaryMinus=undefined) {
+  if (allowUnaryMinus === true && moveUnaryMinus === undefined) {
+    throw Error('Error in `getType`: moveUnaryMinus is undefined');
+  }
+
+  if (isTypeFunc(node)) {
+    return node;
+  }
+  else if (allowUnaryMinus && NodeType.isUnaryMinus(node)) {
+    return moveUnaryMinus(
```
evykassirer: Oooo right cause you have to define this - interesting! I wonder if we could default this to something - will be interesting to see if/how we find other cases for this. May as well leave as is for now

lib/node/Helper.js
```diff
@@ -0,0 +1,78 @@
...
+      NodeHelper.getType(
+        node.args[0], isTypeFunc, allowUnaryMinus, allowParens, moveUnaryMinus));
+  }
+  else if (allowParens && NodeType.isParenthesis(node)) {
+    return NodeHelper.getType(
+      node.content, isTypeFunc, allowUnaryMinus, allowParens, moveUnaryMinus);
+  }
+
+  throw Error('`getType` called on a node that does not belong to specified type');
```
evykassirer: +++

test/Node/Type.test.js
```diff
@@ -113,34 +114,52 @@ describe('isIntegerFraction', function () {
...
   });
   it('-2/3 true', function () {
     assert.deepEqual(
-      Node.Type.isFraction(math.parse('-2/3')),
+      Node.Helper.isFraction(math.parse('-2/3')),
+      true);
+  });
+  it('-(2/3) true', function () {
+    assert.deepEqual(
+      Node.Helper.isFraction(math.parse('-(2/3)')),
```
evykassirer: 🎉

test/Node/Type.test.js
```diff
@@ -113,34 +114,52 @@ describe('isIntegerFraction', function () {
 describe('isFraction', function () {
   it('2/3 true', function () {
     assert.deepEqual(
-      Node.Type.isFraction(math.parse('2/3')),
+      Node.Helper.isFraction(math.parse('2/3')),
```
evykassirer: you repeat using this `assert.deepEqual(Node.Helper.isFraction` thing a lot - you could generalize that out into a test function in this file and only pass in the function string and true/false (a lot of other test files have patterns like this)

evykassirer:
You can clean up the test structure if you want, and I think a more descriptive name would be good - but yeah looks awesome! Messy stuff and weird things to work with, but you handled it all pretty cleanly 🎉

Issue #139, opened 2017-02-22
multiply fractions edge case to support cancelling


evykassirer:
there are some cases where we don't want to multiply fractions with other things, one of them is ` 9/2 * x  -> 9x / 2` (since this doesn't change anything and could be considered less readable)

a case we're not handling with the current logic, that *should* multiply is:

`x * 6/x -> (x * 6) / x` (which is needed for cancelling out)

one solution is just to have better cancelling code that doesn't only work with the fraction `6x/x`, and can recognize that `x` cancels with the denominator in `6/x`

another (possibly easier for now) solution would be to add a case for multiplying a symbol times a term with a symbol in the denominator

... this is a bit messy, so I'm not sure what the best solution is - open to suggestions!

this came up because of a case in #88

kevinbarabash:
my vote would be for better canceling code.

aelnaiem:
I agree with @kevinbarabash, let's improve the canceling code so it handles this case

evykassirer:
I'd love for that to happen :) just haven't had the time to figure out the best way for that to work

ldworkin:
Fixed by https://github.com/socraticorg/mathsteps/commit/344d15b5d5e25251cef7d6976d6a73df16661d7d

PR #186, opened 2017-07-07
multiply fractions with symbol in denominator

ldworkin:
https://github.com/socraticorg/mathsteps/issues/139


ldworkin:
I fixed this issue the cheating way, sorry! Refactoring how cancelling works will have to happen another time :)

lib/simplifyExpression/multiplyFractionsSearch/index.js
```diff
@@ -54,4 +56,19 @@ function multiplyFractions(node) {
...
+function hasConstantNumeratorAndPolynomialInDenominator(node) {
+  if (!(Node.Type.isFraction(node))) {
+    return false;
+  }
+
+  const fraction = Node.Type.getFraction(node);
+  const numerator = fraction.args[0];
+  const denominator = fraction.args[1];
+
+  return (Node.Type.isConstant(numerator) &&
```
ldworkin: Do we need to ensure that the numerator is a constant? I'm not sure, I'm having trouble wrapping my head around the logic. I don't think it hurts.

lib/simplifyExpression/multiplyFractionsSearch/index.js
```diff
@@ -2,22 +2,21 @@ const ChangeTypes = require('../../ChangeTypes');
 const Node = require('../../node');
 const TreeSearch = require('../../TreeSearch');
 
-// If `node` is a product of terms where some are fractions (but none are
-// polynomial terms), multiplies them together.
+// If `node` is a product of terms where:
+// 1) at least one is a fraction
+// 2) either none are polynomial terms, OR
+//    at least one has a symbol in the denominator
+// then multiply them together.
```
evykassirer: this docstring is 👌 

is it noted somewhere why we have the case of none are polynomial terms?

evykassirer: this docstring is 👌

is it noted somewhere why we have the case of none are polynomial terms?

lib/simplifyExpression/multiplyFractionsSearch/index.js
```diff
@@ -26,7 +25,10 @@ function multiplyFractions(node) {
     arg => Node.Type.isFraction(arg));
   const hasPolynomialTerms = node.args.some(
     arg => Node.PolynomialTerm.isPolynomialTerm(arg));
-  if (!atLeastOneFraction || hasPolynomialTerms) {
+  const hasConstantNumeratorAndPolynomialInDenominatorTerms = node.args.some(
+    arg => hasConstantNumeratorAndPolynomialInDenominator(arg));
```
evykassirer: hehe so long name, but descriptive! :D 

you can do 

```
node.args.some(
  hasConstantNumeratorAndPolynomialInDenominator)
```

lib/simplifyExpression/multiplyFractionsSearch/index.js
evykassirer: what cases are we trying to catch? it's for cancelling eventually right?

so if we had `x * 2y / x` we'd still want to multiply fractions so we can cancel. So I think only letting it through if the numerator is a constant will leave some cases uncancelled!

lib/simplifyExpression/multiplyFractionsSearch/index.js
ldworkin: No, I don't think so, and I'm actually not positive why this is the case! Is it because of your comment on the issue here: "there are some cases where we don't want to multiply fractions with other things, one of them is 9/2 * x -> 9x / 2 (since this doesn't change anything and could be considered less readable)" ?

lib/simplifyExpression/multiplyFractionsSearch/index.js
ldworkin: Serdar taught me to write long descriptive names like that, hehe. I was hesitant at first but now I like it!

ldworkin:
Good call re: constant in numerator -- I removed that constraint! I don't totally remember why I had it in the first place 😆

test/simplifyExpression/simplify.test.js
```diff
@@ -46,7 +46,7 @@ describe('can simplify with division', function () {
     ['2x * 5x / 2', '5x^2'],
     ['2x * 4x / 5 * 10 + 3', '16x^2 + 3'],
     ['2x * 4x / 2 / 4', 'x^2'],
-    ['2x * y / z * 10', '20x * y / z'],
+    ['2x * y / z * 10', '(20x * y) / z'],
```
ldworkin: This test case changed slightly because now this case is handled by `multiplyFractionSearch` ... think that's fine?

lib/simplifyExpression/multiplyFractionsSearch/index.js
evykassirer: Hm yeah I think that's it. I feel like that case could be better handled - cause right now 5/2 * x * y/5 won't be combined here... but I guess that'd be handled in collect and combine? Just wondering if we're leaving out anything else with this clause

But yeah adding that as a comment would be helpful for people seeing this later and wondering why, like we both just did haha

evykassirer:
just add that comment explaining what's up 

then 🚢  :D

PR #187, opened 2017-07-07
Foil with exponent fix

aliang8:
fixes: https://s3.amazonaws.com/socratic-ocr-uploads/prod/1ecfffa5-4f44-4fe0-92cb-04f7ff27cdf5/processed.jpeg


lib/checks/canDistribute.js
```diff
@@ -0,0 +1,24 @@
...
+
+  const checks = Node.Type.isOperator(node)
+        && node.op === '^'
+        && Node.Type.isParenthesis(node.args[0])
+        && Node.Type.isConstant(node.args[1]);
+
+  if (!checks) {
+    return false;
+  }
+  else {
```
ldworkin: This doesn't need to be in an `else` statement, because you returned early in the `if`.

lib/checks/canDistribute.js
```diff
@@ -0,0 +1,24 @@
+const Node = require('../node');
+
+function canDistribute(node) {
```
ldworkin: Not sure if this is the best name for this function/file. I'd be in favor of something more descriptive/accurate even if it's long, like `isBinomialSquared` or something? Wondering what Evy thinks.

lib/checks/canDistribute.js
```diff
@@ -0,0 +1,24 @@
...
+  // For the special case when the node is in the form (x - 2)^2
+  // First check if is power node and the exponent is a constant
+  // Then check if the base is a polynomial
+
+  const checks = Node.Type.isOperator(node)
+        && node.op === '^'
+        && Node.Type.isParenthesis(node.args[0])
```
ldworkin: Do we have to check that the thing inside the parenthesis is the sum or difference of exactly two terms?

test/solveEquation/solveEquation.test.js
```diff
@@ -114,17 +114,3 @@ describe('constant comparison support', function () {
   ];
   tests.forEach(t => testSolveConstantEquation(t[0], t[1], t[2]));
 });
-
-function testEquationError(equationString, debug=false) {
```
ldworkin: Why'd you get rid of this stuff?

lib/checks/canDistribute.js
```diff
@@ -0,0 +1,24 @@
...
+        && Node.Type.isConstant(node.args[1]);
```
ldworkin: Do we need to check that the exponent is a 2? What happens if it's cubed or something -- will that break later?

lib/simplifyExpression/distributeSearch/index.js
```diff
@@ -18,6 +19,20 @@ function distribute(node) {
...
+    const [base, exponent] = node.args;
+
+    let newNode;
+    const times = parseFloat(exponent.value);
+
+    if (times > 0) {
+      newNode = Node.Creator.operator('*', Array(times).fill(base));
+      return distributeAndSimplifyOperationNode(newNode);
+    }
+    else {
```
ldworkin: Like before, if you return in the `if` statement, you don't need to have an `else` statement.

test/simplifyExpression/distributeSearch/distributeSearch.test.js
```diff
@@ -47,6 +47,16 @@ describe('distribute', function () {
       ['(-2x^2 * 3x - 2x^2 * -4)',
         '(-6x^3 + 8x^2)']
     ],
+    ['(x - 2)^2',
+      ['(x * (x - 2) - 2 * (x - 2))',
```
ldworkin: Ideally it might be nice if we had an intermediate step that was `(x-2)(x-2)` ... not sure if that's strictly necessary though!

tkosan: @ldworkin the special expansion `(a - b)^2 = a^2 - 2ab + b^2` is usually used in this situation. So `(x - 2)^2` would be rewritten to `x^2 -4x + 4`. ([Source](http://www.mathexpression.com/basic-algebra-formulas.html))

test/solveEquation/solveEquation.test.js
aliang8: It's not an error anymore, we can "solve" the query now.

lib/checks/canDistribute.js
aliang8: hmm I was think that. I will change it.

test/simplifyExpression/distributeSearch/distributeSearch.test.js
evykassirer: agree that having `(x-2)(x-2)` would be good if we go to `(x * (x - 2) - 2 * (x - 2))`

however we could also use the rule based approach Ted's suggesting where we go straight to `x^2 -4x + 4` - I'd want to reference that special case rule though if we did that. And I tend to shy away from those approaches because it's more memorization based. Ideally we'd have `(x-2)(x-2)` and `(x * (x - 2) - 2 * (x - 2))` in substeps, imo at least

lib/checks/canDistribute.js
evykassirer: something like `isBinomialSquared` would be much better yes! thanks Lili

`canDistribute` is a lot more cases than the one this function does and would be confusing

I think there's a name for this rule where there's the minus sign - ideally it'd be called that, cause otherwise it sounds like (x+2)^2 is handled here too 

having trouble thinking of what exactly to call it

lib/checks/canDistribute.js
```diff
@@ -0,0 +1,24 @@
...
+
+  if (!checks) {
```
evykassirer: you could just do if( ..... ) right? why define checks first?

lib/checks/canDistribute.js
evykassirer: if we're also considering sums of two terms, you should update the docstring to show that

lib/checks/canDistribute.js
```diff
@@ -0,0 +1,24 @@
+const Node = require('../node');
+
+function canDistribute(node) {
+  // For the special case when the node is in the form (x - 2)^2
+  // First check if is power node and the exponent is a constant
+  // Then check if the base is a polynomial
+
+  const checks = Node.Type.isOperator(node)
+        && node.op === '^'
```
evykassirer: you can do `Node.Type.isOperator(node, '^')`

lib/checks/canDistribute.js
```diff
@@ -0,0 +1,24 @@
...
+        && Node.Type.isParenthesis(node.args[0])
+        && Node.Type.isConstant(node.args[1]);
+
+  if (!checks) {
+    return false;
+  }
+  else {
+    const base = node.args[0].content;
+    const isPolynomial = base.args.some(Node.PolynomialTerm.isPolynomialTerm);
```
evykassirer: can it be any polynomial? or does it have to be a binomial (this would affect your name above - then readers don't have to guess at the scope of what it does!)

lib/simplifyExpression/distributeSearch/index.js
```diff
@@ -18,6 +19,20 @@ function distribute(node) {
...
     return distributeUnaryMinus(node);
   }
+  else if (checks.canDistribute(node)) {
+    const [base, exponent] = node.args;
+
+    let newNode;
+    const times = parseFloat(exponent.value);
+
+    if (times > 0) {
+      newNode = Node.Creator.operator('*', Array(times).fill(base));
```
evykassirer: beautiful fix! more teaching, little code - love it

test/solveEquation/solveEquation.test.js
evykassirer: maybe add another equation we can't solve, to make sure it's still throwing instead of doing weird stuff? or is there nothing left that falls into that category (where it tries to solve for a while and does weird stuff but doesn't actually solve)

test/solveEquation/solveEquation.test.js
```diff
@@ -114,17 +114,3 @@ describe('constant comparison support', function () {
...
-  it(equationString + ' throws error', (done) => {
-    assert.throws(() => solveEquation(equationString, debug),Error);
-    done();
-  });
-}
-
-describe('solveEquation errors', function() {
-  const tests = [
-    ['( x + 2) ^ ( 2) - x ^ ( 2) = 4( x + 1)']
```
evykassirer: if we can do this now, wanna move it into tests elsewhere? :D

lib/checks/canDistribute.js
evykassirer: I don't think you need that actually, now that I've read through the whole diff - you can expand out the exponentiation of any polynomial, and distribute. If that counts as "simplifying" is a whole other question though

lib/checks/canDistribute.js
aliang8: any polynomial

lib/checks/isPolynomialToConstantPower.js
```diff
@@ -0,0 +1,19 @@
...
+  // First check if is power node and the exponent is a constant
+  // Then check if the base is a polynomial
+
+  if (!(Node.Type.isOperator(node, '^')
+        && Node.Type.isParenthesis(node.args[0])
+        && Node.Type.isConstant(node.args[1]))) {
+    return false;
+  }
+  const base = node.args[0].content;
+  const isPolynomial = base.args.some(Node.PolynomialTerm.isPolynomialTerm);
```
ldworkin: Will this let too many things through? This would include things like: 5*sqrt(4)*x and y/(5+x) and all sorts of things that aren't a sum/difference of polynomial and/or constant terms. Maybe that's fine, I'm not sure? To be safer, we could check that all args are either polynomials or constants, and that the root operator is +.

test/solveEquation/solveEquation.test.js
```diff
@@ -124,7 +124,6 @@ function testEquationError(equationString, debug=false) {
 
 describe('solveEquation errors', function() {
   const tests = [
-    ['( x + 2) ^ ( 2) - x ^ ( 2) = 4( x + 1)']
```
ldworkin: Did you move this test somewhere else like Evy asked?

test/simplifyExpression/distributeSearch/distributeSearch.test.js
```diff
@@ -47,6 +47,18 @@ describe('distribute', function () {
       ['(-2x^2 * 3x - 2x^2 * -4)',
         '(-6x^3 + 8x^2)']
     ],
+    ['(x - 2)^2',
```
ldworkin: I think some more tests would be good. What if there are more than 2 terms? What if we raise to a power that is not 2?

ldworkin:
There's a test failing above ^^ but it should be easy to fix. I think the test is wrong :D

lib/checks/isPolynomialToConstantPower.js
```diff
@@ -0,0 +1,19 @@
+const Node = require('../node');
+
+function isPolynomialToConstantPower(node) {
+  // For the special case when the node is in the form (x + 2)^2
```
evykassirer: I think just to be super clear

"For the special case when the node is a polynomial to a constant power, e.g. (x+2)^2"

cause exactly (x+2)^2 is technically one special case, but you mean to catch more than that :p

lib/checks/isPolynomialToConstantPower.js
evykassirer: Yeah, I think checking for isOperator(+) would be good and also address the issue I was about to raise of "how do you know `base` has args"?

test/solveEquation/solveEquation.test.js
evykassirer: pls

aliang8:
Close this PR because I accidentally handled it in another one. I think that was a better fix. See #194

Issue #41, opened 2016-12-13
Simplify fraction should factor to show common factors being divided out


evykassirer:
e.g.  `15/6 -> (5*3)/(2*3) -> 5/2`

right now it just does  `15/6 -> 5/2`

This involves adding substeps to the existing `ConstantFraction.divideByGCD` function, where you show the gcd of the numerator and denominator being factored out (in the example above, 3 was the gcd for 15 and 6)

PR #188, opened 2017-07-07
greatest common denominator substep

ldworkin:
https://github.com/socraticorg/mathsteps/issues/41

### DIFF
--- a/lib/ChangeTypes.js
+++ b/lib/ChangeTypes.js
@@ -80,6 +80,8 @@ module.exports = {
   SIMPLIFY_FRACTION: 'SIMPLIFY_FRACTION',
   // e.g. 2/-3 -> -2/3
   SIMPLIFY_SIGNS: 'SIMPLIFY_SIGNS',
+  // e.g. 15/6 -> (5*3)/(2*3)
+  FIND_GCD: 'FIND_GCD',
 
   // ADDING FRACTIONS
 
--- a/lib/simplifyExpression/fractionsSearch/divideByGCD.js
+++ b/lib/simplifyExpression/fractionsSearch/divideByGCD.js
@@ -14,7 +14,7 @@ const Node = require('../../node');
 // isn't, we can simplify signs.
 // Returns a Node.Status object
 function divideByGCD(fraction) {
-  let substeps = [];
+  const substeps = [];
 
   if (!Node.Type.isOperator(fraction) || fraction.op !== '/') {
     return Node.Status.noChange(fraction);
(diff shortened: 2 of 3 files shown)


lib/simplifyExpression/fractionsSearch/divideByGCD.js
```diff
@@ -40,6 +42,16 @@ function divideByGCD(fraction) {
...
 
+  const intermediateNumerator = Node.Creator.operator(
+    '/', [Node.Creator.constant(numeratorValue), Node.Creator.constant(gcd)]);
+  const intermediateDenominator = Node.Creator.operator(
+    '/', [Node.Creator.constant(denominatorValue), Node.Creator.constant(gcd)]);
+  const intermediateFraction = Node.Creator.operator(
+    '/', [intermediateNumerator, intermediateDenominator]);
+  const status = Node.Status.nodeChanged(
+    ChangeTypes.FIND_GCD, fraction, intermediateFraction);
+  substeps.push(status);
```
ldworkin: Usually we call `resetChangeGroups` after this but I'm not sure if that's needed here?

lib/ChangeTypes.js
```diff
@@ -80,6 +80,8 @@ module.exports = {
   SIMPLIFY_FRACTION: 'SIMPLIFY_FRACTION',
   // e.g. 2/-3 -> -2/3
   SIMPLIFY_SIGNS: 'SIMPLIFY_SIGNS',
+  // e.g. 15/6 -> (5*3)/(2*3)
+  FIND_GCD: 'FIND_GCD',
```
evykassirer: pedagogy!! 🎉

test/simplifyExpression/fractionsSearch/divideByGCD.test.js
```diff
@@ -15,5 +19,19 @@ describe('simplifyFraction', function() {
...
     ['-1/-1', '1'],
   ];
-  tests.forEach(t => testdivideByGCD(t[0], t[1]));
+  tests.forEach(t => testDivideByGCD(t[0], t[1]));
+});
+
+describe('simplifyFraction', function() {
+  const tests = [
+    ['15/6',
+     ['15/3 / (6/3)'],
```
evykassirer: I think the numerator should have parens around it too (if you keep this version of the rule)

lib/ChangeTypes.js
evykassirer: in your test (and the code you wrote) you'll get `15/3 / (6/3)` and not `(5*3)/(2*3)`

I think the latter is cleaner (nested fractions render more funky in the app) - what do you think is better?

lib/simplifyExpression/fractionsSearch/divideByGCD.js
evykassirer: ah, cause `fraction` and `intermediateFraction` aren't at all part of the nodes used in the next step? Yeah, it's probably fine then

changeGroups are kinda messy in general heh and not well tested so I wouldn't worry too much about it for now (but thinking about them a bit when you can is still good)

lib/ChangeTypes.js
ldworkin: Ha good call -- I agree, changed!

PR #189, opened 2017-07-08
Multiply nthRoots

aliang8:
sqrt(x) * sqrt(x) should simplify to x


lib/simplifyExpression/collectAndCombineSearch/LikeTermCollector.js
```diff
@@ -220,6 +223,19 @@ function getTermsForCollectingMultiplication(node) {
 }
 
 // A helper function for getTermsForCollectingMultiplication
+// e.g. nthRoot(x, 2), append '2': nthRootNode to terms dictionary
+// Takes the terms dictionary and the nthRoot node, and returns an updated
+// terms dictionary.
+function addToTermsforNthRootMultiplication(terms, node) {
+  const nthRootNode = clone(node);
```
ldworkin: Why do you have to clone here?

lib/simplifyExpression/collectAndCombineSearch/LikeTermCollector.js
```diff
@@ -220,6 +223,19 @@ function getTermsForCollectingMultiplication(node) {
 }
 
 // A helper function for getTermsForCollectingMultiplication
+// e.g. nthRoot(x, 2), append '2': nthRootNode to terms dictionary
```
ldworkin: I'd name this something more descriptive than just '2'. Like 'nthRoot2' or something, not sure.

ldworkin:
This looks great! How hard do you think it would be to also support`sqrt(x,2)^2 => x`? We would just need to add support for expanding `sqrt(x,2)^2 => sqrt(x,2)*sqrt(x,2)` and then your code will take it from there.

ldworkin:
I was also taking a look at what we do for `sqrt(x^2,2)` -- that should really result in `abs(x)` not just `x`, right?

test/simplifyExpression/collectAndCombineSearch/collectAndCombineSearch.test.js
```diff
@@ -10,6 +10,21 @@ function testSimpleCollectAndCombineSearch(exprString, outputStr) {
   TestUtil.testSimplification(collectAndCombineSearch, exprString, outputStr);
 }
 
+describe('combineNthRoots multiplication', function() {
+  const tests = [
+    ['nthRoot(x, 2) * nthRoot(x, 2) * nthRoot(x, 3)',
```
ldworkin: Can you add a test case that is just `nthRoot(x, 2) * nthRoot(x, 2)` to make sure it works if there is no collecting, but only combining?

lib/checks/index.js
```diff
@@ -11,9 +12,11 @@ module.exports = {
   canAddLikeTermPolynomialNodes,
   canMultiplyLikeTermConstantNodes,
   canMultiplyLikeTermPolynomialNodes,
+  canMultiplyNthRoots,
```
ldworkin: Should this be `canMultiplyLikeTermNthRoots` to have consistent naming with the other files?

ldworkin:
This might not be super important, but does this work for negative square roots? (i.e. unary minus)?

lib/checks/canMultiplyNthRoots.js
```diff
@@ -0,0 +1,19 @@
+const Node = require('../node');
+
+function canMultiplyNthRoots(node) {
+  // checks if node is a multiplication of nthRoot nodes
+  // all the terms has to have the same root node to be multiplied
+
+  if (!Node.Type.isOperator(node, '*')
+      || !(node.args.every(term => term.fn && term.fn.name === 'nthRoot'))){
```
ldworkin: `NodeType.isFunction(node, 'nthRoot')` is maybe a slightly cleaner way of checking if the node is an `nthRoot`. Not super important.

lib/ChangeTypes.js
```diff
@@ -67,6 +67,8 @@ module.exports = {
   MULTIPLY_COEFFICIENTS: 'MULTIPLY_COEFFICIENTS',
   // e.g. 2x * x -> 2x ^ 2
   MULTIPLY_POLYNOMIAL_TERMS: 'MULTIPLY_POLYNOMIAL_TERMS',
+  // e.g. nthRoot(x, 2) * nthRoot(x, 2) -> nthRoot(x^2, 2)
+  MULTIPLY_NTH_ROOTS: 'MULTIPLY_NTH_ROOTS',
```
ldworkin: I think this should go in the Roots section, rather than the Multiplying Polynomials section.

aliang8:
Great suggestions. Thanks. Regarding the `abs(x)` comment, I think that is a valid point. We should make that change in another PR. I will also make the changes for `(nthRoot(x,2))^2` in a separate PR since it involves manipulating a different set of files.

lib/simplifyExpression/collectAndCombineSearch/LikeTermCollector.js
aliang8: The 2 is for the index, so I think naming it `nthRoot2` would be misleading. I think just having the `2` there would suffice? Not sure about the naming schema or this codebase.

lib/checks/canMultiplyLikeTermsNthRoots.js
```diff
@@ -0,0 +1,19 @@
+const Node = require('../node');
+
+function canMultiplyNthRoots(node) {
```
evykassirer: this should be called `canMultiplyLikeTermsNthRoots` to match the file name

evykassirer:
> something like nthRoot(x, -2) or like nthRoot(-x, 2)? Neither one of these should evaluate. Are you suggesting that I should add checks?

Yeah we should probably make sure these don't try to get evaluated and throw errors - good catch!

lib/checks/canMultiplyNthRoots.js
evykassirer: definitely cleaner! thanks

lib/checks/canMultiplyLikeTermsNthRoots.js
```diff
@@ -0,0 +1,19 @@
...
+  // checks if node is a multiplication of nthRoot nodes
+  // all the terms has to have the same root node to be multiplied
```
evykassirer: can you put docstrings above the function names? and add an example of some math that would be handled here

lib/checks/canMultiplyLikeTermsNthRoots.js
```diff
@@ -0,0 +1,19 @@
...
+
+  if (!Node.Type.isOperator(node, '*')
+      || !(node.args.every(term => Node.Type.isFunction(term, 'nthRoot')))){
```
evykassirer: hmm what happens if you have `nthRoot(x, 2) * nthRoot(x, 2) * 3` ? It won't get combined here right?

Is that handled in collect and combine? If so, add a test for this case (I don't think I see one) -
 if not maybe you can make a separate change to collect them there so they can be combined with this function

lib/checks/index.js
```diff
@@ -11,9 +12,11 @@ module.exports = {
...
   canMultiplyLikeTermConstantNodes,
   canMultiplyLikeTermPolynomialNodes,
+  canMultiplyLikeTermsNthRoots,
   canRearrangeCoefficient,
   canSimplifyPolynomialTerms,
   hasUnsupportedNodes,
   isQuadratic,
   resolvesToConstant,
 };
+
```
evykassirer: no extra new line at end of file :p

lib/simplifyExpression/collectAndCombineSearch/LikeTermCollector.js
```diff
@@ -198,6 +198,9 @@ function getTermsForCollectingMultiplication(node) {
     if (Node.PolynomialTerm.isPolynomialTerm(child)) {
       terms = addToTermsforPolynomialMultiplication(terms, child);
     }
+    else if (child.fn && child.fn.name === 'nthRoot') {
```
evykassirer: `Node.Type.isFunction`

lib/simplifyExpression/collectAndCombineSearch/LikeTermCollector.js
evykassirer: so the '2' is unrelated to the 2 in nthRoot(x, 2)? If so, change the example - I'm also confused like Lili xD

what do you mean by index? I agree more descriptive would be better.

lib/simplifyExpression/collectAndCombineSearch/LikeTermCollector.js
```diff
@@ -220,6 +223,19 @@ function getTermsForCollectingMultiplication(node) {
 }
 
 // A helper function for getTermsForCollectingMultiplication
+// e.g. nthRoot(x, 2), append '2': nthRootNode to terms dictionary
+// Takes the terms dictionary and the nthRoot node, and returns an updated
+// terms dictionary.
+function addToTermsforNthRootMultiplication(terms, node) {
+  const nthRootNode = node;
+  const rootNode = nthRootNode.args[1].value;
```
evykassirer: if you're taking the value I'd call this `rootNodeValue`

lib/simplifyExpression/collectAndCombineSearch/LikeTermCollector.js
evykassirer: I would make a constant value string like CONSTANT at the top of the file, but for nthroot

and then make the key `NTHROOT +  rootNodeValue` or something - makes it easier to debug cause you know it's referring to roots without having to have seen this code

lib/simplifyExpression/collectAndCombineSearch/multiplyLikeTerms.js
```diff
@@ -37,9 +37,36 @@ function multiplyLikeTerms(node, polynomialOnly=false) {
...
 }
 
+function multiplyNthRoots(node) {
+  if (!checks.canMultiplyLikeTermsNthRoots(node)){
+    return Node.Status.noChange(node);
+  }
+
+  let newNode = clone(node);
+
+  // Array of bases of all the nthRoot terms being multiplied
```
evykassirer: these comments are awesome :D thanks

lib/simplifyExpression/collectAndCombineSearch/multiplyLikeTerms.js
```diff
@@ -37,9 +37,36 @@ function multiplyLikeTerms(node, polynomialOnly=false) {
...
+  const radicands = node.args.map(term => term.args[0]);
+
+  // Multiply them
+  const newRadicandNode = Node.Creator.operator('*', radicands);
+
+  // All the args at this point have the same index,
```
evykassirer: I think we've been using a different name for this than index (forget off the top of my head, but the root code will definitely use it - it might just be 'root' i.e why you're calling it rootNode :p)

is index the official name for the "degree" of the root?

aliang8: Yup, it is for nthRoot nodes.

lib/simplifyExpression/collectAndCombineSearch/LikeTermCollector.js
ldworkin: @aliang8 can you make this change? It actually feels important to me. Having just '2' as a key in a dictionary feels like a bad idea. Also it kind of implies that '2' is the common term that we're grouping under, but it's not. I agree with Evy's suggestion above about the constant :)

ldworkin:
@aliang8 When I mentioned negatives, I actually mean -sqrt(x) * sqrt(x), will that get evaluated to -x? Or -sqrt(x) * -sqrt(x) = x? I'm still trying to understand how negatives are handled in the code base.

Other than that, I think everything has been addressed, so @evykassirer do you want to approve?

lib/checks/canMultiplyLikeTermsNthRoots.js
```diff
@@ -0,0 +1,22 @@
+const Node = require('../node');
+
+// Function to check if nthRoot nodes can be multiplied
+// e.g. nthRoot(x, 2) * nthRoot(x, 2) -> true
+// e.g. nthRoot(x, 2) * nthRoot(x, 3) -> false
+function canMultiplyNthRoots(node) {
```
evykassirer: this should be called canMultiplyLikeTermsNthRoots to match the file name

evykassirer:
But yeah once those things have been addressed, good to go :) 
I'll approve and you can merge tomorrow after the fixes!

lib/simplifyExpression/collectAndCombineSearch/LikeTermCollector.js
```diff
@@ -224,7 +224,7 @@ function getTermsForCollectingMultiplication(node) {
 }
 
 // A helper function for getTermsForCollectingMultiplication
-// e.g. nthRoot(x, 2), append '2': nthRootNode to terms dictionary
+// e.g. nthRoot(x, 2), append 'nthRoot2': nthRootNode to terms dictionary
```
evykassirer: oh you had already done this! but just didn't update the comment haha. my bad

PR #191, opened 2017-07-10
small gcd substep fix for changegroups

ldworkin:


lib/simplifyExpression/fractionsSearch/divideByGCD.js
```diff
@@ -50,6 +50,7 @@ function divideByGCD(fraction) {
     '/', [intermediateNumerator, intermediateDenominator]);
   const status = Node.Status.nodeChanged(
     ChangeTypes.FIND_GCD, fraction, intermediateFraction);
+  status.newNode = Node.Status.resetChangeGroups(status.newNode);
```
ldworkin: Looks like this might have been necessary after all -- otherwise in Athena, the whole intermediate node is colored yellow?
img_1303

evykassirer: you can make just the 3s coloured yellow by setting both the 3-node changegroups to 1 (by default the whole thing gets a changegroup, so you'll want to turn off that default in ...arguments to NodeStatus I think?)

Also shouldn't there be another substep for cancelling the 3s?

ldworkin: Yes good call re: another substep!

Would you mind providing a tiny bit more guidance re: the 3-node change groups? I played around with this for awhile but got stuck. 😬  I think I can turn off the default change group by adding `false` as an argument to Node.Status.nodeChanged, so that's good. But how do I add a change group to one of the inner nodes? Is it the Node.Status.childChanged function?

evykassirer: you can set it manually I think

change groups are usually set by default here
https://github.com/socraticorg/mathsteps/blob/master/lib/node/Status.js#L50
which you turn off by passing in false

and then you can just set changeGroup=1 on both the "3" nodes

...does that make sense? It's kind of weird thing so lemme know if you want more explanation

ldworkin: Sweet, that makes sense and it worked! Here's what it looks like now:
image
(Don't pay too much attention to the text around the substeps .. we're actually getting rid of that soon ... which is a whole other conversation)

evykassirer: beautiful 😍

lib/simplifyExpression/fractionsSearch/divideByGCD.js
```diff
@@ -42,30 +44,59 @@ function divideByGCD(fraction) {
     return Node.Status.noChange(fraction);
   }
 
+  // STEP 1: Find GCD
+  // e.g. 15/6 -> 5*3/2*3
```
evykassirer: nit:  (5*3)/(2*3)

PR #192, opened 2017-07-10
fix cancel terms error

aliang8:
sentry bugs:
https://sentry.io/socratic/athena/issues/308248105/
https://sentry.io/socratic/athena/issues/308226210/

Another one:
https://sentry.io/socratic/athena/issues/308235832/

The error was because `polynomialTerms` also includes symbols and powers, which I do not want to be cancelled.

### DIFF
--- a/lib/simplifyExpression/fractionsSearch/cancelLikeTerms.js
+++ b/lib/simplifyExpression/fractionsSearch/cancelLikeTerms.js
@@ -314,7 +314,8 @@ function cancelTerms(numerator, denominator) {
       return new CancelOutStatus(numerator, denominator, false);
     }
 
-    // Sometimes the fraction reduces to a constant e.g. 6 / 2 -> 3
+    // Sometimes the fraction reduces to a constant e.g. 6 / 2 -> 3,
+    // in which case `newCoeff` (the denominator coefficient) should be null
     if (Node.Type.isConstant(reduceStatus.newNode)) {
       numerator = reduceStatus.newNode;
       newCoeff = null;
@@ -341,6 +342,8 @@ function cancelTerms(numerator, denominator) {
       return new CancelOutStatus(reduceStatus.newNode, null, true);
     }
 
+    // Sometimes the fraction reduces to a constant e.g. 6 / 2 -> 3,
+    // in which case `newCoeff` (the denominator coefficient) should be null
     if (Node.Type.isConstant(reduceStatus.newNode)) {
       numerator = reduceStatus.newNode;
       denominator = null;
--- a/lib/simplifyExpression/fractionsSearch/cancelLikeTerms.js
+++ b/lib/simplifyExpression/fractionsSearch/cancelLikeTerms.js
@@ -22,6 +22,7 @@ function cancelLikeTerms(node) {
   if (!Node.Type.isOperator(node) || node.op !== '/') {
     return Node.Status.noChange(node);
   }
+
   let newNode = clone(node);
   const numerator = newNode.args[0];
   const denominator = newNode.args[1];
@@ -30,6 +31,7 @@ function cancelLikeTerms(node) {
   if (!isMultiplicationOfTerms(numerator) &&
       !isMultiplicationOfTerms(denominator)) {
     const cancelStatus = cancelTerms(numerator, denominator);
+
     if (cancelStatus.hasChanged) {
       newNode.args[0] = cancelStatus.numerator || Node.Creator.constant(1);
       if (cancelStatus.denominator) {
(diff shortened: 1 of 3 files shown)


test/simplifyExpression/fractionsSearch/cancelLikeTerms.test.js
```diff
@@ -23,6 +23,7 @@ describe('cancel like terms', function () {
     ['-x / -x', '1'],
     ['2/ (4x)', '1 / (2x)'],
     ['2/ (4x^2)', '1 / (2x^2)'],
+    ['2 a / a', '2'],
```
ldworkin: Can you add the cases that errored somewhere? I guess they might not go into this test because we _don't_ want to cancel terms for them, but let's add them as test cases somewhere else.

lib/simplifyExpression/fractionsSearch/cancelLikeTerms.js
```diff
@@ -291,7 +293,9 @@ function cancelTerms(numerator, denominator) {
   // case 4: the numerator is a constant and denominator is a polynomial term
```
ldworkin: Can you update the comment explaining that the denominator has to be a polynomial term with a coefficient, and that we need to ignore cases like 2/a or 2/x^2 (as you explained to me in Slack :D)

lib/simplifyExpression/fractionsSearch/cancelLikeTerms.js
```diff
@@ -335,7 +341,14 @@ function cancelTerms(numerator, denominator) {
       return new CancelOutStatus(reduceStatus.newNode, null, true);
     }
 
-    [numerator, denominator] = reduceStatus.newNode.args;
+    if (Node.Type.isConstant(reduceStatus.newNode)) {
```
ldworkin: Maybe just add the same comment you have above down here too.

ldworkin:
@evykassirer I'm gonna go ahead and merge this one, just cause it's a hot fix for a bunch of errors we're seeing in Athena. If you have any comments on it, we can definitely address them tomorrow in another PR!

lib/simplifyExpression/fractionsSearch/cancelLikeTerms.js
```diff
@@ -288,10 +290,14 @@ function cancelTerms(numerator, denominator) {
     return new CancelOutStatus(numerator, denominator, true);
   }
 
-  // case 4: the numerator is a constant and denominator is a polynomial term
+  // case 4: the numerator is a constant and denominator is a polynomial term that has a coefficient
+  // or is multiplication node
   // e.g. 2 / 4x -> 1 / 2x
+  // e.g. ignore cases like:  2 / a and 2 / x^2
```
evykassirer: I'm curious what the error was that was happening - the description being only links to sentry isn't great in github because all us non-Socratic peeps can't see it

can you update the PR name, or comment here (or both) with what the error looked like?

cause ` 2 / a` and `2 / x^2` don't cancel anyways right?

lib/simplifyExpression/fractionsSearch/cancelLikeTerms.js
```diff
@@ -308,8 +314,15 @@ function cancelTerms(numerator, denominator) {
       return new CancelOutStatus(numerator, denominator, false);
     }
 
-    [numerator, newCoeff] = reduceStatus.newNode.args;
-
+    // Sometimes the fraction reduces to a constant e.g. 6 / 2 -> 3,
+    // in which case `newCoeff` (the denominator coefficient) should be null
```
evykassirer: aw, sorry for missing this case in review :(

cancelling things is surprisingly tricky ._.

but it's nice that you have so many users that you can catch bugs fast :D I love that about the app

lib/simplifyExpression/fractionsSearch/cancelLikeTerms.js
evykassirer: @ldworkin ?

aliang8: Oh sorry about the Sentry link. I was unaware that you can't view it! I added an extra check to ignore cases where the denominator isn't a coefficient * polynomialTerm. Without that, we would be stuck in that `if` statement.

lib/simplifyExpression/fractionsSearch/cancelLikeTerms.js
aliang8: Yup! It's all good.  💯

lib/simplifyExpression/fractionsSearch/cancelLikeTerms.js
ldworkin: I think (@aliang8 correct me if I'm wrong) the problem was, if there is no coefficient in the denominator, then line 308 `const frac = Node.Creator.operator('/', [numerator, coeff]);` was failing with error "Array containing Nodes expected for parameter "args"". Because `coeff` is null or undefined or something.

evykassirer: ah, yeah if there is no coefficient "getting" the coefficient returns some null/undefined value

woops! would be good to remember to test coefficient/no coefficient cases in the future

PR #193, opened 2017-07-11
fixing a bug in nth root terms multiplication

ldworkin:
The previous code would error if you left out the root node, i.g. `nthRoot(x, 2) * nthRoot(x, 2)` worked but `nthRoot(x) * nthRoot(x)` didn't.

Instead of `node.args[1]` we should always use `getRootNode(node)`.

### DIFF
--- a/lib/checks/canMultiplyLikeTermsNthRoots.js
+++ b/lib/checks/canMultiplyLikeTermsNthRoots.js
@@ -1,4 +1,5 @@
 const Node = require('../node');
+const NthRoot = require('../simplifyExpression/functionsSearch/nthRoot');
 
 // Function to check if nthRoot nodes can be multiplied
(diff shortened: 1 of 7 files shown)


ldworkin:
Going to merge this guys, cause I want to update Athena and I don't want to pull in broken code. Happy to make changes afterwards though!

lib/checks/canMultiplyLikeTermsNthRoots.js
```diff
@@ -1,4 +1,5 @@
 const Node = require('../node');
+const NthRoot = require('../simplifyExpression/functionsSearch/nthRoot');
```
evykassirer: oh yeah!! I forgot this handy thing exists. Glad you found it :D

evykassirer:
awesome fix! ✨

PR #197, opened 2017-07-15
Distribute exponent 

aliang8:
1) Given an expression in the form (base)^exponent, where base is a multiplication node and the exponent is any term, we want to distribute the exponent over each term.

e.g. (x^2 y^2)^(2) -> x^4 y^4
e.g. ((x + 1) (x + 2))^2 -> (x + 1)^2 (x + 2)^2

2) Given a node raised to a negative power, negate the exponent and reciprocate the node. Returns a nodeChanged Status object.

e.g. (x y)^-1 -> 1/(x y)
e.g. (2x^2)^-1 -> 1/(2x^2)

TODO: I will be adding support for distributing exponents over nthRoots soon. There is quite a few things going on in this PR and it would be better for you to start reviewing as I add more support.


lib/simplifyExpression/distributeSearch/index.js
```diff
@@ -7,8 +7,7 @@ const ChangeTypes = require('../../ChangeTypes');
 const Negative = require('../../Negative');
 const Node = require('../../node');
 const TreeSearch = require('../../TreeSearch');
-
-const search = TreeSearch.postOrder(distribute);
+const search = TreeSearch.preOrder(distribute);
```
ldworkin: Why this change?

lib/simplifyExpression/distributeSearch/index.js
```diff
@@ -22,13 +21,117 @@ function distribute(node) {
     return distributeAndSimplifyMultiplication(node);
   }
   else if (Node.Type.isOperator(node, '^')) {
+    /*
+      If it is a power node
```
ldworkin: Love love love this docstring

lib/simplifyExpression/stepThrough.js
```diff
@@ -33,7 +33,7 @@ function stepThrough(node, debug=false) {
   const steps = [];
 
   const originalExpressionStr = print(node);
-  const MAX_STEP_COUNT = 20;
+  const MAX_STEP_COUNT = 100;
```
ldworkin: Did you mean to commit this or was it for debugging only?

test/simplifyExpression/distributeSearch/distributeSearch.test.js
```diff
@@ -97,13 +97,49 @@ describe('expand base', function () {
...
+    ['(x y)^(-2x^2)','1 / ((x * y)^(2x^2))'],
+    ['(x y)^(-(x + 1))', '1 / ((x * y)^((x + 1)))'],
+  ];
+
+  tests.forEach(t => testDistribute(t[0], t[1]));
+});
+
+describe('distribute exponent', function () {
+  const tests = [
+    // When terms are polynomialTerms
```
ldworkin: Omg Anthony this pull is amazing ... so many docstrings and test cases ... almost like you knew me and Evy would bug you about those 😁

test/simplifyExpression/simplify.test.js
```diff
@@ -99,6 +100,25 @@ describe('distribution', function () {
...
+    ['((x + 1) y)^2', 'x^2 * y^2 + 2x * y^2 + y^2'],
+    ['(2x * y * z)^2', '4x^2 * y^2 * z^2'],
+    ['(2x^2 * 3y^2)^2', '36x^4 * y^4'],
+    ['((x + 1)^2 (x + 1)^2)^2', 'x^8 + 8x^7 + 28x^6 + 56x^5 + 70x^4 + 56x^3 + 28x^2 + 8x + 1'],
+    ['(x * y * (2x + 1))^2', '4x^4 * y^2 + 2x^3 * y^2 + 2x^3 * y^2 + x^2 * y^2'],
+    ['((x + 1) * 2y^2 * 2)^2', '16x^2 * y^4 + 16y^4 * x + 16y^4 * x + 16y^4'],
+    ['(2x * (x + 1))^2', '4x^4 + 8x^3 + 4x^2'],
+    ['(x^2 y)^2.5', 'x^5 * y^2.5'],
+    // Fractional exponents
+    ['(x^2 y^2)^(1/2)', 'x * y'],
```
ldworkin: WOW

lib/simplifyExpression/distributeSearch/index.js
```diff
@@ -22,13 +21,117 @@ function distribute(node) {
...
+    const innerBase = Node.Type.isConstant(term) || Node.Type.isSymbol(term) || Node.Type.isParenthesis(term)
+          ? term
+          : term.args[0];
+
+    // e.g. innerExponent for x is null and (x + 1)^2 is 2
+    const innerExponent = Node.Type.isConstant(term) || Node.Type.isSymbol(term) || Node.Type.isParenthesis(term)
+          ? null
+          : term.args[1];
+
+    // If innerExponent is null, just return the o``uterExponent
```
ldworkin: Got some extra chars in your comment there

lib/simplifyExpression/distributeSearch/index.js
```diff
@@ -22,13 +21,117 @@ function distribute(node) {
...
+  const newTerms = base.args.map(function(term) {
+    // Recurses when a term is a multiplication node
+    // e.g. (2x^2 * y)^2 -> 2^2 * x^(2 * 2) * y^2
+    if (Node.Type.isOperator(term, '*')) {
+      const newTerm = Node.Creator.operator('^', [term, outerExponent]);
+
+      return distributeExponent(newTerm).newNode;
+    }
+
+    // e.g. innerBase for x is x, 2 is 2, (x + 1) is (x + 1)
```
ldworkin: So I'm trying to follow this logic below, and I'm wondering why you can't just `return Node.Creator.operator('^', [term, outerExponent]);` right here. 

Say you have `(x^2 * y^2)^3`. Then `newTerms = [(x^2)^3, (y^2)^3]`, and our new node would be `(x^2)^3 * (y^2)^3`. In the next step, wouldn't this get simplified to `x^6 * y^6`?

ldworkin:
This is awesome! I have one question above about your logic for exponent distribution. Also, wondering if the negative exponent business ever gets us into trouble. Are there ever times where we don't want to do this? I might need to think about this more to come up with a possible example ...

lib/simplifyExpression/distributeSearch/index.js
aliang8: The changes that I make to `distributeSearch` in this pull request requires us to traverse the tree in a postOrder fashion starting from the root node down to the leafs. This way we can catch the higher order operations that should be evaluated first. For example in `(x (x + 1))^2`, we want to distribute the exponent first before distributing the inner multiplication. Let me know if this is still confusing!

lib/simplifyExpression/distributeSearch/index.js
aliang8: Ah there's more of this to come ^_^

lib/simplifyExpression/distributeSearch/index.js
aliang8: Haha, I made this fix in my next commit :D

lib/simplifyExpression/distributeSearch/index.js
aliang8: Thanks!

lib/simplifyExpression/stepThrough.js
aliang8: Yup, there are some expressions that take awhile to return a solution.

test/simplifyExpression/distributeSearch/distributeSearch.test.js
aliang8: Of course, I'm a mind reader.

lib/simplifyExpression/distributeSearch/index.js
aliang8: Addressing the second part of your comment, I think we want to show that when we distribute exponents, we can multiply the outer and inner exponents for pedagogy reasons.

aliang8:
This is officially ready for review :D

lib/simplifyExpression/distributeSearch/index.js
ldworkin: That makes sense -- just seems like a big change and I'd be nervous that it might break something else. But all the tests are passing, so ... probably fine!

test/simplifyExpression/simplify.test.js
```diff
@@ -99,15 +100,32 @@ describe('distribution', function () {
...
+    ['(x^3 y^3)^(1/3)', 'x * y'],
+    ['(2x^2 * y^2)^(1/2)', '2^1 / 2 * x * y'],
+    // nthRoot to a power
     ['(nthRoot(x, 2) * nthRoot(x, 2))^2', 'x^2'],
-    // TODO: fix nthRoot to evaluate nthRoot(x^3, 2)
```
ldworkin: Wait, does this work now? How so?

ldworkin:
Wondering, what was the motivation for the negative exponent change? Was it this Asana task: "x/x^2 -> x^-1, but should be 1/x"?

ldworkin:
Also you say in the description "I will be adding support for distributing exponents over nthRoots soon." but from the test cases it looks like that works fine? e.g. `['(nthRoot(x, 2) * nthRoot(x, 3))^2', 'nthRoot(x, 2)^2 * nthRoot(x, 3)^2'`

lib/simplifyExpression/distributeSearch/index.js
```diff
@@ -22,30 +21,175 @@ function distribute(node) {
...
   else {
     return Node.Status.noChange(node);
   }
 }
 
-// Expand a power node with a non-constant base and a positive exponent > 1
-// e.g. (nthRoot(x, 2))^2 -> nthRoot(x, 2) * nthRoot(x, 2)
-// e.g. (2x + 3)^2 -> (2x + 3) (2x + 3)
+// Helper to get content inside parens
+function getContent(node) {
```
ldworkin: Wondering if this function should be somewhere else, like in Util or Node (Evy, thoughts?)

lib/simplifyExpression/distributeSearch/index.js
```diff
@@ -22,30 +21,175 @@ function distribute(node) {
...
+  return Node.Type.isParenthesis(node)
+    ? node.content
+    : node;
+}
+
+function convertNthRootToExponent(node) {
```
ldworkin: Add a docstring :)

lib/simplifyExpression/distributeSearch/index.js
```diff
@@ -22,30 +21,175 @@ function distribute(node) {
...
+  const base = getContent(node.args[0]);
+  const exponent = getContent(node.args[1]);
+
+  const [radicand, root] = base.args;
```
ldworkin: I think you should use the getRadicand and getRoot functions to avoid the bug we had in the other PR!

ldworkin:
I think this PR also fixes this Asana task, right?! https://app.asana.com/0/329831092427581/385493588159834 - "Make (nthRoot(x, 2))^2 one step"

lib/simplifyExpression/distributeSearch/index.js
```diff
@@ -22,30 +21,175 @@ function distribute(node) {
...
+    const exponent = getContent(node.args[1]);
+
+    if (Negative.isNegative(exponent)
+             || (exponent && Negative.isNegative(exponent))) {
+      return reciprocateTerm(node);
+    }
+    else if (Node.Type.isFunction(base, 'nthRoot')) {
+      return convertNthRootToExponent(node);
+    }
+    else if (Node.Type.isOperator(base, '*') || Node.Type.isOperator(base, '^')){
```
ldworkin: I think I would check for these two cases separately, and call `distributeExponent` for mult and `distributeOverPowerExponent` for power. That's what's happening now anyway.

lib/simplifyExpression/distributeSearch/index.js
```diff
@@ -22,30 +21,175 @@ function distribute(node) {
...
+
+      // When the base is a power node, multiply inner and outer exponents
+      // and wrap the new exponent in parens
+      // e.g. (x^2)^2 -> x^(2 * 2)
+      if (Node.Type.isOperator(term, '^')) {
+        return distributeOverPowerNode(term, outerExponent);
+      }
+
+      // e.g. (nthRoot(x,2) * nthRoot(x,3))^2 -> nthRoot(x, 2)^2 * nthRoot(x, 3)^2
+      if (Node.Type.isFunction(term, 'nthRoot')) {
```
ldworkin: I don't think you need this case, since it's handled the same as the default case below.

lib/simplifyExpression/distributeSearch/index.js
```diff
@@ -22,30 +21,175 @@ function distribute(node) {
...
+        return Node.Creator.operator('^', [term, outerExponent]);
+      }
+
+      return Node.Creator.operator('^', [term, outerExponent]);
+    });
+  }
+  else if (Node.Type.isOperator(base, '^')) {
```
ldworkin: As per my other comment, it's a little confusing to have this case stuck at the bottom here. I'd have a different function handle the power case. It's okay if it shares a bit of code with the multiplication case.

lib/simplifyExpression/distributeSearch/index.js
ldworkin: Your fix looks great, thanks!! I'm in love with this PR.

evykassirer:
I think I'll wait for you to address Lili's comments before I take a look :) Then hopefully it'll be a quick review!

PR #200, opened 2017-07-16
quick fix for grouping coefficients change group

ldworkin:
The color was off because the change group was being set on both the coefficient sum *and* the whole node, rather that just the coefficient sum.


ldworkin:
Gonna merge this because it's a one word fix, haha, and I want to deploy Athena before going on jury duty tomorrow!

evykassirer:
ah, good call 👍

PR #201, opened 2017-07-16
Taking the nthRoot of both sides

aliang8:
More support for finding roots of an equation.

e.g. `x^2 - 2 = 0` - > `x = [nthRoot(2, 2)]`
e.g. `(y + 1)^2 = 1` - > `y = 0`


lib/simplifyExpression/stepThrough.js
```diff
@@ -84,10 +84,10 @@ function step(node) {
     breakUpNumeratorSearch,
     // e.g. 3/x * 2x/5 => (3 * 2x) / (x * 5)
     multiplyFractionsSearch,
-    // e.g. (2x + 3)(x + 4) => 2x^2 + 11x + 12
-    distributeSearch,
     // e.g. abs(-4) => 4
     functionsSearch,
+    // e.g. (2x + 3)(x + 4) => 2x^2 + 11x + 12
```
ldworkin: Why this change of order?

lib/ChangeTypes.js
```diff
@@ -180,6 +180,8 @@ module.exports = {
   SIMPLIFY_RIGHT_SIDE: 'SIMPLIFY_RIGHT_SIDE',
   // e.g. x + 3 = 2 -> x + 3 - 3 = 2 - 3
   SUBTRACT_FROM_BOTH_SIDES: 'SUBTRACT_FROM_BOTH_SIDES',
+  // e.g. x^2 = 1, nthRoot(x^2) = nthRoot(1)
+  NTH_ROOT_BOTH_SIDES: 'NTH_ROOT_BOTH_SIDES',
```
ldworkin: Can we do +/- nthRoot(..)? What would we have to do to support +/-?

lib/solveEquation/EquationOperations.js
```diff
@@ -125,13 +124,22 @@ EquationOperations.isolateSymbolOnLeftSide = function(equation, symbolName) {
   let nonSymbolTerm = Symbols.getLastNonSymbolTerm(leftNode, symbolName);
 
   let inverseOp, inverseTerm, changeType;
-  if (!nonSymbolTerm) {
+
+  // If there is no constant term on the left side of the equation
+  // and the left side is not a polynomialTerm
+  // or the left side is a symbol, return no change
+  // e.g. x = 0, no change
```
ldworkin: Anthony is the new docstring king 👑

lib/ChangeTypes.js
aliang8: We just have to display the +/- sign in latex :D

lib/solveEquation/EquationOperations.js
```diff
@@ -125,13 +124,22 @@ EquationOperations.isolateSymbolOnLeftSide = function(equation, symbolName) {
...
+  // e.g. 2x + 3x = 1, no change
+  // e.g. x^2 = 1, change
+  if (!nonSymbolTerm && !Node.Type.isOperator(leftNode, '^')) {
     return EquationStatus.noChange(equation);
   }
 
   // Clone it so that any operations on it don't affect the node already
   // in the equation
-  nonSymbolTerm = clone(nonSymbolTerm);
+  nonSymbolTerm = nonSymbolTerm
```
ldworkin: Why do we need the ternary operator?

lib/solveEquation/EquationOperations.js
```diff
@@ -177,8 +185,9 @@ EquationOperations.isolateSymbolOnLeftSide = function(equation, symbolName) {
       }
     }
     else if (leftNode.op === '^') {
-      // TODO: support roots
```
ldworkin: Loving killing TODOs! 🎉

lib/solveEquation/EquationOperations.js
```diff
@@ -203,11 +212,21 @@ function performTermOperationOnEquation(equation, op, term, changeType) {
   const oldEquation = equation.clone();
 
   const leftTerm = clone(term);
-  const rightTerm = clone(term);
+
+  // The root is the exponent of the polynomial term on the left side
+  const root = op === 'nthRoot'
+        ? leftTerm.args[1]
```
ldworkin: Use the getRoot helper function!!!!

ldworkin:
Just did a very cursory pass, will come back and re-read later today, but looks great!

lib/ChangeTypes.js
ldworkin: Yup there's a latex sign for that! /pm I think

ldworkin: Hmm I actually think the plus/minus might be more complicated ... I ran into it when I was thinking about using the quadratic equation also. It might be tricky cause mathjs doesn't support it. I'm thinking maybe we could use a custom function? Let's talk about it.

test/solveEquation/solveEquation.test.js
```diff
@@ -88,17 +86,32 @@ describe('solveEquation for =', function () {
...
+    ['x^2 = 1', 'x = 1'],
+    ['x^3 - 3 = 0', 'x = nthRoot(3, 3)'],
+    ['(x^2 - 2) (x^2 - 5) = 0', 'x = [nthRoot(2, 2), nthRoot(5, 2)]'],
+    ['(x^2 + 2) (x^3 - 7) = 0', 'x = [nthRoot(-2, 2), nthRoot(7, 3)]'],
+    ['x^2 + 1 = 0', 'x = nthRoot(-1, 2)'],
+    ['(y + 1)^2 = 1', 'y = 0'],
+    ['(y + 1)^3 = 8', 'y = 1'],
+    ['(2x + 1)^3 = 1', 'x = 0'],
+    ['(3x + 2)^2 = 2', 'x = nthRoot(2, 2) / 3 - 2/3'],
+    // TODO: fix this
```
ldworkin: Yea, looks like it fails due to complex #, so should just be moved below

test/solveEquation/solveEquation.test.js
```diff
@@ -88,17 +86,32 @@ describe('solveEquation for =', function () {
...
+    // ['(3x + 2)^2 + 2 = 1']
     // TODO: fix these cases, fail because lack of factoring support, for complex #s,
     // for taking the sqrt of both sides, etc
```
ldworkin: This comment is no longer right, because we do take the square root of both sides now, so any cases that were previously failing because of that shouldn't be anymore (and should be uncommented out)

lib/simplifyExpression/stepThrough.js
aliang8: We want to evaluate the multiplication of roots before distributing.

ldworkin: Can you give an example so I can see what you mean by this / why it's important?

Also, just thought of something: Does this PR conflict with your PR to expand exponents? E.g. for `(y + 1)^2 = 1`, do we want `y+1 = sqrt(1)` or `(y+1)(y+1) = 1`?

PR #202, opened 2017-07-17
fix case where exponent is in parens

ldworkin:


ldworkin:
Actually this should be handled in removeUnnecessaryParenthesis.

ldworkin:
@evykassirer Do we remove unnecessary parens when dealing with equations? Look like maybe not? e.g. x^(2) = 0 is not getting converted to x^2 = 0, whereas plain x^(2) will get converted. It looks like the only place parens get removed while solving equations is in `solveConstantEquation`. Am I missing something?

evykassirer:
ah, yeah that's probably a bug. I think I was assuming they'd get removed in `simplifyExpression` but if there are no steps we probably don't update it (or maybe that's a recent change because of Anthony's updated) It's probably cleaner to removeUnnecessaryParens in `solveEquation` anyways

ldworkin:
Ooh makes sense, I wasn't thinking about how how `solveEquation` calls `simplifyExpression`  --
I'll take another look at that flow tomorrow and check if a bug was introduced recently.

PR #203, opened 2017-07-17
take the absolute value of negative terms

ldworkin:


test/simplifyExpression/functionsSearch/absoluteValue.test.js
```diff
@@ -10,6 +10,9 @@ describe('abs', function () {
   const tests = [
     ['abs(4)', '4'],
     ['abs(-5)', '5'],
+    ['abs(-x)', 'x'],
+    ['abs(-xy)', 'xy'],
```
evykassirer: this is sort of tricky because if x has value -3 and y ha value 2, then abs(-xy) isn't xy - it really depends on the values of x and y which we don't know since they're variable

but it seems like cymath does this too...  weird. It's like one of those 'teach people to do something wrong because it's simpler for now' things

evykassirer: (that's why we had the check in unsupporteNodes that the argument to abs was a constant, cause otherwise it gets more complicated)

technically abs(-x) is

- x when x >= 0
- -x when x < 0

ldworkin: Oh man, good point! I was basing this off of Cymath and didn't think it through. Ok, will close this.

PR #204, opened 2017-07-18
distribute exponent over div

ldworkin:
I'm making this a PR into @aliang8's existing PR for distributing exponents over multiplication. This simply extends it to do the same for division. I realized we don't currently handle simple cases like (1/2)^2, but this should fix that!


lib/simplifyExpression/distributeSearch/index.js
```diff
@@ -166,6 +171,36 @@ function distributeExponent(node) {
...
+  Given an expression in the form (base)^exponent, where base is a division node
+  and the exponent is any term, we want to distribute the exponent over each term.
+  Returns a nodeChanged Status object.
+
+  e.g. (x^2/y^2)^(2) -> x^4/y^4
+*/
+function distributeExponentOverDivision(node) {
+  // Base is required to be a division node
+  const base = getContent(node.args[0]);
+  if (!Node.Type.isOperator(base, '/')) {
```
aliang8: I don't think we need this check.

lib/simplifyExpression/distributeSearch/index.js
```diff
@@ -166,6 +171,36 @@ function distributeExponent(node) {
...
+    return Node.Status.noChange(node);
+  }
+
+  // Exponent can be any term: polynomial, constant, fraction, etc
+  const outerExponent = getContent(node.args[1]);
+
+  const numerator = base.args[0];
```
aliang8: Should we use `getContent` here in case the numerator/denominator are wrapped in parens: e.g `(x + 3)/(x -3) ^2`

lib/simplifyExpression/distributeSearch/index.js
ldworkin: Yea, I know it's redundant because we call it before calling this function, but I sort of like having it in here too? Because sometimes we call functions without knowing if they will "work" or not, so I feel like it's good practice to check first.

lib/simplifyExpression/distributeSearch/index.js
ldworkin: Ooh so this is a good point and I actually tried it, but then it broke my tests, and I realized we actually want to keep the parens. E.g. `((5x)/6)^2` should go to `(5x)^2/6^2`, not `5x^2/6^2`, so that it's clear the exponent is operating on `5x` and not just `x`.

ldworkin:
Gonna merge this in to Anthony's original PR now, so that way @evykassirer only has to review that one for both of us :)

PR #205, opened 2017-07-18
remove unnecessary parens before solving equations

ldworkin:
Previously unnecessary parens were removed when we simplified left/right sides of an equation. But now that we have this alternate "find roots" flow, that doesn't always happen. e.g. `x^(2) = 0` solves `x = 0` only, so `x` is simplified and `0` is simplified, but never `x^2`. So instead let's simplify an equation right away to be safe.

### DIFF
--- a/lib/solveEquation/stepThrough.js
+++ b/lib/solveEquation/stepThrough.js
@@ -51,9 +51,9 @@ function stepThrough(leftNode, rightNode, comparator, debug=false) {
   const MAX_STEP_COUNT = 20;
   let iters = 0;
 
-  // Remove unnecessary parenthesis here, before doing the find roots check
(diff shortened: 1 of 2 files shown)


lib/solveEquation/stepThrough.js
```diff
@@ -51,6 +51,13 @@ function stepThrough(leftNode, rightNode, comparator, debug=false) {
   const MAX_STEP_COUNT = 20;
   let iters = 0;
 
+  // Remove unnecessary parentheses here, before doing the find roots check
+  // If we have roots, we return early and do not go through simplification,
+  // so we can't rely on that flow for parentheses removal
+  // e.g. x^(2) = 0
```
evykassirer: e.g. x^(2) = 0 -> x^2 = 0

PR #206, opened 2017-07-18
distribution flag

ldworkin:
So after spending a billion hours trying to make this change: https://cl.ly/0g162T0g3d19 (i.e. replace multiplication dots with parenthesis during distribution) I finally came up with this solution. We just set a flag on the nodes that are a "result of distribution" and remove the flag when they are simplified. Then Athena can display these terms however she wants.

I realize it's a little weird but I've spent the past 24 hours (no joke) on this, haha, and this was the only solution I came up with that worked in all cases!

### DIFF
--- a/lib/node/Status.js
+++ b/lib/node/Status.js
@@ -91,11 +91,6 @@ Status.childChanged = function(node, childStatus, childArgIndex=null) {
   }
   else if ((Type.isOperator(node) || Type.isFunction(node) &&
             childArgIndex !== null)) {
-
-    if (childStatus.newNode.distribution == false && newNode.distribution == true) {
-      newNode.distribution = false;
-    }
-
     oldNode.args[childArgIndex] = childStatus.oldNode;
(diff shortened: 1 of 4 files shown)


evykassirer:
I guess this makes sense, but I'd like to wait until our conversation tomorrow - since if you do make a fork, I think it's make sense to merge that there but not in mathsteps (it's kinda a specific use case for Athena/Socratic :p)

ldworkin:
Yes, I realize that, but this repo is also owned by Socratic, and when I spend two days on a bug fix we need that doesn't negatively affect code style or functionality, I'd like to be able to merge it in.

evykassirer:
oh ok if it's urgent before tomorrow's meeting then go for it!

ldworkin:
Nah it can wait, sorry for the snark 🙊 ! Was just frustrated after working on this for so freaking long, haha. Definitely makes sense to see what the outcome of the meeting is first.

ldworkin:
Closing, will do it in our fork!

PR #207, opened 2017-07-18
distribute fractions normally

ldworkin:
Not sure why we had special logic for fractions before.

Previously:
4(5/4 + x) => (5 * 4)/4 + x * 4

Now:
4(5/4 + x) => 4 * (5/4) + 4 * x


evykassirer:
omg lol I think I know why this was the case - remember how the cancelling didn't work that well before? I think this was to get around that (not totally sure but that's my guess for now)

test/simplifyExpression/distributeSearch/distributeSearch.test.js
```diff
@@ -53,32 +53,31 @@ describe('distribute', function () {
...
-        '((1 * (2x + x^2)) / (x^2 - 4) + (8x^3 + 4x^4))']
+    ['(2x + x^2) * (1/x + 4x)',
+      ['(2x * (1 / x + 4x) + x^2 * (1 / x + 4x))',
+        '((2x * 1 / x + 8x^2) + (x^2 * 1 / x + 4x^3))']
     ],
 
-    ['(2x + x^2) * (3x^2 / (x^2 -4) + 4x^2)',
-      ['((3x^2 * (2x + x^2)) / (x^2 - 4) + 4x^2 * (2x + x^2))',
-        '((3x^2 * (2x + x^2)) / (x^2 - 4) + (8x^3 + 4x^4))']
+    ['(x - 4) * ((3x^2)/ (x^2 - 4) + 1)',
```
evykassirer: is there a reason why you changed these tests?

ldworkin: Well the old ones were failing -- the answers were equivalent but just came out looking different due to the change. I tried fixing them, but they were pretty complicated and it was hard to follow, so I replaced them with some slightly simpler ones, so it's easier to see what's going on.

ldworkin:
Make sense re: cancelling ... I think?! Though the change I made to the cancelling code was to make it work for parens, and there are no parens here ... anyway hopefully this doesn't break anything terribly! I tested a bunch and Gen will do so more.

Issue #163, opened 2017-05-04
Better sum-product factoring steps


aelnaiem:
The function `factorSumProductRule` currently returns one step with no substeps, but it should have substeps. The substeps are outlined in the code, but just need to be formalized, given a change type etc.
```
        // 1. Break apart the middle term into two terms using our factor pair (p and q):
        //    e.g. ax^2 + bx + c -> ax^2 + px + qx + c
        // 2. Consider the first two terms together and the second two terms
        //    together (this doesn't require any actual change to the expression, so doesn't need a substep)
        //    e.g. first group: [ax^2 + px] and second group: [qx + c]
        // 3. Factor both groups separately (can be two substeps)
        //    e.g first group: [ux(rx + s)] and second group [v(rx + s)]
        // 4. Finish factoring by combining the factored terms through grouping:
        //    e.g. (ux + v)(rx + s)
```

This should make it easier to understand for students to understand the process for what can be a complicated set of steps.

evykassirer:
if you're new to `mathsteps` and want to know more about what this would involve,

- The file you'd be updating for this change would be [here](https://github.com/socraticorg/mathsteps/blob/master/lib/factor/factorQuadratic.js#L228)
- I think [multiplying fractions](https://github.com/socraticorg/mathsteps/blob/master/lib/simplifyExpression/fractionsSearch/addConstantFractions.js) is a pretty good example of how substeps work

alainakafkes:
Assuming this is still open, I'd like to work on it! 🙋

evykassirer:
Ooo hi @alainakafkes thanks so much for wanting to work on this!

Right now we've put a hold on changes to mathsteps because we're doing a big refactor, but Anthony ( @aliang8 ) who's heading this up can help you find a task related to the refactor to work on, if you're interested! (he's working on mathsteps as his full time job right now!) 

Sorry for the deceiving open issues heh, but very excited that you want to work on mathsteps! Let us know if you have any questions and hopefully we can find something else part of the refactor that you'd like to work on :)

nitin42:
@alainakafkes would you like to work on the website ?

evykassirer:
(to give more details)

The website is a demo site that will let people try out `mathsteps` to see how it works without having to install it locally first. 😺 Nitin's still working on the base of how it'll look so it might be a week or so (?) before we have a base codebase for that and you can contribute.

alainakafkes:
[...]

PR #210, opened 2017-07-21
#163 - Better sum-product factoring steps

alainakafkes:
Following the conventions used in [addConstantFractions.js](https://github.com/socraticorg/mathsteps/blob/master/lib/simplifyExpression/fractionsSearch/addConstantFractions.js), I broke down factorQuadratic into 3 substeps:

- Break apart the middle term into two terms using our factor pair (p and q): ax^2 + bx + c --> ax^2 + px + qx + c
- Factor both groups separately: first group as [ux(rx + s)] and second group as [v(rx + s)]
- Finish factoring by combining the factored terms through grouping: (ux + v)(rx + s)

Please let me know if there are any problems and how I can fix them! 😊

### DIFF
--- a/lib/ChangeTypes.js
+++ b/lib/ChangeTypes.js
@@ -35,12 +35,14 @@ module.exports = {
   // e.g. 2 - - 3 -> 2 + 3
   RESOLVE_DOUBLE_MINUS: 'RESOLVE_DOUBLE_MINUS',
 
-  // COLLECT AND COMBINE
+  // COLLECT AND COMBINE AND BREAK UP
 
   // e.g. 2 + x + 3 + x -> 5 + 2x
   COLLECT_AND_COMBINE_LIKE_TERMS: 'COLLECT_AND_COMBINE_LIKE_TERMS',
   // e.g. x + 2 + x^2 + x + 4 -> x^2 + (x + x) + (4 + 2)
   COLLECT_LIKE_TERMS: 'COLLECT_LIKE_TERMS',
+  // e.g. 2x^2 + 4x + 2 -> 2x^2 + 2x + 2x + 2
+  BREAK_UP_TERM: 'BREAK_UP_TERM',
 
   // MULTIPLYING CONSTANT POWERS
   // e.g. 10^2 * 10^3 -> 10^(2+3)
--- a/lib/ChangeTypes.js
+++ b/lib/ChangeTypes.js
@@ -41,8 +41,6 @@ module.exports = {
   COLLECT_AND_COMBINE_LIKE_TERMS: 'COLLECT_AND_COMBINE_LIKE_TERMS',
   // e.g. x + 2 + x^2 + x + 4 -> x^2 + (x + x) + (4 + 2)
   COLLECT_LIKE_TERMS: 'COLLECT_LIKE_TERMS',
-  // e.g. 2x^2 + 4x + 2 -> 2x^2 + 2x + 2x + 2
-  BREAK_UP_TERM: 'BREAK_UP_TERM',
 
   // MULTIPLYING CONSTANT POWERS
   // e.g. 10^2 * 10^3 -> 10^(2+3)
@@ -187,4 +185,6 @@ module.exports = {
   FACTOR_PERFECT_SQUARE: 'FACTOR_PERFECT_SQUARE',
   // e.g. x^2 + 3x + 2 -> (x + 1)(x + 2)
   FACTOR_SUM_PRODUCT_RULE: 'FACTOR_SUM_PRODUCT_RULE',
+  // e.g. 2x^2 + 4x + 2 -> 2x^2 + 2x + 2x + 2
+  BREAK_UP_TERM: 'BREAK_UP_TERM',
 };
--- a/lib/factor/factorQuadratic.js
+++ b/lib/factor/factorQuadratic.js
@@ -245,7 +245,8 @@ function factorSumProductRule(node, symbol, aValue, bValue, cValue, negate) {
         const ax2 = Node.Creator.polynomialTerm(symbol, Node.Creator.constant(2), a);
         const bx = Node.Creator.polynomialTerm(symbol, null, b);
 
-        // OPTIONAL SUBSTEP (this happens iff a is negative): ax^2 + bx + c -> -(-ax^2 - bx - c)
+        // OPTIONAL SUBSTEP (this happens iff a is negative)
+        // ax^2 + bx + c -> -(-ax^2 - bx - c)
         if (negate) {
           newNode = Node.Creator.operator('+', [ax2, bx, c], true);
           newNode = Negative.negate(newNode);
--- a/lib/factor/factorQuadratic.js
+++ b/lib/factor/factorQuadratic.js
@@ -209,6 +209,8 @@ function factorPerfectSquare(node, symbol, aValue, bValue, cValue, negate) {
 // e.g. x^2 + 3x + 2 -> (x + 1)(x + 2) or
 // or   2x^2 + 5x + 3 -> (2x - 1)(x + 3)
 function factorSumProductRule(node, symbol, aValue, bValue, cValue, negate) {
+  let newNode;
+
   if (bValue && cValue) {
     // we factor out the gcd first, providing us with a modified expression to
     // factor with new a, b and c values
@@ -225,9 +227,9 @@ function factorSumProductRule(node, symbol, aValue, bValue, cValue, negate) {
     for (const pair of factorPairs) {
       if (pair[0] + pair[1] === bValue) {
         // To factor, we go through some transformations
-        // TODO: these should be actual substeps
-        // 1. Break apart the middle term into two terms using our factor pair (p and q):
-        //    e.g. ax^2 + bx + c -> ax^2 + px + qx + c
+        // TODO: these should be actual substeps *alaina*
+        // 1. Break apart the middle term into two terms using our factor pair
+        //    (p and q): e.g. ax^2 + bx + c -> ax^2 + px + qx + c
         // 2. Consider the first two terms together and the second two terms
         //    together (this doesn't require any actual change to the expression)
         //    e.g. first group: [ax^2 + px] and second group: [qx + c]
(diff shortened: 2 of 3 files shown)


lib/factor/factorQuadratic.js
```diff
@@ -225,19 +227,44 @@ function factorSumProductRule(node, symbol, aValue, bValue, cValue, negate) {
...
         //    e.g. first group: [ax^2 + px] and second group: [qx + c]
         // 3. Factor both groups separately
         //    e.g first group: [ux(rx + s)] and second group [v(rx + s)]
         // 4. Finish factoring by combining the factored terms through grouping:
         //    e.g. (ux + v)(rx + s)
+        const substeps = [];
+        let status;
+
+        // 1. Break apart the middle term into two terms using our factor pair
+        // (p and q)
```
evykassirer: I think it's a bit redundant to put the steps above and also here - I could see putting them in detail at the beginning (so you can see them all together) and then just `// STEP 1` above step 1 for example - but putting each detailed comment above each step is also fine. 

The way you have it here probably is also fine (yay comments) so feel to disagree

lib/factor/factorQuadratic.js
```diff
@@ -225,19 +227,44 @@ function factorSumProductRule(node, symbol, aValue, bValue, cValue, negate) {
...
         const pValue = pair[0];
         const qValue = pair[1];
 
+        // SUBSTEP: ax^2 + bx + c -> ax^2 + px + qx + c
+        const sq = Node.Creator.constant(2);
```
evykassirer: `sq` was confusing for me as I read it over, at first - maybe define it right before you create the node that uses it, or rename it to `two` or just put it right in the other creator function `Node.Creator.polynomialTerm(symbol, Node.Creator.constant(2), a);` or some combination of those ^_^

lib/factor/factorQuadratic.js
```diff
@@ -225,19 +227,44 @@ function factorSumProductRule(node, symbol, aValue, bValue, cValue, negate) {
...
+        const b = Node.Creator.constant(bValue);
+        const c = Node.Creator.constant(cValue);
+        const p = Node.Creator.constant(pValue);
+        const q = Node.Creator.constant(qValue);
+        const ax2 = Node.Creator.polynomialTerm(symbol, sq, a);
+        const px = Node.Creator.polynomialTerm(symbol, null, p);
+        const qx = Node.Creator.polynomialTerm(symbol, null, q);
+        newNode = Node.Creator.operator('+', [ax2, px, qx, c], true);
+        status = Node.Status.nodeChanged(
+          ChangeTypes.SIMPLIFY_ARITHMETIC, node, newNode);
```
evykassirer: I'd make a new change type for this since it's a pretty unique use case and it'd be nice to describe it more accurately :)

lib/factor/factorQuadratic.js
```diff
@@ -225,19 +227,44 @@ function factorSumProductRule(node, symbol, aValue, bValue, cValue, negate) {
...
+        substeps.push(status);
+        newNode = Node.Status.resetChangeGroups(status.newNode);
+
+        // 2. Consider the first two terms together and the second two terms
```
evykassirer: I find it odd how we point this out in the code but not in the steps we show. I think we should either wrap them in parens `(ax^2 + px) + (qx + c)` and make it a step, or take out references to it in the code and change 'Factor the first group` to `Factor the first two terms` or something

Not your fault - you did what it said - just leftover thinking about Ahmed's planning as we move into concrete substeps ^_^

lib/factor/factorQuadratic.js
```diff
@@ -249,6 +276,7 @@ function factorSumProductRule(node, symbol, aValue, bValue, cValue, negate) {
         const firstParen = Node.Creator.parenthesis(
           Node.Creator.operator('+', [rx, s]));
 
+        // 3B. Factor the second group
```
evykassirer: I'm wondering if it'd be better to factor the first group and second group in two different substeps instead of the same one - what do you think?

lib/factor/factorQuadratic.js
```diff
@@ -262,8 +290,17 @@ function factorSumProductRule(node, symbol, aValue, bValue, cValue, negate) {
         const secondParen = Node.Creator.parenthesis(
           Node.Creator.operator('+', [ux, v]));
 
+        // SUBSTEP: ux(rx + s) + v(rx + s)
```
evykassirer: can you do a `before -> after` here too like you did above? :D

(and add it below too)

lib/factor/factorQuadratic.js
```diff
@@ -225,19 +227,44 @@ function factorSumProductRule(node, symbol, aValue, bValue, cValue, negate) {
...
+        let status;
+
+        // 1. Break apart the middle term into two terms using our factor pair
+        // (p and q)
         const pValue = pair[0];
         const qValue = pair[1];
 
+        // SUBSTEP: ax^2 + bx + c -> ax^2 + px + qx + c
+        const sq = Node.Creator.constant(2);
+        const a = Node.Creator.constant(aValue);
```
evykassirer: I think there are some edge cases to consider here - the only one I can think of off the top of my head is that 'a' could defs be 1 here (i.e. x^2 + bx + c) - maybe poke around a bit in the logic that calls this code to see what is for sure defined and what isn't? I'll think a bit more about this too

lib/factor/factorQuadratic.js
alainakafkes: Agree! I'm going to implement this.

lib/factor/factorQuadratic.js
alainakafkes: Ah, I'm a bit confused by this. What would be the problem when a = 1?

lib/ChangeTypes.js
```diff
@@ -35,12 +35,14 @@ module.exports = {
...
 
-  // COLLECT AND COMBINE
+  // COLLECT AND COMBINE AND BREAK UP
 
   // e.g. 2 + x + 3 + x -> 5 + 2x
   COLLECT_AND_COMBINE_LIKE_TERMS: 'COLLECT_AND_COMBINE_LIKE_TERMS',
   // e.g. x + 2 + x^2 + x + 4 -> x^2 + (x + x) + (4 + 2)
   COLLECT_LIKE_TERMS: 'COLLECT_LIKE_TERMS',
+  // e.g. 2x^2 + 4x + 2 -> 2x^2 + 2x + 2x + 2
+  BREAK_UP_TERM: 'BREAK_UP_TERM',
```
evykassirer: might make more sense to put this rule where the factoring changegroups are, since that helps give context into what it's used for

lib/factor/factorQuadratic.js
evykassirer: oh I think in that case you'd end up with a step like `1x^2 + 2x + 3x + 6` or something, where the 1 shouldn't actually be there

lib/factor/factorQuadratic.js
```diff
@@ -225,45 +227,88 @@ function factorSumProductRule(node, symbol, aValue, bValue, cValue, negate) {
...
-        // u = gcd(a, p), r = a/u and s = p/u
+        const a = Node.Creator.constant(aValue);
+        const c = Node.Creator.constant(cValue);
+        const p = Node.Creator.constant(pValue);
+        const q = Node.Creator.constant(qValue);
+        const ax2 = Node.Creator.polynomialTerm(symbol, Node.Creator.constant(2), a);
+        const px = Node.Creator.polynomialTerm(symbol, null, p);
+        const qx = Node.Creator.polynomialTerm(symbol, null, q);
+
+        // SUBSTEP: ax^2 + bx + c -> ax^2 + px + qx + c
```
evykassirer: mrah sorry for not commenting on this last time, but I think it's confusing to say substep after step - I think just merge them into one comment at the top

```
// SUBSTEP 1: ax^2 + bx + c -> ax^2 + px + qx + c
...
// SUBSTEP 2 ...
```

you see what I mean?

aelnaiem:
Dope!

alainakafkes:
Okay @evykassirer I think I addressed all of your requested changes! 🎉  Please let me know if I should make any more.

lib/factor/factorQuadratic.js
evykassirer: @alainakafkes ^

evykassirer:
changes look great! just one unaddressed comment I commented on

also

> I'd love to see some new tests showing how this works - there are some tests already that test substeps (the fractions code you were basing this off of probably has some!) so you can do some stuff similar to that!

lib/factor/factorQuadratic.js
alainakafkes: @evykassirer I think based on the tests I wrote, that extra "1" coefficient doesn't show up in front of x^2. Let me know if I've misunderstood!

evykassirer: oh cool that must be handled somewhere else in the code then!

yay tests and yay code that just does what it should 🎉

test/factor/factorQuadratic.test.js
```diff
@@ -46,3 +46,64 @@ describe('factorQuadratic', function () {
...
+    ],
+    ['x^2 + x - 2',
+      ['x^2 - x + 2x - 2',
+        '(x^2 - x) + (2x - 2)',
+        'x * (x - 1) + (2x - 2)',
+        'x * (x - 1) + 2 * (x - 1)',
+        '(x - 1) * (x + 2)']
+    ],
+    ['-x^2 - 3x - 2',
+      ['x^2 + x + 2x + 2', // per factorQuadratic, negative is added back in at the end
```
evykassirer: oo good test

this made sense when we didn't have substeps - here can we wrap all the substeps in unary minus?

the substeps should probably look like:

```
-x^2 - 3x - 2
- (x^2 + 3x + 2)
- ((x^2 + x) + (2x + 2))
- (x * (x + 1) + (2x + 2))
....
- (x * (x + 1) + 2 * (x + 1))
- -(x + 1) * (x + 2)
```

what do you think? sorry I keep asking you to do more things hehe

test/factor/factorQuadratic.test.js
```diff
@@ -46,3 +46,64 @@ describe('factorQuadratic', function () {
...
+
+describe('factorSumProductRule', function() {
+  const tests = [
+    // sum product rule
+    ['x^2 + 3x + 2',
+      ['x^2 + x + 2x + 2',
+        '(x^2 + x) + (2x + 2)',
+        'x * (x + 1) + (2x + 2)',
+        'x * (x + 1) + 2 * (x + 1)',
+        '(x + 1) * (x + 2)']
```
evykassirer: eee this is awesome look at this sum product rule awesomeness

test/factor/factorQuadratic.test.js
alainakafkes: Lol it's ok! So I wasn't sure what to do here when I wrote this test. I knew that the negative sign would be added back at the end, but I didn't know how to convey that in the test case.

When I tried what you just suggested, I got this error from deepEqual. Should I leave what I had or does this point to a deeper problem with deepEqual?
screen shot 2017-08-09 at 1 51 26 pm

evykassirer: mm so I think the unary minus just gets dropped in the code right now (unless I'm missing something) so you'd wanna go in and update the code when making the steps to add the unary minus bit around each substep if needed

evykassirer: edge casssesssss

alainakafkes: Yes that makes much more sense. 😂  Looks I just had to negate each ```newNode``` along the way.

evykassirer: sweet - not too bad of a fix then whew

PR #211, opened 2017-07-26
Add LaTeX printing

### LINKED ISSUE #160: Add a function for printing equations in LaTeX
We can print equations to asciimath, but there's no one function call for LaTeX.

It would live in https://github.com/socraticorg/mathsteps/blob/master/lib/equation/Equation.js below `print` and look like:

`this.leftNode.toTex() + this.comparator + this.rightNode.toTex()`

SirFurness:
Follows these steps:
- Get LaTeX from 'toTex()'
- Change '+-' based on the 'showPlusMinus' variable
- Replace multiplication of the type 2*x with just 2x

I added some tests for it but I am not sure if I did them correctly. Attempts to resolve #160


SirFurness:
The returned LaTeX does not have '$' before and after as would normally be had for math mode. I decided to leave them out because the provided 'toTex()' function did not include it. This can easily be added if it turns out it's needed.

SirFurness:
Realized now that it only simplifies multiplication for variables a-zA-Z when there is a function in Symbols.js for helping with this. I'll make sure to update the identification of the variables once I wake up tomorrow.

SirFurness:
Turns out there are mathjs options for pretty much everything I added. Off to fix this!

evykassirer:
oooo nice  - yay mathjs! I'll wait for your fixes before reviewing ^_^

SirFurness:
No, the mathjs option didn't work. It only makes the multiplication implicit if it was passed in that way. I'm done with all my fixes.

evykassirer:
don't have time to look in depth, but before I review could you move the printing in latex code to `print.js`? then we can use it for expressions as well as equations :) 

let me know if you have any questions or concerns about that

SirFurness:
Added!

evykassirer:
oh, and then call that code in the equation code so it's not duplicated ^_^

SirFurness:
Of course, thanks for reminding me

lib/util/print.js
```diff
@@ -106,4 +107,46 @@ function printTreeTraversal(node, parentNode) {
   }
 }
 
+// Prints an expression node in LaTeX
+function printLatex(node, showPlusMinus=false) {
+  // Creates the LaTeX
+  function intoTex(node, showPlusMinus=false) {
+    const originalNodeTex = node.toTex();
+
+    let outputNodeTex;
```
evykassirer: I think you could just call this `nodeTex` and not have both `originalNodeTex` and `outputNodeTex` since you only seem to use one at a time - what do you think? Or did you deliberately put this here to be more readable (I could be convinced ^^)

lib/util/print.js
```diff
@@ -106,4 +107,46 @@ function printTreeTraversal(node, parentNode) {
...
+
+    if (!showPlusMinus) {
+      // Replaces '+-' with '-'
```
evykassirer: can you add the `// (The + - is needed to support the conversion of subtraction to addition of....` comment from above here too? it's a weird thing and I want people looking at this code to know why this is happening

test/equation.test.js
```diff
@@ -26,3 +41,12 @@ describe('Equation constructor', () => {
   ];
   tests.forEach(t => testEquationConstructor(t[0], t[1], t[2], t[3]));
 });
+
+describe('Latex printer', () => {
+  const tests = [
+    ['2*x^2 + x', '4', '=', '2{x}^{2}+x = 4'],
+    ['x^2 + 2*y + 2', '0', '>=', '{x}^{2}+2y+2 >= 0'],
+    ['2*x + -1', '0', '<=', '2x-1 <= 0']
```
evykassirer: this should just be `2x - 1` (see comment below on `+ -`

lib/util/print.js
```diff
@@ -106,4 +107,46 @@ function printTreeTraversal(node, parentNode) {
...
+    }
+
+    const nodeSymbols = Symbols.getSymbolsInExpression(node);
+
+    // Creates the regex parts separately so that current symbol can be added
+    // in the middle
+    const regexFirstHalf = '\\s*?\\\\cdot\\s*?({?)\\s*?';
+    const regexSecondHalf = '\\s*?(}?)';
+
+    // Checks each symbol for explicit multiplication and removes it
```
evykassirer: Hmmm don't think we need this stuff. More discussion in the general comment of this review. I'm open to discussion - let me know what you think!

SirFurness:
You're right, the person calling the function should have control over the explicit/implicit multiplication. I just found it ugly and couldn't help myself but remove it entirely. I'll make sure to fix it all up.

test/util/print.test.js
```diff
@@ -29,6 +34,16 @@ describe('print asciimath', function () {
   tests.forEach(t => testPrintStr(t[0], t[1]));
 });
 
+describe('print latex', function() {
+  const tests = [
+    ['2+3+4', '2+3+4'],
+    ['2 + (4 - x) + - 4', '2+\\left(4-x\\right)-4'],
```
SirFurness: I discovered 'TestUtil.flattenAndParse()' and thought it would be good to use instead. I also found that many of the tests aren't using it. Is it ok if I switch them over to it?

evykassirer:
Hahaha. It can be ugly sometimes but sometimes it's good for teaching to show it explicitly. Thanks for cleaning it up

SirFurness:
So it turns out flattening the tree makes, only where possible, all the explicit multiplication implicit. So all the explicit multiplication is being hidden anyway.

.gitignore
```diff
@@ -1,5 +0,0 @@
-*swp
```
evykassirer: oh noes can you bring this back?

lib/equation/Equation.js
```diff
@@ -13,9 +13,19 @@ class Equation {
...
+    const leftSide = printNode.ascii(this.leftNode, showPlusMinus);
+    const rightSide = printNode.ascii(this.rightNode, showPlusMinus);
+    const comparator = this.comparator;
+
+    return `${leftSide} ${comparator} ${rightSide}`;
+  }
+
+  // Prints an Equation properly using LaTeX
+  latex(showPlusMinus=false) {
+
```
evykassirer: don't need this newline

lib/util/print.js
```diff
@@ -106,4 +106,19 @@ function printTreeTraversal(node, parentNode) {
...
+
+  if (!showPlusMinus) {
+    // Replaces '+ -' with '-'
+    nodeTex = nodeTex.replace(/\s*?\+\s*?\-\s*?/g, ' - ');
+  }
+
+  return nodeTex;
+}
+
+module.exports.ascii = print;
```
evykassirer: I think you can do 

```
module.exports = {
   ascii: print
   latex: printLatex
}
```

also can you rename print to printAscii?

test/Negative.test.js
```diff
@@ -1,15 +1,12 @@
...
 const print = require('../lib/util/print');
 
 const Negative = require('../lib/Negative');
 
 const TestUtil = require('./TestUtil');
 
 function testNegate(exprString, outputStr) {
-  const inputStr = Negative.negate(flatten(math.parse(exprString)));
-  TestUtil.testFunctionOutput(print, inputStr, outputStr);
+  const inputStr = Negative.negate(TestUtil.flattenAndParse(exprString));
```
evykassirer: yay thanks for updating this everywhere - would you mind find+replacing `flattenAndParse` everywhere with `parseAndFlatten`? heh. just cause it happens in that order and therefore that'd be a bit clearer

test/equation.test.js
```diff
@@ -26,3 +43,12 @@ describe('Equation constructor', () => {
   ];
   tests.forEach(t => testEquationConstructor(t[0], t[1], t[2], t[3]));
 });
+
+describe('Latex printer', () => {
+  const tests = [
+    ['2*x^2 + x', '4', '=', '2~{ x}^{2}+ x = 4'],
+    ['x^2 + 2*y + 2', '0', '>=', '{ x}^{2}+2~ y+2 >= 0'],
+    ['2*x - 1', '0', '<=', '2~ x - 1 <= 0']
```
evykassirer: try some without the `*` too! your fave

e.g. this one could be `2x - 1`

.gitignore
SirFurness: Woops. I tried to remove mine from going into the commit.

SirFurness:
Should be good now

Issue #209, opened 2017-07-20
Solve/simplify using scope (a la math.js' eval)


Ephs05msm:
Hello,

Thanks so much for your work on this project. Super interesting and the mission is such a worthy one. I found you all via the math.js site -- then read the blog post -- and am curious whether mathsteps can use an array of symbols (e.g., `{ x: 10, foo: 20, bar: "foo + x", baz: "bar^2" }`) passed in as a scope. In math.js this can be done in math.eval(expr, scope). But math.js returns just the solution and not the steps. I imagine this boils down to (at least) a "change type" of substituting symbols for their respective values. Given the potentially recursive nature of traversing the symbols (e.g., `baz` in the example above), this may not be a simple thing ... but that's also a bit of what mathsteps is doing I believe.

I poked around the codebase and don't see an obvious way to accomplish this so I thought I would ask.

Thanks again,
Matt

tkosan:
Substitution of symbols is a capability of computer algebra systems (CAS), and MathSteps is slowly accumulating CAS capabilities. MathSteps is able to perform symbol substitutions, but I don't think support has been included yet for the example you provided. However, I don't think it would be difficult to implement. I created the following solution to the problem you posted using the MathPiper CAS:

[Evaluate a list using substitution](http://patternmatics.org/mathsteps/tmp/eval_using_substitution.html)

Is this approach similar to the one you have in mind?

Also, what are you interested in using this kind of capability for?

Ephs05msm:
Thanks for the reply and MathPiper example. Yes, that approach works well.

My interest in this capability is in the context of how spreadsheets are sometimes used: one might build up a range of inputs and then reference those inputs in formulas elsewhere. Evaluating the formulas requires substituting cell addresses (e.g., A1, B2, etc.) with the values. I'd like to explore an expanded way of auditing formulas that shows the steps. Excel includes a built-in tool to evaluate formulas that does the substitution and can step through the calculation, but it will not display more than one step at a time.

I'm interested in trying to use MathSteps to display many/all of the calculation steps (including the substitution step) to see if it helps with understanding more complex formulas.

Is there something that I can do to help add this capability to MathSteps? I'm somewhere between beginner and intermediate in terms of experience. With some guidance/oversight, I believe I can be helpful.

Thanks again.

tkosan:
[...]

PR #212, opened 2017-07-27
WIP: Substitute symbols for expressions from optional scope

Ephs05msm:
This is some work based on a discussion in #209. This happens to be my first ever pull request, so please provide any feedback on things that seem out of sorts. I referenced the contributing guidelines and then followed the instructions from [this video](https://egghead.io/courses/how-to-contribute-to-an-open-source-project-on-github) to get to this point.

With that said, the approach I ended up taking toward this issue was to:

- Add an optional scope argument to solveEquationString and simplifyExpressionString
- Add a module, Substitute.js, that iterates over a provided scope and replaces symbols with their respective expressions. The intent is to account for nested symbols and the unit test I added to solveEquation.test.js uses such a scenario. I placed the module in ./lib but it likely belongs somewhere else.
- Add a ChangeType, SUBSTITUTE_SCOPE_SYMBOL, that captures instances where symbols are being substituted. I felt this was important to keep the user experience consistent with what it currently is.
- Add a tree search type, substituteScopeSearch, which compares SymbolNodes against the scope. When a match is found, simplifyExpressionString is run (awkwardly due to a TypeError: "...not a function" issue I ran into -- please assist) against the substituted expression and the resulting list of steps (if any) are passed to Node.Status.nodeChanged.

All tests pass, although I only added one additional test to begin to cover this functionality.

I did my best to employ similar patterns to those I found in the codebase as much as I could. There are sure to be some clumsily-handled areas; please provide feedback on those as I am still learning.

Thanks and I look forward to your review,
Matt Marino


Ephs05msm:
Among other mistakes, I accidentally created a [test file](https://github.com/socraticorg/mathsteps/pull/212/files#diff-571cbdf24fea0bb75d02830eb02d4b6d) early on and then never ended up using it. Should probably be removed. Sorry about that.

evykassirer:
yeah feel free to go through and remove anything not directly related to your PR

I noticed there's also a `yarn.lock` file which shouldn't be in this PR either - you could add it to the gitignore file!

I think it's getting late for me tonight but I hope to take a look at this in the next couple days :) Thanks for your hard work (it's so cool seeing your learning process and how you do so much without much formal educational background!!!) and for your patience :)

Ephs05msm:
Sounds good. Couple more commits to get rid of some of the obviously unnecessary stuff. Still a bit of a noob with `.gitignore` so struggled to get the `.vscode` folder out. Sorry about that.

Saw a [conflicting note regarding `yarn.lock`](https://twitter.com/_ericelliott/status/792437742328090624?lang=en) so just left it alone for now.

Look forward to your feedback whenever you have time. No rush at all on my end. Thanks again!

Ephs05msm:
Not directly related to this issue, but I wanted to ask a quick question:

While using this code on another project I came across the following scenario that confused me. I added it to the test suite on my local branch and got the same result:

capture2
capture

Same result on `master` as well (using `['x = 100 + (25^2 / 4)', 'x = 256.25']`).

Is it by design that MathSteps won't simplify fractions that could be expressed as a float but not an (unrounded) integer?

evykassirer:
ooo that seems familiar but I'd have to go code hunting to be sure

I think the reason it's there is that in general we don't want to divide fractions midway through simplification (unless they're integers) because then you get like 5/3 x - 2/3 x turning into decimal subtraction which isn't what people usually want to see. I could see us adding a last step at the very end to divide the fraction but honestly I think people tend to prefer fractions over decimals anyways. It'd be neat to be able to show both the fraction and decimal at the end (I think cymath does this) but not sure how to fit that into how the code currently works, since we don't really let there be two answers right now

good question!

evykassirer:
answered in #215 :)

Ephs05msm:
Ran into a case where a symbol was nested inside of a UnaryMinus node type and adjusted to account for this. Also added a test to cover. My editor (VS Code) was having end of line sequence issues so fixing those makes it look like I changed a lot of things when in fact only the tests, TreeSearch, and scopeSubstitution search type were meaningfully changed. Sorry about the extra noise. All precommit checks pass.

.gitignore
```diff
@@ -1,5 +1,5 @@
-*swp
-node_modules
-*.log
-.DS_Store
-.idea
+*swp
+node_modules
+*.log
+.DS_Store
+.idea
```
evykassirer: can you keep the new line at the end of the file? Github seems to like it and puts a 🚫 if you don't, so I've adopted liking it too ^_^

I see it in a few other files, so just scroll this PR and you'll see where!

lib/TreeSearch.js
```diff
@@ -5,50 +5,55 @@ const TreeSearch = {};
...
   }
+  // Moved isUnaryMinus check above Symbol check to address symbol nested inside unaryminus
   else if (Node.Type.isUnaryMinus(node)) {
-    status = search(simplificationFunction, node.args[0], preOrder);
+    status = search(simplificationFunction, node.args[0], preOrder, scope);
     if (status.hasChanged()) {
       return Node.Status.childChanged(node, status);
     }
   }
+  // Break out isSymbol test and add a changeType for SUBSTITUTE_SYMBOL?
```
evykassirer: don't need this comment (it's more like a commit message and less like a long term comment, like above)

Though if you want to keep a comment here it'd probably be more like `// symbols can be simplified through variable substitution`

lib/TreeSearch.js
```diff
@@ -5,50 +5,55 @@ const TreeSearch = {};
...
     if (status.hasChanged()) {
       return status;
     }
   }
 
-  if (Node.Type.isConstant(node) || Node.Type.isSymbol(node)) {
+  if (Node.Type.isConstant(node)) {
     return Node.Status.noChange(node);
   }
+  // Moved isUnaryMinus check above Symbol check to address symbol nested inside unaryminus
```
evykassirer: try to avoid saying things like "moved" since future people looking at this code will only know the current state, not the past and that'd be confusing!

you could instead say something like "isUnaryMinus needs to be checked before Symbol because ..."

Though I'm confused why it has to be first. What test would fail if you checked symbol first? A unary minus is not a symbol, and unary minus recurses on its child! (let me know if I'm being confusing, not sure if I'm explaining this well ^^)

lib/Substitute.js
```diff
@@ -0,0 +1,16 @@
+function substituteScope(scope) {
```
evykassirer: can you name this file `substituteScope` to match its name? :)

lib/simplifyExpression/index.js
```diff
@@ -1,7 +1,9 @@
 const math = require('mathjs');
 const stepThrough = require('./stepThrough');
+const Substitute = require('../Substitute');
 
-function simplifyExpressionString(expressionString, debug=false) {
+function simplifyExpressionString(expressionString, debug=false, scope={}) {
```
evykassirer: it'd be nice if debug was the last argument, though I know that's a lot of moving code around so let me know if you want help going through and editing it all

evykassirer: also I think we should replace `scope` with `options` - I'll say more about this in a top level comment on this PR

lib/simplifyExpression/stepThrough.js
```diff
@@ -69,6 +70,9 @@ function step(node) {
   const simplificationTreeSearches = [
     // Basic simplifications that we always try first e.g. (...)^0 => 1
     basicsSearch,
+    // Substitute symbols present in the optional scope object with
+    // their respective expressions.
```
evykassirer: can you add an example similar to the other simplification searches here?

lib/simplifyExpression/substituteScopeSearch/index.js
```diff
@@ -0,0 +1,54 @@
+const ChangeTypes = require('../../ChangeTypes');
+const Node = require('../../node');
+const TreeSearch = require('../../TreeSearch');
+
+// Searches through the tree, prioritizing deeper nodes, and substitutes
+// in-scope values for their respective expressions on a symbol node if possible.
+// Returns a Node.Status object.
```
evykassirer: awesome documentation :D

lib/Substitute.js
evykassirer: also can you add a comment explaining the purpose of this function? I'm a bit confused about what it does - is it replacing scope things? Shouldn't that happen in the steps instead?

test/solveEquation/solveEquation.test.js
```diff
@@ -96,6 +96,8 @@ describe('solveEquation for =', function () {
     ['2/(1 + 1 + 4x) = 1/3', 'x = 1'],
     ['(3 + x) / (x^2 + 3) = 1', 'x = [0, 1]'],
     ['6/x + 8/(2x) = 10', 'x = 1'],
+    // Use of nested scope (i.e., baz depends on bar)
```
evykassirer: oo also could you test substeps to show how things get sub'd in one at a time? there are some tests that already test substeps you can probably base off of but let me know if you want help setting that up

lib/simplifyExpression/substituteScopeSearch/index.js
```diff
@@ -0,0 +1,54 @@
...
+
+// SUBSTITUTES
+// Returns a Node.Status object with substeps
+function substituteAndSimplifySymbols(node, scope) {
+  if (!Node.Type.isSymbol(node)) {
+    return Node.Status.noChange(node);
+  }
+
+  let symbolName;
+  if (node.type === 'SymbolNode') {
```
evykassirer: you can use `Node.Type.isSymbol(node)` instead

evykassirer: WAIT does this return true for unary minus too?

okay this makes sense https://github.com/socraticorg/mathsteps/blob/344d15b5d5e25251cef7d6976d6a73df16661d7d/lib/node/Type.js#L32

For any isSymbol thing you've added in this PR can you pass in `false` for the `allowUnaryMinus` argument? Then it can naturally recurse instead of you having to deal with it here. 

Sorry, that's weird that we did that (though I'm assuming there's a reason I've now forgotten XD but feel free to investigate if you're curious or open an issue)

lib/simplifyExpression/substituteScopeSearch/index.js
```diff
@@ -0,0 +1,54 @@
...
+    symbolName = node.name;
+  }
+  else if (Node.Type.isUnaryMinus(node)) {
+    symbolName = node.args[0].name;
+  }
+
+  if (scope.hasOwnProperty(symbolName)) {
+    // when declared at top, kept getting TypeError: ___ is not a function
```
evykassirer: can you explain a bit more about when the error happens? I can try replicating it on my computer and see what's up. Did it say `___ is not a function` or did it give a specific function name there?

lib/simplifyExpression/substituteScopeSearch/index.js
```diff
@@ -0,0 +1,54 @@
...
+    const simplifyExpression = require('../../simplifyExpression');
+    const substeps = simplifyExpression(scope[symbolName]);
```
evykassirer: instead of simplifying the expression here, I think it's best to just do `newNode = math.parse(scope[symbolName])`

then any simplification of the value can happen in later steps instead of substeps and it makes this code less complicated!

what do you think?

evykassirer:
ok so I looked into it and there's no reason for `allowUnaryMinus` to default to true - I've updated that on master, feel free to pull that in and then your code should work as it was before :)

Ephs05msm:
Thanks so much for the detailed feedback! I'm all for your suggestions and will get to work on them this week. If I get in over my head anywhere I'll ask for help.

Also, I really appreciate the encouragement. I've found there's a lot of self-doubt along the learning path (at least doing it the way I have been), so even a little bit of affirmation that some things are being done correctly is hugely helpful. Thank you.

 On we go :-)

lib/simplifyExpression/substituteScopeSearch/index.js
Ephs05msm: I'm a bit turned around here, perhaps due to the approach I took. Since simplifyExpression is where the scope substitution takes place, I'm concerned calling math.parse(scope[symbolName]) here won't be effective since it will not perform the scope substitution. That may be why you advised to put the substituteScope call deeper in the simplify function?

Left as is in my recent (incremental) PR.

test/solveEquation/solveEquation.test.js
Ephs05msm: Made some progress on this, but I struggled to get any of the simplify tests to pass. Decided to push my incremental changes in hopes we can work through this together. The simplify tests weren't originally designed with scope in mind, and my quick attempt to incorporate it fell short.

.gitignore
Ephs05msm: Hopefully caught all these in my latest PR. Will pay attention to this going forward.

lib/simplifyExpression/stepThrough.js
Ephs05msm: Added an example but let me know if it needs to be reworded.

lib/simplifyExpression/substituteScopeSearch/index.js
Ephs05msm: I left this alone for now since you followed up with the change to `allowUnaryMinus`. Let me know if we should revisit.

Ephs05msm:
I replaced the `scope` arg with `options` as you suggested, and also attempted to reorder the args so `debug` is last. Tests pass but it's certainly possible I didn't address it in every place/case. Happy to double back with you if you find issues.

Ephs05msm:
Thanks again for taking the time to review. I think this latest round of changes get closer to the mark, but there are still some loose ends (especially on the testing side). I felt it better to commit some progress than hold back just to spin my wheels in a few spots. Look forward to continuing to work with you on this :-).

test/solveEquation/solveEquation.test.js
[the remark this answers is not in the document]
evykassirer: >  The simplify tests weren't originally designed with scope in mind, and my quick attempt to incorporate it fell short.

Yeah I think putting the tests for this in their own completely different file would help - then we don't need to update the other tests to deal with scope and can do a different setup for these tests. Let me know if you want help going through that - I'm a lot more free this week and next than I have been recently and would love to help (sorry for just getting to this PR again now!)

lib/substituteScope.js
```diff
@@ -0,0 +1,24 @@
+function substituteScope(scope) {
```
evykassirer: just copying from before so it's easier to keep track of what's left (no worries for not getting to it yet)

> can you name this file substituteScope.js to match its name? :) 
> also can you add a comment explaining the purpose of this function? I'm a bit confused about what it does - is it replacing scope things? Shouldn't that happen in the steps instead?

lib/simplifyExpression/stepThrough.js
evykassirer: looks good to me! awesome
I think you can take out the comma after the `e.g.` though

lib/simplifyExpression/substituteScopeSearch/index.js
```diff
@@ -31,8 +35,10 @@ function substituteAndSimplifySymbols(node, scope) {
     symbolName = node.args[0].name;
   }
 
+  const scope = options.scope;
   if (scope.hasOwnProperty(symbolName)) {
-    // when declared at top, kept getting TypeError: ___ is not a function
+    // when declared at top, kept getting
+    // TypeError: <nameIGaveToFunction> is not a function
```
evykassirer: ^ still gotta figure this thing out

evykassirer: I will look into this...

lib/TreeSearch.js
```diff
@@ -5,50 +5,57 @@ const TreeSearch = {};
...
     }
   }
 
-  if (Node.Type.isConstant(node) || Node.Type.isSymbol(node)) {
+  if (Node.Type.isConstant(node)) {
     return Node.Status.noChange(node);
   }
+  // Check isUnaryMinus before Symbol because symbols nested inside unaryMinus
+  // (e.g., 2x - symbol) need to be subtracted from the expression. Checking
+  // symbol first causes the symbol to be added.
```
evykassirer: I think with my unary minus change, it doesn't have to be checked first - can you switch it back and see if that works? Before, symbols with unary minuses were counted as symbols too by default, but now it `Node.Type.isSymbol(node)` on a unary minus node should return false by default :D

lib/simplifyExpression/substituteScopeSearch/index.js
```diff
@@ -0,0 +1,60 @@
...
+
+  if (!Node.Type.isSymbol(node)) {
+    return Node.Status.noChange(node);
+  }
+
+  let symbolName;
+  if (node.type === 'SymbolNode') {
+    symbolName = node.name;
+  }
+  else if (Node.Type.isUnaryMinus(node)) {
```
evykassirer: now that I updated the `Node.Type.isSymbol` definition, you'll never get unary minuses in here anymore which makes things much easier hopefully!

so you can take out `if (node.type === 'SymbolNode') {` and `else if (Node.Type.isUnaryMinus(node)) {` and always just do ` symbolName = node.name;`

lib/simplifyExpression/substituteScopeSearch/index.js
```diff
@@ -0,0 +1,60 @@
...
+    symbolName = node.args[0].name;
+  }
+
+  const scope = options.scope;
+  if (scope.hasOwnProperty(symbolName)) {
+    // when declared at top, kept getting
+    // TypeError: <nameIGaveToFunction> is not a function
```
evykassirer: what is `nameIGaveToFunction` - which function?

lib/simplifyExpression/substituteScopeSearch/index.js
```diff
@@ -0,0 +1,60 @@
...
+    const simplifyExpression = require('../../simplifyExpression');
```
evykassirer: so `simplifyExpression` is a pretty high level function - it's the big thing at the top that calls everything else. Because of that I feel weird about using it here.

What you have here also means if you have `x = 2/3 + 4/6`, the substitution will go from `x` to `8/6` with the adding of fractions as substeps, which from a teaching perspective also feels weird to me.

My suggestion instead is:

1. take out the `const newScope = Substitute(options.scope);` above and keep the options exactly as they were passed in
2. here, instead of doing `const substeps = simplifyExpression(scope[symbolName]);` do `newNode = math.parse(scope[symbolName])`

I know it's a bit different from your idea, so we could chat more about this and make sure we're on the same page and both happy with whatever we decide to do. I think adding some tests that show multiple steps will also make it more clear what this changes. Let me know if you have more questions or comments about this!

evykassirer:
ok I think I addressed all your remaining comments/questions!

looks really awesome so far, I'm a fan of the `options` thing and `debug` last is 👌 (I am also not sure if you caught them all but seems good!) Thanks for your patience with my review and for working on this new really interesting feature!!

Ephs05msm:
Thanks, @evykassirer! Planning to spend time on this next week.

lib/simplifyExpression/substituteScopeSearch/index.js
Ephs05msm: ...this was my way of trying to communicate that, no matter what name I gave to the function that made this call, it would fail with that error message. `nameIGaveToFunction` would be substituted with whatever I tried. IIRC, I pointed this out because I thought the issue might have been with a name conflict of some kind. Doesn't appear to be the case, though.

evykassirer: which is the function making the call? is it in the PR here? what's it called right now?

There's so many functions in the code haha so confused which one you're referring to!

Ephs05msm:
Hi @evykassirer, just wanted to to circle back and let you know that I haven't lost sight of this. Just been focused on a few other things instead. Since this is an isolated branch representing a marginal feature, is it safe to assume that it's alright if I take a break for a bit? If not, I'll make the time.

Thanks again for your help.

evykassirer:
no worries! I think you're fine, but if the porting change goes through soon and messes you up I'd be happy to help you refactor your stuff (it'd be just a bunch of really small changes) :)

Issue #162, opened 2017-04-29
Bug: rounding error with negative numbers


evykassirer:
There is an issue because of this line of code:

https://github.com/socraticorg/mathsteps/blob/master/lib/simplifyExpression/arithmeticSearch/index.js#L52

Negative numbers go into the `if` statement right now, which keeps only 4 significant digits

e.g. `-1953125` becomes `-1953000`

That line should be changed to `if (-1 < result < 1)`, which means if the number is `0. ....` or  `-0. ....` it will round to only 4 digits after the decimal, which was my original intention (but I forgot about negative numbers 😅)

This is a one line code change (plus adding a test to make sure it's not broken anymore), which is great for a first PR!

to read more about how we found this issue, see #72

starside:
Hi, I would love to take on this issue.  I have been lurking the mailing list for a while, this seems like a good first step.

nitin42:
Also see #160

evykassirer:
Hi @starside ! I'm so excited that you're wanting to get involved with mathsteps!

Unfortunately, for the next few weeks I'm planning to block changes to this project, because there are some people working on a huge change to the codebase that'll probably change almost every line of code here, haha - so I wanna wait until that's done before making any other changes (and I'll need to update open issues and everything once it's done too). You can check out what's going on by looking at the `porting` branch or reading the open pull requests, or chatting with @aliang8 :)

So so sorry that the moment you expressed interest is one when it's hard for you to contribute, but I'd love for you to get involved as soon as this big reshuffle is done! (Or maybe @aliang8 can find some small tasks to get you started in helping with porting, if you're interested 😄 )

Let me know if you have any questions

starside:
Thanks!  I guess I will have to wait, unless @aliang8 has something for me to do.  You are moving to a custom parser correct?

evykassirer:
Yes! It's going to be awesome. We're also changing the way you define simplifying rules to be more intuitive and easier to write. I'm hoping it'll make contributing to mathsteps much easier, and that it'll also make it easier to implement more complicated cool things soon!!

aliang8:
Hey @starside!!! Sorry for not responding on this thread. I've been somewhat busy with graduation related events. But regardless, I'm super excited for you to join and if you are interested, I do have some tasks for you to do. Maybe we can move away from this thread and chat privately on gitter?

starside:
Gitter works for me

evykassirer:
[...]

Issue #198, opened 2017-07-16
Operator nodes can have <= 1 arguments


ldworkin:
See the discussion here: https://github.com/socraticorg/mathsteps/pull/190#discussion_r127018474

Essentially, if check for `args.length > 1` in the `isOperator` check, some tests involving nthRoots will fail. Why is that? All operator nodes should have at least 2 arguments.

evykassirer:
thanks for opening this :D

Issue #109, opened 2017-01-30
Prioritize simple fraction simplification


aelnaiem:
`2 - 6/6` will convert 2 into a fraction and do the subtraction instead of simplifying 6/6 to 1. `2 - 12/6` will do the same.

This should be a simple check to see if any fractions can evenly divide out before we combine them.

evykassirer:
I thought this logic already existed somewhere, so probably a bug in ordering of simplifications? I thought simplify arithmetic always went first and divided things first if they became integers

aelnaiem:
Yea, I was surprised when I saw this happen. Could just be a bug

PR #216, opened 2017-07-29
#162 bugfix - Large negtive number rounding

starside:
Fixed bug #162, and added tests to check that large negative
numbers do not round.  Also commented out broken test related to
test #162.

### DIFF
--- a/test/simplifyExpression/arithmeticSearch/arithmeticSearch.test.js
+++ b/test/simplifyExpression/arithmeticSearch/arithmeticSearch.test.js
@@ -12,7 +12,7 @@ describe('evaluate arithmeticSearch', function () {
     ['2*3*5', '30'],
     ['6*6', '36'],
     ['9/4', '9/4'], //  does not divide
-    ['16 + -1953125', '-1953109'], // verify large negative number round correctly
+    ['16 -1953125', '-1953109'], // verify large negative number round correctly
   ];
   tests.forEach(t => testArithmeticSearch(t[0], t[1]));
 });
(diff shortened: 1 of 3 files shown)


test/solveEquation/solveEquation.test.js
```diff
@@ -110,6 +109,7 @@ describe('solveEquation for =', function () {
     // TODO: figure out what to do about errors from rounding midway through
     // this gives us 6.3995 when it should actually be 6.4 :(
     // ['x - 3.4= ( x - 2.5)/( 1.3)', 'x = 6.4']
+    // ['( u )/( 0.3) = 4u + 6.28', 'u = -9.42'], //Commented out for bug #162
```
evykassirer: I don't think you need to add the comment, it's now just another case of the bug and people in the future don't need to know it once passed

I find it so funny how it was passing because we were rounding wrong and now is failing haha

test/simplifyExpression/arithmeticSearch/arithmeticSearch.test.js
```diff
@@ -12,6 +12,7 @@ describe('evaluate arithmeticSearch', function () {
     ['2*3*5', '30'],
     ['6*6', '36'],
     ['9/4', '9/4'], //  does not divide
+    ['16 + -1953125', '-1953109'], // verify large negative number round correctly
```
evykassirer: I think this will work with `16 -1953125` which makes more sense - can you try that?

evykassirer:
🎉 yay! thanks for fixing this

just two small comments then let's merge this!! :D

starside:
Thanks for the comments,  I have made the changes.  Assuming this passes, should I pick some open issue and give it a go?

evykassirer:
yup! feel free to work on something else open. I'm hoping to move the focus to porting over to math-rules and math-parser soon, so feel free to check those out at https://github.com/semantic-math

If you wanna do some more stuff in the meantime:

- #198 would be goo to look into
- #146 would be really helpful and will continue to stay in the codebase despite changing to a new parse tree (not sure what else will stay vs change)
- #109 is a neat teaching related one to look into (we don't only want to get the right answer, we want to get there in a way that makes sense!)

feel free to check out some other open issues, those were just some that stood out to me

thanks for spending your time on this project! I really appreciate it. If you have any questions or need help with anything, just let me know :)

Issue #232, opened 2018-01-19
Cancel like terms is dependent of the order of factors in the numerator and denominator


marpol2:
step 1: (45 * 3)/(45 * 3)
step 2: 45 / 45
step 3: (1 * 45) / (1 * 45)
step 4: 1

while instead
step 1: (45 * 3)/(3 * 45)
step 2: (15 * 3) / 45
step 3: (1 * 3) / 3
step 4: 3 / 3
step 5: (1 * 3) / (1 * 3)
step 6: 1

This can probably be fixed by just adding a check before proceeding to "cancelTerms" in "cancelLikeTerms".

evykassirer:
oh- good catch! thank so much :)

I'm not prioritizing my time on writing code for mathsteps right now, but would be happy to review a change if you (or anyone else) puts one up!

PR #237, opened 2018-05-16
Fix order dependence on cancel like terms

marpol2:
-fixed issue #232

-changed such that in the expression 5 / 128 + 32 / (32 * 4) the denominator in the second fractions is the first to be evaluated since both denominators are equal.


marpol2:
...

evykassirer:
hey sorry for letting this slide! I've been prioritizing a lot less time on mathsteps lately

next week I'm gonna have a lot of free time and I'll take a look :)

PR #250, opened 2020-04-09
Better lone parenthesis handling

GabrielMahan:
https://github.com/socraticorg/mathsteps/issues/249

Hi there! This is my first contribution to mathsteps, so apologies if this is not a very robust solution; happy to investigate further.

PR #251, opened 2020-04-10
Parenthesis on both sides

GabrielMahan:
Follow-up from #250 .

In #250 I updated `isolateSymbolOnLeftSide` but neglected to also update `removeSymbolFromRightSide`.

I think because https://github.com/socraticorg/mathsteps/pull/250/files#diff-2dc14e1c8322de958fc4c52d204c30b5R42-R45, the right side (rightly) returns a node with a symbol, but because I didn't modify `removeSymbolFromRightSide`, an error was occurring when the right side couldn't handle the parenthesis node.

Thanks @evykassirer !

### DIFF
--- a/lib/solveEquation/EquationOperations.js
+++ b/lib/solveEquation/EquationOperations.js
@@ -68,7 +68,13 @@ EquationOperations.removeSymbolFromDenominator = function(equation, symbolName) 
 // by multiplying or dividing by a symbol term.
 // TODO: support inverting functions e.g. sqrt, ^, log etc.
 EquationOperations.removeSymbolFromRightSide = function(equation, symbolName) {
-  const rightNode = equation.rightNode;
+  let rightNode = equation.rightNode;
+
+  if (Node.Type.isParenthesis(rightNode)) {
+    // if entire left node is a parenthesis, we can ignore the parenthesis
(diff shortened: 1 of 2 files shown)


evykassirer:
hm ok so as I was laying in bed last night trying to sleep, I remembered there's a 'removeUnnecessaryParens` that theoretically should solve all these issues, and it's called here https://github.com/socraticorg/mathsteps/blob/master/lib/solveEquation/stepThrough.js#L79 which I think should be before the point where your error comes up?

would you mind looking into this and seeing why you were having those problems despite 'removeUnnecessaryParens`? I remember now that we were trying to avoid putting parenthesis code throughout the codebase, we mostly assumed that unnecessary parens just wouldn't be there

xZGit:
https://github.com/socraticorg/mathsteps/blob/master/lib/solveEquation/stepThrough.js#L58

Set the value of the param "rootNode"  to true can fix it 
```
 equation.leftNode = removeUnnecessaryParens(equation.leftNode, true);
 equation.rightNode = removeUnnecessaryParens(equation.rightNode, true);
```

GabrielMahan:
hey @evykassirer sorry for the delay! will close this PR and address separately

